PR #14, opened 2019-08-23
[firebase_dynamic_links] Don't crash if activity is missing

mehmetf:
registrar.activity can be null with dynamic links since it is possible to open the app and execute this code before the activity is ready.

This has already been tested by @bparrishMines. It is identical to https://github.com/flutter/plugins/pull/1989.


mehmetf:
This fixes https://github.com/FirebaseExtended/flutterfire/issues/7

mehmetf:
Since this has been reviewed and tested by Maurice already, I assume I have his LGTM.

PR #15, opened 2019-08-23
[firebase_dynamic_links] Change the OnLinkError object to be a real exception

mehmetf:
Firebase dynamic links creates a new error object. This should extend PlatformException (it effectively carries the same signature) so that it can be thrown by the app if needed.


mehmetf:
This fixes https://github.com/FirebaseExtended/flutterfire/issues/6

collinjackson:
LGTM once bots are happy and changelog included

PR #16, opened 2019-08-23
[firebase_auth] Add a constructor for token result object with a token

mehmetf:
We have applications that implement FirebaseUser class. In order to support this use case, classes such as IdTokenResult need to be open to being constructed as opposed to having only an internal API.

Fixes https://github.com/FirebaseExtended/flutterfire/issues/5


mehmetf:
Yes, I will write a test for this.

mehmetf:
Done, please take another look.

mehmetf:
Indeed, that's a better solution.

A test would not be required in that case, I would presume since you can always implement any extra classes that get introduced. Would you agree?

collinjackson:
> Indeed, that's a better solution.
> 
> A test would not be required in that case, I would presume since you can always implement any extra classes that get introduced. Would you agree?

Yes, I agree.

mehmetf:
Closing this PR; nothing to see here :-)

PR #31, opened 2019-08-26
[cloud_firestore] Fix possible NullPointerException

BugsBunnyBR:
## Description
Removed Activity Reference from plugin to avoid possible NullPointerException.
AsyncTasks are not needed to run Transactions, however [transactions are not guaranteed run in UI Thread](https://firebase.google.com/docs/firestore/manage-data/transactions#passing_information_out_of_transactions) so the Handler is still needed to send data to Dart.

## Related Issues

https://github.com/flutter/flutter/issues/38692

## Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [X] All existing and new tests are passing.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] I read and followed the [Flutter Style Guide].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I signed the [CLA].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

### DIFF
--- a/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
+++ b/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
@@ -4,8 +4,8 @@
 
 package io.flutter.plugins.firebase.cloudfirestore;
 
-import android.app.Activity;
 import android.os.AsyncTask;
+import android.os.Handler;
 import android.util.Log;
 import android.util.SparseArray;
 import androidx.annotation.NonNull;
@@ -59,8 +59,7 @@ public class CloudFirestorePlugin implements MethodCallHandler {
 
   private static final String TAG = "CloudFirestorePlugin";
   private final MethodChannel channel;
-  private final Activity activity;
-
+  private final Handler handler = new Handler();
   // Handles are ints used as indexes into the sparse array of active observers
   private int nextListenerHandle = 0;
   private int nextBatchHandle = 0;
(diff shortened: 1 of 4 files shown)


packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
```diff
@@ -419,7 +423,7 @@ public void notImplemented() {
                       // Wait till transaction is complete.
                       try {
                         String timeoutKey = "transactionTimeout";
-                        long timeout = ((Number) arguments.get(timeoutKey)).longValue();
+                        Long timeout = call.argument(timeoutKey);
```
collinjackson: I found this surprising, I would expect that you'd need to call `longValue` here if an `int` was passed from Dart. Are you sure?

BugsBunnyBR:
@collinjackson there is no [documentation](https://firebase.google.com/docs/firestore/manage-data/transactions) in cloud firestore saying that we need to offload the actions to another thread. With my experience using firebase in android, that is never the case, as any work that need to be in a separated thread is executed with `Task` api.
If you want I can try to improve the sample so it touches more places in plugin, but I would not recommend bringing the AsyncTasks back.

packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
BugsBunnyBR: Well, it should work.
If you look [here](https://flutter.dev/docs/development/platform-integration/platform-channels#platform-channel-data-types-support-and-codecs) you can see that a dart `int` can fit a java `Long`.

BugsBunnyBR: Ohh, I tested. It broke. Moving to Integer solves.

BugsBunnyBR:
@collinjackson I partly understood the problem.
When you run a transaction with just one operation it seems to work alright.
When the transaction has multiple operations the AsyncTask makes it work, but it is really odd that need. I included some actions in the example so it will easier to spot.

BugsBunnyBR:
I am closing this PR as I am not able to keep it updated.

PR #39, opened 2019-08-26
[all] Switch Android from api dependencies to implementation dependencies

collinjackson:
Flutterfire plugins don't expose a public API to other Java libraries (with the exception of the plugin registration process). I am wondering if it would be reasonable to allow a developer to mix and match plugins that depend on different versions of Firebase Android SDK.

Using implementation dependencies instead of api dependencies gets our dependencies off the classpath of the developer's app and reduces the chance of Gradle build errors due to version resolution conflicts. Being able to mix and match Flutterfire plugins on Android might reduce the need for "symbolic" Dart packages.

Here's an article on the differences of api vs implementation in libraries:
https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

I think this might be a more ideal approach for the problem that is addressed by https://github.com/FirebaseExtended/flutterfire/pull/12 where adding an explicit Firebase dependency to the developer's app-level build.gradle can lead to unexpected build breaks elsewhere.

Possible con: It might increase the size of builds slightly if developers are depending on multiple implementations of Firebase Android SDKs. However, I'm not seeing this in practice.

#### Related Issues

flutter/flutter#37837

### DIFF
--- a/packages/cloud_firestore/android/build.gradle
+++ b/packages/cloud_firestore/android/build.gradle
@@ -34,7 +34,7 @@ android {
         disable 'InvalidPackage'
     }
     dependencies {
-        api 'com.google.firebase:firebase-firestore:19.0.0'
+        implementation 'com.google.firebase:firebase-firestore:19.0.0'
         implementation 'com.google.firebase:firebase-common:16.1.0'
         implementation 'androidx.annotation:annotation:1.0.0'
     }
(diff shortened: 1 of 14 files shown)


kroikie:
We are switching to using BoM for managing Android dependencies so closing this.

PR #71, opened 2019-08-31
Update issue templates

kroikie:
Create templates for bug reports and feature requests.


kroikie:
Since this is not a code change I'll merge this, ignoring the build.

PR #141, opened 2019-09-12
[firebase_auth] Update iOS CocoaPod dependency.

collinjackson:
Avoids deprecation warnings relating to `UIWebView`.

Running `pod repo update` may be required.

Related issue: flutter/flutter#39470

### DIFF
--- a/packages/firebase_auth/ios/firebase_auth.podspec
+++ b/packages/firebase_auth/ios/firebase_auth.podspec
@@ -21,7 +21,7 @@ Firebase Auth plugin for Flutter.
   s.public_header_files = 'Classes/**/*.h'
   s.ios.deployment_target = '8.0'
   s.dependency 'Flutter'
-  s.dependency 'Firebase/Auth', '~> 6.0'
+  s.dependency 'Firebase/Auth', '~> 6.8'
   s.dependency 'Firebase/Core'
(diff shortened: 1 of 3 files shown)


chenna-reddy:
@kroikie Could you plese check? Couldn't make any iOS release from last 3 weeks due to this deprecated dependency.

bparrishMines:
I think removing the pod dependency and telling users to use `pod update` is the better choice. Unless there was a specific reason we want to choose a specific version.

collinjackson:
OK, I'm going to go ahead and close this one without merging then. Hopefully search results will lead people encountering the error to follow the instructions at https://github.com/flutter/flutter/issues/39470#issuecomment-532825508

packages/firebase_auth/ios/firebase_auth.podspec
```diff
@@ -21,7 +21,7 @@ Firebase Auth plugin for Flutter.
   s.public_header_files = 'Classes/**/*.h'
   s.ios.deployment_target = '8.0'
   s.dependency 'Flutter'
-  s.dependency 'Firebase/Auth', '~> 6.0'
```
chenna-reddy: Should this version be removed and let it pick latest version always??

PR #142, opened 2019-09-12
[firebase_dynamic_links] Update CocoaPods dependency

collinjackson:
Avoids deprecation warnings relating to `UIWebView`.

Running `pod repo update` may be required.

Related issue: flutter/flutter#39470


bparrishMines:
I think removing the pod dependency and telling users to use `pod update` is the better choice. Unless there was a specific reason we want to choose a specific version.

collinjackson:
OK, I'm going to go ahead and close this one without merging then. Hopefully search results will lead people encountering the error to follow the instructions at https://github.com/flutter/flutter/issues/39470#issuecomment-532825508

PR #169, opened 2019-09-17
[firebase_admob] Native ads

bparrishMines:
## Description

Add basic support for [Firebase Native Ads](https://developers.google.com/admob/android/native/start) on Android.

## Related Issues

https://github.com/flutter/flutter/issues/35082

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
```diff
@@ -219,4 +233,47 @@ void show() {
...
+    }
+
+    @Override
+    void load(String adUnitId, Map<String, Object> targetingInfo) {
+      status = Status.LOADING;
+
+      final AdLoader adLoader = new AdLoader.Builder(activity, adUnitId)
+          .forUnifiedNativeAd(new UnifiedNativeAd.OnUnifiedNativeAdLoadedListener() {
+            @Override
+            public void onUnifiedNativeAdLoaded(UnifiedNativeAd unifiedNativeAd) {
```
bparrishMines: NativeAds don't call `onLoaded` like all other Ads. They call this method instead because it passes the user a `UnifiedNativeAd`. 

Since the information in `UnifiedNativeAd` is used to create a View is only useful on the Platform side, I have this Ad call the normal `onLoaded` method.

packages/firebase_admob/lib/firebase_admob.dart
```diff
@@ -286,6 +287,48 @@ class BannerAd extends MobileAd {
...
+/// Native ads are ad assets that are presented to users via UI components that
+/// are native to the platform. They're shown using the same types of views with
+/// which you're already building your layouts, and can be formatted to match
+/// the visual design of the user experience in which they live. In coding
+/// terms, this means that when a native ad loads, your app receives a NativeAd
+/// object that contains its assets, and the app (rather than the Google Mobile
+/// Ads SDK) is then responsible for displaying them.
+class NativeAd extends MobileAd {
+  NativeAd(
+      {@required String adUnitId,
```
amirh: nit: use the same formatting as in MobileAd above

packages/firebase_admob/lib/firebase_admob.dart
```diff
@@ -286,6 +287,48 @@ class BannerAd extends MobileAd {
...
+            targetingInfo: targetingInfo,
+            listener: listener) {
+    if (!Platform.isAndroid) {
+      throw UnimplementedError(
+        '$NativeAd is currently only available for Android',
+      );
+    }
+  }
+
+  /// A platform-specific AdMob test ad unit ID for Native ads. This ad unit
```
amirh: nit: this comment seems to be repeated multiple times, may be better to use a macro.

packages/firebase_admob/example/android/app/src/main/java/io/flutter/plugins/firebaseadmobexample/MainActivity.java
```diff
@@ -5,13 +5,23 @@
...
 public class MainActivity extends FlutterActivity {
   @Override
   protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     GeneratedPluginRegistrant.registerWith(this);
+    FirebaseAdMobPlugin.setNativeAdGenerator((UnifiedNativeAd ad) -> {
+      final TextView myAdView = new TextView(this);
+      myAdView.setWidth(200);
+      myAdView.setHeight(200);
+      myAdView.setText("This is an ad!");
```
amirh: Let's show some real add content here

packages/firebase_admob/README.md
```diff
@@ -184,6 +184,34 @@ RewardedVideoAd.instance.listener =
...
+ 
+and add this line to the `onCreate()` method:
+
+```java
+FirebaseAdMobPlugin.setNativeAdGenerator((UnifiedNativeAd ad) -> {
+  // return a View;
+});
+```
+
+## iOS
```
amirh: nit: missing one `#`

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
```diff
@@ -24,6 +27,12 @@
 
   RewardedVideoAdWrapper rewardedWrapper;
 
+  static Function<UnifiedNativeAd, View> nativeAdGenerator;
```
amirh: Should this always build a UnifiedNativeAdView? (seems like it's a requirement according to: https://developers.google.com/admob/android/native/advanced )

amirh: Is it reasonable for an app to have multiple Native Ad factories? (e.g the same app wants to potentially display native ads in multiple ways?)

amirh: nit: I'd consider calling this factory

packages/firebase_admob/README.md
```diff
@@ -184,6 +184,34 @@ RewardedVideoAd.instance.listener =
 Because `RewardedVideoAd` is a singleton object, it does not offer a `dispose`
 method.
 
+## Using native ads
+
```
amirh: nit: May be worth to add a paragraph explicitly explaining that in order to use native adds you will have to provide Android/iOS code for displaying it.

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
BugsBunnyBR: Please, avoid static usage for anything with a View or context. It may become a memory leak quite easily.

packages/firebase_admob/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.1
+
+* Add basic Native ad support for Android.
```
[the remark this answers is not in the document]
bparrishMines: done https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-036d5f53983cfc25a7b8afedd57aa9b3R3

packages/firebase_admob/README.md
bparrishMines: done https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-759637e0c287b36498a4cd4377d8a349R189

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
bparrishMines: done https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-7c229379ed87228a4d21cd88176456f7R30

bparrishMines: I don't believe it is required to return a UnifiedNativeAdView. I saw this as well and it only states `This class is a ViewGroup that publishers should use as the root for the UnifiedNativeAd`.

If it was required i would assume it would say `must`? There may be a situation where it isn't required, but I haven't looked further into it.

packages/firebase_admob/README.md
bparrishMines: done https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-759637e0c287b36498a4cd4377d8a349R223

packages/firebase_admob/example/android/app/src/main/java/io/flutter/plugins/firebaseadmobexample/MainActivity.java
bparrishMines: I displayed the body of the ad. https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-729cfe12b0f986dfdcd85db6a241500aR24

Screenshot_20190923-133427

packages/firebase_admob/lib/firebase_admob.dart
bparrishMines: done https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-8324a81da49ba02b4e4b58408c3db9cbR271

I believe this is how it is done according to https://github.com/dart-lang/dartdoc#macros

packages/firebase_admob/lib/firebase_admob.dart
bparrishMines: done https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-8324a81da49ba02b4e4b58408c3db9cbR314

packages/firebase_admob/lib/firebase_admob.dart
```diff
@@ -286,6 +287,48 @@ class BannerAd extends MobileAd {
   }
 }
 
+/// A NativeAd for the [FirebaseAdMobPlugin].
+///
+/// Native ads are ad assets that are presented to users via UI components that
```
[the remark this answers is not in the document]
bparrishMines: done. https://github.com/FirebaseExtended/flutterfire/pull/169/files#diff-8324a81da49ba02b4e4b58408c3db9cbR296

I added details on what is shown and explicitly stated it doesn't support Flutter Widgets.

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
bparrishMines: I'm not sure what you mean by this. The user should only be passing a `View` through a static method and setting it to a member `ViewGroup`. How would the `View` leak?

packages/firebase_admob/example/android/app/src/main/java/io/flutter/plugins/firebaseadmobexample/MainActivity.java
bparrishMines: Nvm. I missed a key point on the website that states it's required.

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
bparrishMines: Nvm. I missed a key point on the website that states it's required.

bparrishMines: Couldn't the factory handle this situation? Is there more information that the factory would need to display Native Ads differently?

Although, this does highlight that the creation of the Ad is done on the platform side while the placement is set on the dart side (e.g. anchoring the ad to the top or bottom). The user might want to place the ad based on its content.

A simple solution would be to pass the `UnifiedNativeAd` data to the user on the dart side so the platform and dart sides will sync on how to place and create the ad. This could be a separate PR though after the initial basic implementation for Android and iOS.

BugsBunnyBR: That anonymous function in the Activity will keep an implicit reference to that Activity. That anonymous function is used as factory in the plugin and it is kept as a static reference in the plugin class. It will cause the Activity to leak. A better way to get a reference is to avoid the static reference and make the Activity implement an interface. All of this without using `static` anywhere.

XSiyabonga:
Can I implement this branch before its merged with master?

zukinaru:
> Can I implement this branch before its merged with master? Native ads are critical for my app

yes for me too...

waiting this for months...

packages/firebase_admob/README.md
```diff
@@ -184,6 +184,46 @@ RewardedVideoAd.instance.listener =
...
+Native ads are presented to users via UI components that
+are native to the platform. (e.g. A
+[View](https://developer.android.com/reference/android/view/View) on Android or a
+[UIView](https://developer.apple.com/documentation/uikit/uiview?language=objc)
+on iOS). Using Flutter widgets to create native ads is NOT supported by
+this.
```
amirh: nit: by this plugin

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
amirh: Interesting, I'm probably missing something, how would this work for an app with 2 ads on the same screen, each should be styled differently (say one is presented over a light background and the other is presented over a dark background).

When this factory is invoked, how does it know whether it should style for the dark background or the light background?

charleshan:
Look like some progress is being made 🙌

https://github.com/FirebaseExtended/flutterfire/pull/169

jesusmartinoza:
Is it safe to use this PR?

bparrishMines:
@jesusmartinoza @XSiyabonga @zukinaru You can fork and use this branch before its merged, but at your own risk. This code doesn't have tests and I will be making changes to it at some point that will probably break your code. Most likely within a week or so.

charleshan:
Any updates @bparrishMines?

A few hundred people are waiting for this to be added at
https://github.com/flutter/flutter/issues/12114

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
bparrishMines: There are [NativeAdOptions](https://developers.google.com/android/reference/com/google/android/gms/ads/formats/NativeAdOptions) for NativeAds, so we could add a parameter in `NativeAd` that passes custom arguments. Similar to `PlatformViewFactory.`

I would add this in a different PR after iOS is implemented.

bparrishMines: I disagree that an anonymous function will keep a reference of the object instantiating it. However, in this context, the function does reference the `Activity` that it is in, so it would keep a reference. This may have been what you meant.

I also agree that your approach with the interface would be cleaner especially as we move away from `static` variables with the v2 embedding migration. I set up a `NativeAdFactory` interface instead.

amirh: NativeAdOptions seems to be a very constrained object, how would you e.g express light/dark background using it?

amirh:
One thing I'm a little hesitant about is designing this around the assumption that the current activity implements some interface specific for this plugin which defines the looks of native ads. I can imagine an add2app scenario with multiple activities that are all using the same Flutter fragment which will require some hairy plumbing to get this properly working. (More abstractly, it feels like a concern that does not belong in the Activity class.)

I'd love for us to use a more implicit "give me the admob plugin for this engine" on the Java side, and then register a native ads factory on it.

With the v1 embedding we could use PluginRegistry#publish for the AdMob plugin to publish an object that accepts registrations of native view factories, and then the developer will use PluginRegistry#valuePublishedBy to fetch that object and register their native ads factory.

AFAIK we do not support something like that in the v2 embedding, I think we should add a similar mechanism to the v2 embedding.

@matthew-carroll @xster - thoughts?

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
bparrishMines: Discussed offline:

We'll include a `Map<String, dynamic> customOptions` parameter in the `NativeAdFactory. createNativeAd` method.

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,95 @@ RewardedVideoAd.instance.listener =
...
+
+### Android
+The Android Admob Plugin requires a class that implements `NativeAdFactory` which calls `createNativeAd(
```
amirh: s/calls/implements/

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,95 @@ RewardedVideoAd.instance.listener =
...
+[UnifiedNativeAd](https://developers.google.com/android/reference/com/google/android/gms/ads/formats/UnifiedNativeAd) nativeAd,
+Map<String, Options> customOptions)` and returns a
+[UnifiedNativeAdView](https://developers.google.com/android/reference/com/google/android/gms/ads/formats/UnifiedNativeAdView).
+
+You can implement this in your `MainActivity.java` or create a separate class in the same directory
```
amirh: Do we want to recommend doing this in the activity?

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,95 @@ RewardedVideoAd.instance.listener =
...
+class NativeAdFactoryExample implements NativeAdFactory {
+  @Override
+  public UnifiedNativeAdView createNativeAd(
+      UnifiedNativeAd nativeAd, Map<String, Object> customOptions) {
+    // Create UnifiedNativeAdView
+  }
+}
+```
+ 
+After, pass an instance of `NativeAdFactory` to the `FirebaseAdMobPlugin`. This is done differently whether
```
amirh: nit: landing here after reading the code snippet I find the "After" a bit confusing, what do you think about phrasing this something like "An instance of a NativeAdFactory should be passed to..."

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,95 @@ RewardedVideoAd.instance.listener =
...
+you are using the Embedding V1 or Embedding V2. If you're using the Embedding V1, you're `MainActivity.java`
```
amirh: nit: I'd avoid saying your main activity *should* look like this, instead I'd suggest explaining what has to happen, and then showing an example.

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,95 @@ RewardedVideoAd.instance.listener =
...
+import io.flutter.plugins.firebaseadmob.FirebaseAdMobPlugin;
+
+public class MainActivity extends FlutterActivity {
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    GeneratedPluginRegistrant.registerWith(this);
+
+    final FirebaseAdMobPlugin adMobPlugin =
+        valuePublishedByPlugin("io.flutter.plugins.firebaseadmob.FirebaseAdMobPlugin");
```
amirh: How about making a static method on FirebaseAdMobPlugin to take care of the registration?

So this would look like:
```java
FirebaseAdMobPlugin.registerNativeAdFactory(this, <potential other params>);
```

We could overload this method for the v2 use case as well, so for V2 it looks like:

```java
FirebaseAdMobPlugin.registerNativeAdFactory(flutterEngine, <potential other params>);
```

This has the benefit of keeping the implementation details of how to fetch the plugin (e.g what is the plugin's class name) as part of the plugin's implementation.

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,95 @@ RewardedVideoAd.instance.listener =
...
+
+import com.google.android.gms.ads.formats.UnifiedNativeAd;
+import com.google.android.gms.ads.formats.UnifiedNativeAdView;
+import io.flutter.plugins.firebaseadmob.FirebaseAdMobPlugin.NativeAdFactory;
+import java.util.Map;
+
+class NativeAdFactoryExample implements NativeAdFactory {
+  @Override
+  public UnifiedNativeAdView createNativeAd(
+      UnifiedNativeAd nativeAd, Map<String, Object> customOptions) {
```
amirh: What's your thoughts on allowing multiple factories to be registered with a name (key). And then on the Dart side users can specify the factory to use for a given ad?

While the same can be achieved through customOptions (essentially passing a factory "key" as an option and changing the rendering logic accordingly) I wonder if it will be more convenient.

Another scenario to think about is having 2 packages that depend on the admob plugin being used by the same app, if we only allow a single factory their factories will conflict and override each other.

amirh:
> One thing I'm a little hesitant about is designing this around the assumption that the current activity implements some interface specific for this plugin which defines the looks of native ads. I can imagine an add2app scenario with multiple activities that are all using the same Flutter fragment which will require some hairy plumbing to get this properly working. (More abstractly, it feels like a concern that does not belong in the Activity class.)
> 
> I'd love for us to use a more implicit "give me the admob plugin for this engine" on the Java side, and then register a native ads factory on it.
> 
> With the v1 embedding we could use PluginRegistry#publish for the AdMob plugin to publish an object that accepts registrations of native view factories, and then the developer will use PluginRegistry#valuePublishedBy to fetch that object and register their native ads factory.
> 
> AFAIK we do not support something like that in the v2 embedding, I think we should add a similar mechanism to the v2 embedding.
> 
> @matthew-carroll @xster - thoughts?

Maurice pointed out that this is doable with the current v2 implementation through `getPlugins` 🎉

XSiyabonga:
> can fork and use this branch before its merged, but at your own risk. This code doesn't have tests and I will be making changes to it at some point tha

Sorry if this is a stupid question but how would I pull this branch exactly
dependencies:   
     kittens:
       git:
         url: git://github.com/munificent/kittens.git????
         ref: some-branch???

packages/firebase_admob/README.md
bparrishMines: The example of implementing a `UnifiedNativeAdView` requires a `LayoutInflater`, which you would get from an `Activity`. I assumed it would be ok to recommend the path of least resistance.

packages/firebase_admob/README.md
bparrishMines: I assumed that the move to the v2 embedding included avoiding the model of static plugins. I assumed we wanted to make everything about a plugin to be an instance. In an add-to-app scenario, what if we have multiple `FlutterView`s that want to use their own `NativeAdFactory`s.

packages/firebase_admob/README.md
bparrishMines: For a plugin to set a `NativeAdFactory`, it would need access to the `FlutterEngine.getPlugins()`. This access is now deprecated for `FlutterPluginBinding`. 

A Flutter user would have to set this in their `MainActivity.configureFlutterEngine`, so I can only think of a scenario where two separate plugins have a public Java `NativeAdFactory`. Then, the user would register both in `MainActivity`. Do you think this would be common enough to need to register based on key?

packages/firebase_admob/README.md
amirh: Not sure I follow, the v2 version of this method registers for a given `FlutterEngine`... if you have 2 FlutterEngine you could still register a separate factory for each engine...

packages/firebase_admob/README.md
amirh: f2f discussion: we believe cross-plugin interaction on the platform side is useful, and we will want to add it on the mid-term. It doesn't seem to cost much to design this to be ready for that day, so we might as well do it now and prevent a future public API change.

packages/firebase_admob/README.md
```diff
@@ -182,6 +182,104 @@ RewardedVideoAd.instance.listener =
...
+}
+```
+
+When creating the `NativeAd` in Flutter, the `factoryId` parameter should match the one you used to
+add the factory to `FirebaseAdMobPlugin`.
+
+An example of displaying a `UnifiedNativeAd` with a `UnifiedNativeAdView` can be found
+[here](https://developers.google.com/admob/android/native/advanced). The example app also inflates
+a custom layout and displays the test Native ad.
+
```
amirh: Would be useful to add some explanation about how to use the Dart-side API

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
```diff
@@ -226,4 +248,64 @@ void show() {
...
+              .withNativeAdOptions(new NativeAdOptions.Builder().build())
+              .build();
+
+      AdRequestBuilderFactory factory = new AdRequestBuilderFactory(targetingInfo);
+      adLoader.loadAd(factory.createAdRequestBuilder().build());
+    }
+
+    @Override
+    void show() {
+      if (status == Status.LOADING) {
```
amirh: Seems like this is already done in showAdView

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
```diff
@@ -226,4 +248,64 @@ void show() {
...
+        status = Status.PENDING;
+        return;
+      }
+
+      showAdView(nativeAdFactory.createNativeAd(nativeAd, customOptions));
+    }
+
+    @Override
+    void dispose() {
+      super.dispose();
```
amirh: Is this the same code in Banner's dispose? maybe move to a common method?

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
amirh: unrelated to this PR, just wondering - are we / should we be protecting against multiple calls to show?

packages/firebase_admob/README.md
bparrishMines: Oh, were u suggesting that we have a static method only for the v1 embedding?

bparrishMines: clarified f2f. done

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/FirebaseAdMobPlugin.java
```diff
@@ -34,6 +38,28 @@
   // This is always null when not using v2 embedding.
   private FlutterPluginBinding pluginBinding;
   private RewardedVideoAdWrapper rewardedWrapper;
+  private final Map<String, NativeAdFactory> nativeAdFactories = new HashMap<>();
+
+  /**
+   * Interface used to display a {@link com.google.android.gms.ads.formats.UnifiedNativeAd}.
+   *
+   * <p>Creates {@link com.google.android.gms.ads.formats.UnifiedNativeAdView}s from Flutter Native
+   * Ads when added to a {@link io.flutter.plugins.firebaseadmob.FirebaseAdMobPlugin}.
```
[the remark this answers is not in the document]
bparrishMines: I changed it to `from a Native Ad created in Dart`.

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
bparrishMines: Yes these lines should protect against multiple shows

```
if (activity.findViewById(id) == null) {
  // Add view to activity and set id to id
}
```

bparrishMines: I created an abstract class called `MobileAdWithView` that `Banner` and `NativeAd` extend.

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
bparrishMines: I created an abstract class called `MobileAdWithView` that `Banner` and `NativeAd` extend.

packages/firebase_admob/README.md
bparrishMines: done

packages/firebase_admob/example/android/app/src/main/java/io/flutter/plugins/firebaseadmobexample/EmbeddingV1Activity.java
```diff
@@ -7,11 +7,15 @@
...
+import io.flutter.plugins.firebaseadmob.FirebaseAdMobPlugin;
+import io.flutter.plugins.firebaseadmob.FirebaseAdMobPlugin.NativeAdFactory;
 
 public class EmbeddingV1Activity extends FlutterActivity {
   @Override
   protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     GeneratedPluginRegistrant.registerWith(this);
+    final NativeAdFactory factory = new NativeAdFactoryExample(getLayoutInflater());
+    FirebaseAdMobPlugin.registerNativeAdFactory(this, "adFactoryExample", factory);
```
amirh: We should probably call unregisterNativeAdFactory in `onStop`?

packages/firebase_admob/example/android/app/src/main/java/io/flutter/plugins/firebaseadmobexample/MainActivity.java
```diff
@@ -8,12 +8,16 @@
...
 
 public class MainActivity extends FlutterActivity {
   // TODO(bparrishMines): Remove this once v2 of GeneratedPluginRegistrant rolls to stable. https://github.com/flutter/flutter/issues/42694
   @Override
   public void configureFlutterEngine(FlutterEngine flutterEngine) {
     flutterEngine.getPlugins().add(new E2EPlugin());
     flutterEngine.getPlugins().add(new FirebaseAdMobPlugin());
+
+    final NativeAdFactory factory = new NativeAdFactoryExample(getLayoutInflater());
+    FirebaseAdMobPlugin.registerNativeAdFactory(flutterEngine, "adFactoryExample", factory);
```
amirh: We should probably call unregisterNativeAdFactory on `cleanUpFlutterEngine`.

For this and the previous, we'd probably want to update the examples in README as well.

bparrishMines: Do we need to do that? I thought cleanup such as that isn't necessary for the V2 Embedding. (e.g. calling `MethodChannel.setMethodCallHandler(null)`). `FirebaseAdmobPlugin` doesn't maintain any static references, so the reference to the `NativeAdFactory` would be destroyed along with the `FirebaseAdmobPlugin` instance.

packages/firebase_admob/example/android/app/src/main/java/io/flutter/plugins/firebaseadmobexample/EmbeddingV1Activity.java
bparrishMines: According to [android documentation](https://developer.android.com/guide/components/activities/activity-lifecycle#onstop) for `onStop`, it is called when the app is no longer visible. For example, a would answer a call and then return to the app. This would call `onStop`, then `onRestart`.  NativeAds would no longer work, but the app would be visible. 

It also states: "You should also use onStop() to perform relatively CPU-intensive shutdown operations." I don't believe the NativeAdFactory qualifies for this.

packages/firebase_admob/android/src/main/java/io/flutter/plugins/firebaseadmob/MobileAd.java
```diff
@@ -226,4 +263,47 @@ void show() {
...
+                    }
+                  })
+              .withAdListener(this)
+              .withNativeAdOptions(new NativeAdOptions.Builder().build())
+              .build();
+
+      AdRequestBuilderFactory factory = new AdRequestBuilderFactory(targetingInfo);
+      adLoader.loadAd(factory.createAdRequestBuilder().build());
+    }
+
```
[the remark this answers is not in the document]
bparrishMines: I'm not actually sure. It's not clear in the [AdView](https://developers.google.com/android/reference/com/google/android/gms/ads/AdView) documentation about when to call these. The class has `resume()` and `pause()` methods too. Their documentation only states that they should be called in their respective `Activity` method. (e.g. `adview.pause()` called in `activity.onPause()`)

However, neither Banner, Interstitial, or Native ads examples talk about using these.
https://developers.google.com/admob/android/banner.

Either way, I copied this from the `MobileAd` class and don't think the issue of ad lifecycle should be addressed in this PR. I could add a TODO though?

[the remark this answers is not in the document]
amirh: offline discussion: sounds like we do not understand whether we should be forwarding activity lifecycle calls to the adviews, and we're currently partially doing so. Let's file an issue to follow-up on that.

PR #237, opened 2019-09-29
[firebase_auth] Fix NoSuchMethodError exception in `reauthenticateWithCredential`

fabryx92:
Fix NoSuchMethodError exception when calling FirebaseUser.reauthenticateWithCredential

## Description

The stacktrace of the issue was this:
```
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:51:5)
#1      new AuthResult._ (package:firebase_auth/src/auth_result.dart:12:36)
#2      FirebaseUser.reauthenticateWithCredential (package:firebase_auth/src/firebase_user.dart:213:23)
<asynchronous suspension>
#3      ...............
```

This is the relevant part in `auth_result.dart`:
```dart
class AuthResult {
  AuthResult._(this._data, FirebaseApp app)
      : user = FirebaseUser._(_data['user'].cast<String, dynamic>(), app);

  final Map<String, dynamic> _data;
```

So, the `data` parameter passed to the `FirebaseUser` constructor was `null`.

From the `reauthenticateWithCredential` method:
```dart
Future<AuthResult> reauthenticateWithCredential(
      AuthCredential credential) async {
    assert(credential != null);
    final Map<String, dynamic> data =
        await FirebaseAuth.channel.invokeMapMethod<String, dynamic>(
      'reauthenticateWithCredential',
      <String, dynamic>{
        'app': _app.name,
        'provider': credential._provider,
        'data': credential._data,
      },
    );
    return AuthResult._(data, _app);
  }
```
the `data` parameter passed to `AuhtResult` came from the `reauthenticateWithCredential` invoked channel method.

By looking inside the `handleReauthenticateWithCredential` method in the `FirebaseAuthPlugin.java` I notice the line:
```java
currentUser
        .reauthenticate(credential)
        .addOnCompleteListener(new TaskVoidCompleteListener(result));
```

The `reauthenticate` has a return type of `Task<void>`, for this reason, the `data` in the `reauthenticateWithCredential` was filled with null.
The `FirebaseUser`'s `reauthenticateAndRetrieveData` method has `Task<AuthResult>` as return type. Also the `SignInCompleteListener` returns a map containing the Firebase user data inside the "user" key (exactly what the `AuthResult` needs in its constructor).
By applying these two modifications the code works properly.

## Related Issues

Fixes https://github.com/FirebaseExtended/flutterfire/issues/76
Fixes https://github.com/FirebaseExtended/flutterfire/issues/1085

## Checklist

- [x] I signed the [CLA].

## Breaking Change


fabryx92:
I can't figure out why the build fails

hungnt-kayac:
@fabryx92 i think you also have to fix ios plugin.
in that case, i've took a look at 
https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m#L238
and found that it didn't set return object.
in my opinion, i think it should be
```
completion:^(FIRAuthDataResult *r, NSError *error) {
    [self sendResult:result forAuthDataResult:r error:error];
}
```

fabryx92:
@hungnt-kayac thank you! You are right, I focused only on the Android part. I'm not very familiar with ios development, maybe because of this, I forgot it. Your reply was really helpful!

packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
```diff
@@ -239,8 +239,8 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
     [[self getAuth:call.arguments].currentUser
         reauthenticateAndRetrieveDataWithCredential:[self getCredential:call.arguments]
                                          completion:^(FIRAuthDataResult *r,
-                                                      NSError *_Nullable error) {
-                                           [self sendResult:result forObject:nil error:error];
+                                                      NSError *error) {
+                                           [self sendResult:result forAuthDataResult:nil error:error];
```
hungnt-kayac: i should be `forAuthDataResult:r` with `r` is `FIRAuthDataResult `

fabryx92: Oh, I didn't notice that. Thank you!

packages/firebase_auth/android/src/main/java/io/flutter/plugins/firebaseauth/FirebaseAuthPlugin.java
```diff
@@ -491,8 +491,8 @@ private void handleReauthenticateWithCredential(
     AuthCredential credential = getCredential((Map<String, Object>) call.arguments());
 
     currentUser
-        .reauthenticate(credential)
-        .addOnCompleteListener(new TaskVoidCompleteListener(result));
+        .reauthenticateAndRetrieveData(credential)
```
BugsBunnyBR: How about to change the method name in the dart/android/ios side to keep the real representation of what is being called? instead of `reauthenticateWithCredential` => `reauthenticateAndRetrieveData` ?

fabryx92: Maybe it could be used the name already used in the ios side: `reauthenticateAndRetrieveDataWithCredential `

BugsBunnyBR: Whatever you think is the best.

fabryx92: By looking better at the code I think that your proposal is better than mine. You are right, it's better to adhere to what is called from the Firebase Auth native library.

BugsBunnyBR: Yes, it would be a breaking change.
Since the code is not working today, I wouldn't mind this breaking change... but I think @collinjackson or @kroikie are better suited to decide what is pertinent for this PR.

packages/firebase_auth/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.14.0+7
+* Fix NoSuchMethodError exception in `reauthenticateWithCredential`
```
collinjackson:
```suggestion

* Fix NoSuchMethodError exception in `reauthenticateWithCredential`
```

packages/firebase_auth/android/src/main/java/io/flutter/plugins/firebaseauth/FirebaseAuthPlugin.java
collinjackson: In newer versions of the Firebase iOS Auth SDK,`reauthenticateWithCredential` and `reauthenticateAndRetrieveDataWithCredential ` both do the same thing and return a `FIRAuthDataResultCallback`. I think we should go with the non-deprecated `reauthenticateWithCredential` which is what this PR is already doing, and no name change is needed.

Reference: 
 https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRUser#-reauthenticatewithcredential:completion:

fabryx92: @collinjackson can you help me in this decision? Should I left all the names as they are or should I change them to make them more coherent with the real method called, making it a breaking change?

packages/firebase_auth/CHANGELOG.md
creativecreatorormaybenot: Maybe I should not comment here, but `NoSuchMethodError` exception is somewhat  contradictory as [exceptions and errors are not the same in Dart](https://stackoverflow.com/a/58361954/6509751).

fabryx92:
Why now the build fails? Editing the changelog only? I did again all the steps written in the Contribution guide to be sure that all is fine, and it is.

PR #1266, opened 2019-10-16
[firebase_analytics] Support Android v2 embedding

kroikie:
Fixes: https://github.com/flutter/flutter/issues/41867


packages/firebase_analytics/android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java
```diff
@@ -16,20 +22,55 @@
...
+    FirebaseApp.initializeApp(applicationContext);
+    firebaseAnalytics = FirebaseAnalytics.getInstance(applicationContext);
+    methodChannel = new MethodChannel(binaryMessenger, "plugins.flutter.io/firebase_analytics");
+    methodChannel.setMethodCallHandler(this);
+  }
+
+  @Override
+  public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
+    firebaseAnalytics = null;
+    setActivity(null);
```
matthew-carroll: I'm not sure that this line will ever have an effect. Activity should always be null by the time this runs.

packages/firebase_analytics/android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java
```diff
@@ -62,6 +103,10 @@ public void onMethodCall(MethodCall call, Result result) {
     }
   }
 
+  private void setActivity(Activity activity) {
+    this.activity = activity;
```
matthew-carroll: Where is this activity being used?

packages/firebase_analytics/android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java
```diff
@@ -16,20 +22,55 @@
 import java.util.Map;
 
 /** Flutter plugin for Firebase Analytics. */
-public class FirebaseAnalyticsPlugin implements MethodCallHandler {
-  private final PluginRegistry.Registrar registrar;
-  private final FirebaseAnalytics firebaseAnalytics;
+public class FirebaseAnalyticsPlugin implements MethodCallHandler, FlutterPlugin, ActivityAware {
```
matthew-carroll: Have you run the new implementation manually to verify that it works as expected?

kroikie: Yes I have run it and it work as expected, the example app that is.

packages/firebase_analytics/android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java
kroikie: There is a event that logs the current screen `handleSetCurrentScreen`, the native SDK requires you pass in the current activity.

packages/firebase_analytics/android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java
kroikie: Good to know, would it also already be null for the other `detach` callbacks like `onDetachedFromActivityForConfigChanges` etc?

packages/firebase_analytics/example/android/app/src/main/java/io/flutter/plugins/firebaseanalyticsexample/MainActivity.java
```diff
@@ -4,14 +4,13 @@
...
+import io.flutter.embedding.engine.FlutterEngine;
+import io.flutter.plugins.firebaseanalytics.FirebaseAnalyticsPlugin;
 
 public class MainActivity extends FlutterActivity {
   @Override
-  protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
-    GeneratedPluginRegistrant.registerWith(this);
+  public void configureFlutterEngine(FlutterEngine flutterEngine) {
+    flutterEngine.getPlugins().add(new FirebaseAnalyticsPlugin());
```
collinjackson: You'll need to add the E2EPlugin here as well, otherwise the integration tests won't run for the new embedding.

packages/firebase_analytics/android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java
kroikie: Removed activity reset from "Detach" callbacks.

packages/firebase_analytics/example/android/app/src/main/java/io/flutter/plugins/firebaseanalyticsexample/MainActivity.java
kroikie: Done

PR #1291, opened 2019-10-21
Cast error.code to long to avoid using NSInteger as %ld format warnings.

mehmetf:


mehmetf:
@collinjackson could you LGTM this when you get the chance? We are being blocked from rolling this plugin.

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -10,7 +10,7 @@
 static FlutterError *getFlutterError(NSError *error) {
   if (error == nil) return nil;
 
-  return [FlutterError errorWithCode:[NSString stringWithFormat:@"Error %ld", error.code]
+  return [FlutterError errorWithCode:[NSString stringWithFormat:@"Error %ld", (long)error.code]
```
cyanglaz: LGTM!
Another way instead of force casting would be always use NSObjects
such as 
```suggestion
  return [FlutterError errorWithCode:[NSString stringWithFormat:@"Error %@", @(error.code)]
```

But I'm ok with whichever you prefer.

mehmetf: Thanks! Good to know!

PR #1292, opened 2019-10-21
Update plugin dependencies in android/build.gradle

kroikie:
Resolve build issues with firebase_analytics plugin.


collinjackson:
Seems like the build is still failing on stable.

kroikie:
I think the analytics_plugin is now building successfully on master and stable. Abandoning this PR will make another if necessary.

Issue #617, opened 2019-10-13
FieldPath


kroikie:
i think  FieldPath is missing in flutter firrestore, so please i think its an important thing that should not be left out the current stage of flutter. thanks

kroikie:
@rayjadore

The issue at https://github.com/flutter/flutter/issues/22849 has been closed and moved here. Future collaboration on this issue will be done here.

BondarenkoStas:
https://stackoverflow.com/a/52469844

PR #1296, opened 2019-10-22
[cloud_firestore] Add FieldPath

creativecreatorormaybenot:
## Description

[This](https://github.com/FirebaseExtended/flutterfire/issues/617#issuecomment-545022185) brought my attention to `FieldPath` again. While my [solution on StackOverflow](https://stackoverflow.com/a/52469844/6509751) works, it is not particularly nice to use.

## Related Issues

 * #617

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/cloud_firestore/lib/src/field_path.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017, the Chromium project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of cloud_firestore;
+
+/// A [FieldPath] refers to a field in a document.
+class FieldPath {
+  /// The path to the document id, which can be used in queries.
+  static String get documentId => '__name__';
```
collinjackson: This should return a `FieldPath` rather than a `String` to match the native API.

You can follow the pattern established for `FieldValue`: https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_firestore/lib/src/field_value.dart

packages/cloud_firestore/example/android/gradle.properties
```diff
@@ -1 +1,2 @@
 org.gradle.jvmargs=-Xmx1536M
+android.enableR8=true
```
collinjackson: Do we need this?

packages/cloud_firestore/lib/src/field_path.dart
creativecreatorormaybenot: This is how I structured the pull request at first, however, I noticed that doing so would require an awful amount of changes, i.e. every query `String` parameter would need to be converted to `dynamic` and all the methods would need to convert the `FieldValue` to a `String` at some point. Another option would be to require the user to call `FieldValue.documentId.toString()`, but that is very strange usage and just a worse version of returning `String` right away. What do you think?

packages/cloud_firestore/example/android/gradle.properties
creativecreatorormaybenot: Sorry, I missed this. It seems that this automatically added by Android Studio.

packages/cloud_firestore/lib/src/field_path.dart
creativecreatorormaybenot: This was how I structured my pull request at first, however, I thought that there were too many changes for this simple addition, but I can implement it.

creativecreatorormaybenot:
@collinjackson What kind of formatter are you using in `flutterfire`? 

When I use the formatter (which the IDE does automatically 😞), the whole file is changed and the Cirrus CI test also fails.. It is very cumbersome for me to manually adjust the format.

collinjackson:
> @collinjackson What kind of formatter are you using in `flutterfire`?
> 
> When I use the formatter (which the IDE does automatically 😞), the whole file is changed and the Cirrus CI test also fails.. It is very cumbersome for me to manually adjust the format.

We're using `pub global run flutter_plugin_tools format`. That's all I know...

Pro tip: The failure message gives you a patch command that will fix the formatting for you:

```
patch -p1 <<DONE
diff --git a/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java b/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
index f231617..524589d 100644
--- a/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
+++ b/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
@@ -133,14 +133,15 @@ public class CloudFirestorePlugin implements MethodCallHandler {
         if (field instanceof FieldPath) {
           if (field == FieldPath.documentId()) {
-            throw new IllegalArgumentException("You cannot use {start/end}{At/After}Document " +
-                    "when ordering by the document id. When using any of the mentioned methods, " +
-                    "the library will order by the document id implicitly as " +
-                    "it needs to add other fields to the order clause.");
+            throw new IllegalArgumentException(
+                "You cannot use {start/end}{At/After}Document "
+                    + "when ordering by the document id. When using any of the mentioned methods, "
+                    + "the library will order by the document id implicitly as "
+                    + "it needs to add other fields to the order clause.");
           } else {
             // Unsupported type.
           }
-        } else if (field instanceof  String) {
+        } else if (field instanceof String) {
           String orderByFieldName = (String) field;
           if (orderByFieldName.contains(".")) {
             String[] fieldNameParts = orderByFieldName.split("\\.");
@@ -313,7 +314,7 @@ public class CloudFirestorePlugin implements MethodCallHandler {
       boolean descending = (boolean) order.get(1);
       Query.Direction direction =
-              descending ? Query.Direction.DESCENDING : Query.Direction.ASCENDING;
+          descending ? Query.Direction.DESCENDING : Query.Direction.ASCENDING;
       if (fieldName != null) {
         query = query.orderBy(fieldName, direction);
diff --git a/packages/cloud_firestore/lib/src/field_path.dart b/packages/cloud_firestore/lib/src/field_path.dart
index e144cd0..64cdde8 100644
--- a/packages/cloud_firestore/lib/src/field_path.dart
+++ b/packages/cloud_firestore/lib/src/field_path.dart
@@ -16,5 +16,6 @@ class FieldPath {
   final _FieldPathType type;
   /// The path to the document id, which can be used in queries.
-  static FieldPath get documentId => const FieldPath._(_FieldPathType.documentId);
+  static FieldPath get documentId =>
+      const FieldPath._(_FieldPathType.documentId);
 }
diff --git a/packages/cloud_firestore/lib/src/firestore_message_codec.dart b/packages/cloud_firestore/lib/src/firestore_message_codec.dart
index 1a555fc..9f658cd 100644
--- a/packages/cloud_firestore/lib/src/firestore_message_codec.dart
+++ b/packages/cloud_firestore/lib/src/firestore_message_codec.dart
@@ -34,7 +34,7 @@ class FirestoreMessageCodec extends StandardMessageCodec {
   static const Map<_FieldPathType, int> _kFieldPathCodes =
       <_FieldPathType, int>{
     _FieldPathType.documentId: _kDocumentId,
-      };
+  };
   @override
   void writeValue(WriteBuffer buffer, dynamic value) {
diff --git a/packages/cloud_firestore/test/cloud_firestore_test.dart b/packages/cloud_firestore/test/cloud_firestore_test.dart
index d437202..63a3f06 100755
--- a/packages/cloud_firestore/test/cloud_firestore_test.dart
+++ b/packages/cloud_firestore/test/cloud_firestore_test.dart
@@ -1264,7 +1264,8 @@ bool _deepEquals(dynamic valueA, dynamic valueB) {
   if (valueA is FieldValue) {
     return valueB is FieldValue && _deepEqualsFieldValue(valueA, valueB);
   }
-  if (valueA is FieldPath) return valueB is FieldPath && valueA.type == valueB.type;
+  if (valueA is FieldPath)
+    return valueB is FieldPath && valueA.type == valueB.type;
   return valueA == valueB;
 }
DONE
```

creativecreatorormaybenot:
@collinjackson Thanks a bunch!

creativecreatorormaybenot:
Unfortunately, I had to do the formatting manually again (see https://github.com/flutter/flutter/issues/43364).

However, I think that I am done with the new implementation. I added some exceptions that will help users understand what about their queries was incorrect and most importantly added assertions when possible.

packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
```diff
@@ -129,16 +129,32 @@ private Source getSource(Map<String, Object> arguments) {
...
-          for (int i = 1; i < fieldNameParts.length - 1; i++) {
-            current = (Map<String, Object>) current.get(fieldNameParts[i]);
+        final Object field = order.get(0);
+
+        if (field instanceof FieldPath) {
+          if (field == FieldPath.documentId()) {
+            // This is also checked by an assertion on the Dart side.
+            throw new IllegalArgumentException(
+                "You cannot order by the document id when using"
+                    + "{start/end}{At/After/Before}Document a the library will order by the document id"
```
BugsBunnyBR: It looks like there is some typo here, mind to check?

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -31,17 +31,36 @@
...
+      NSObject *field = orderByParameters[0];
+
+      if ([field isKindOfClass:[FIRFieldPath class]]) {
+        if ([field isEqual:FIRFieldPath.documentID]) {
+          // This is also checked by an assertion on the Dart side.
+          [NSException
+               raise:@"Invalid use of FieldValue.documentId"
+              format:
+                  @"You cannot order by the document id when using"
+                   "{start/end}{At/After/Before}Document a the library will order by the document"
```
BugsBunnyBR: here too.

packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
creativecreatorormaybenot: thanks 👍

packages/cloud_firestore/lib/src/field_path.dart
collinjackson: Thank you!

PR #1349, opened 2019-11-01
Run the publish with the pub version from flutter stable

amirh:
Copy of https://github.com/flutter/plugins/pull/2250

I think we probably shouldn't need to copy such changes across repos, filed https://github.com/flutter/flutter/issues/44000 to track.


amirh:
Will land once CI is green

PR #1351, opened 2019-11-01
Run publish check on stable

bparrishMines:
https://github.com/flutter/plugins/pull/2250


creativecreatorormaybenot:
Does this make #1354 obsolete? I guess merging it would still help while the plugin is not fixed, but will integration tests catch this in the future?

PR #1372, opened 2019-11-05
[firebase_dynamic_links] support v2 embedding

bparrishMines:
## Description

Fixes https://github.com/flutter/flutter/issues/41872


packages/firebase_dynamic_links/android/build.gradle
```diff
@@ -34,8 +34,33 @@ android {
...
+                containsEmbeddingDependencies = true
+                break
+            }
+        }
+    }
+    if (!containsEmbeddingDependencies) {
+        android {
+            dependencies {
+                def lifecycle_version = "1.1.1"
+                compileOnly "android.arch.lifecycle:runtime:$lifecycle_version"
```
xster: are there existing androidx transitive dependencies? (since there's an androidx annotation already) Is this a downgrade?

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
```diff
@@ -27,15 +31,30 @@
...
-  private final MethodChannel channel;
+public class FirebaseDynamicLinksPlugin
+    implements FlutterPlugin, ActivityAware, MethodCallHandler, NewIntentListener {
+  private Registrar registrar;
+  private MethodChannel channel;
+  private ActivityPluginBinding activityBinding;
 
   private FirebaseDynamicLinksPlugin(Registrar registrar, MethodChannel channel) {
     this.registrar = registrar;
     this.channel = channel;
```
xster: You don't have to pass this in right? The less potentially dangling nulls there are (depending on the path you take), the less likely there will be accidental null pointer exceptions

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
```diff
@@ -27,15 +31,30 @@
...
-public class FirebaseDynamicLinksPlugin implements MethodCallHandler, NewIntentListener {
-  private final Registrar registrar;
-  private final MethodChannel channel;
+public class FirebaseDynamicLinksPlugin
+    implements FlutterPlugin, ActivityAware, MethodCallHandler, NewIntentListener {
+  private Registrar registrar;
+  private MethodChannel channel;
+  private ActivityPluginBinding activityBinding;
 
   private FirebaseDynamicLinksPlugin(Registrar registrar, MethodChannel channel) {
```
xster: You added annotations right? Tag Registrar nonnull. Even better, since you only pull the activity out of the registrar, just change the instance variable to activity so future maintainers don't accidentally do registrar.messenger() and expand the encapsulation in the future and trip on a null half of the time.

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
```diff
@@ -138,19 +183,22 @@ public void onSuccess(PendingDynamicLinkData pendingDynamicLinkData) {
...
   }
 
+  private Activity getActivity() {
+    return registrar != null ? registrar.activity() : activityBinding.getActivity();
+  }
+
   private OnCompleteListener<ShortDynamicLink> createShortLinkListener(final Result result) {
     return new OnCompleteListener<ShortDynamicLink>() {
       @Override
-      public void onComplete(@NonNull Task<ShortDynamicLink> task) {
```
xster: Were these causing issues?

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
```diff
@@ -4,9 +4,9 @@
 
 package io.flutter.plugins.firebasedynamiclinks;
 
+import android.app.Activity;
 import android.content.Intent;
 import android.net.Uri;
-import androidx.annotation.NonNull;
```
xster: You went from androidx to android.arch in the gradle, which seems to cause this to not work. Though I'm not sure if it's intentional. Or did firebase already conceptually "upgrade" to androidx @mklim

xster:
LGTM

packages/firebase_dynamic_links/android/build.gradle
bparrishMines: These use the latest instruction from the migration wiki: https://github.com/flutter/flutter/wiki/Experimental:-Create-Flutter-Plugin.

Quoting: https://github.com/flutter/plugins/pull/2221#issue-329949430
`We introduced a constraint that the app uses the plugin has to be migrated to androidx.
This patch removes the constraint.`

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
bparrishMines: I should need it. The FirebaseDynamicLinks plugins makes callbacks to dart: https://github.com/FirebaseExtended/flutterfire/pull/1372/files#diff-3f6c424218a2662c8ddc4b322f30bf66R81. 

I could hold on to a `BinaryMessenger` and instantiate a `MethodChannel` every time we want to call `invokeMethod`, but I think this looks cleaner.

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
bparrishMines: I removed the annotations in this PR because the ecosystem team decided that we won't include androidx in any plugin that doesn't require it. It's so that a Flutter app that doesn't include androidx can use this plugin. 

However, this is a good point about someone accidently using the registrar. What about adding the `@Deprecated` annotation with an explanation about registrar remaining until we decide to only support v2 embedding flutter apps.

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
bparrishMines: answered in https://github.com/FirebaseExtended/flutterfire/pull/1372#discussion_r345961234

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
bparrishMines: answered in https://github.com/FirebaseExtended/flutterfire/pull/1372#discussion_r345961234

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
bparrishMines: This does remind me that I also need to add documentation for the plugin now. Updated PR

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
mklim: The latest thing we've talked about here is:

1. android.arch for the magic lifecycle dependencies in the Gradle script. That way Jetifier can run on the artifacts in cases where it's needed, and it will resolve as-is for unmigrated apps.
2. `annotation` should just be removed from both Gradle and source code files entirely. If it's present in source code Jetifier won't be able to run on it and it will cause compile failures for when there's AndroidX incompatibilities. I don't think that's worth just having the annotations.

If this is being stripped from the source code but it's still in the Gradle files as `android.arch.annotation` it should probably be removed from the Gradle files too. If it's not in the Gradle files at all this makes sense to me and shouldn't be causing known issues.

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
mklim: Sorry, this tab was open and I missed the new comments. I agree with Maurice's summary about annotation, commented [here](https://github.com/FirebaseExtended/flutterfire/pull/1372#discussion_r345966532).

For `@Deprecated`, I vaguely remember floating the idea by @amirh offline and he was against it. The reason being that for now the static registrar registerWith method _is_ fully supported, and we don't want to flag it as deprecated before we're actually really deprecating it. I think those uses will show up as compile time warnings too, so probably better not to bring them in.

packages/firebase_dynamic_links/android/build.gradle
xster: @mklim @blasten we had this conversation separately too. I guess we're saying mixing is ok and we're just letting plugins build as aars as fallbacks?

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
xster: We don't want to `@Deprecate` the actual Registrar API in the engine like @mklim said. 
If you meant `@Deprecate` the instance variable, that sounds ok. I don't have a strong opinion. I was just making a general comment (regardless of what the class is) to minimize access. If your plugin instance wants an activity, just keep an activity and not its provider.

bparrishMines: I agree. I remember that we originally kept an instance of `Registrar` and not `Activity` because the method `Registrar.getActivity()` would sometimes return different activities in add-to-app scenarios. 

Since we're now migrating to v2 embedding, I don't this this would be the case anymore. I updated the plugin to hold a reference to `Activity`.

xster:
LGTM

PR #1392, opened 2019-11-08
Remove @visibleForTesting from FirebaseApp constructor

kroikie:
## Description

Remove @visibleForTesting annotation from FirebaseApp constructor. This will resolve CI issues while a new FirebaseApp creation strategy is WIP.


collinjackson:
Also I think we should remove the const from the constructor since we might want to do runtime checks in the future.

PR #1404, opened 2019-11-11
[CI] Use available simulator for CI

kroikie:
## Description

Current iOS simulator is not available in CI. This PR updates the selected simulator to one that is available in CI.


hterkelsen:
It seems like this works in this PR because on Cirrus it "auto-upgraded" the Mac OS image to `mojave-xcode-10.2-flutter`. However, on my PR it didn't auto-upgrade and the iOS tests fail because this simulator isn't there. Perhaps we should use the new image in our `cirrus.yml`?

PR #1426, opened 2019-11-14
[CI] Add version to example apps

kroikie:
## Description

Add version to 3 apps that were causing [CI to fail](https://cirrus-ci.com/build/5115422860378112). It seems like other apps were not failing because the versions were hard coded instead of using the Flutter placeholders `$(FLUTTER_BUILD_NAME)` and `$(FLUTTER_BUILD_NUMBER)`. This PR fixes the build failure a future CI could update the other examples to return to the Flutter placeholders.


kroikie:
I removed the version from the pubspec.yaml but that means we must now hardcode the version into the iOS app's Info.plist file.

PR #1451, opened 2019-11-18
[firebase_auth] Fix reauthenticateWithCredential

collinjackson:
Fixes `reauthenticateWithCredential`.


collinjackson:
It looks like this is fixed by https://github.com/FirebaseExtended/flutterfire/pull/237 so I'm going to move the integration test into that PR and land that one.

PR #1453, opened 2019-11-18
[cloud_firestore] [firebase_performance] fix analyzer warnings

collinjackson:
Analyzer is complaining about fields that are unused. We can safely remove them.


kroikie:
I expect the iOS driver tests will fail here, they are being skipped in https://github.com/FirebaseExtended/flutterfire/pull/1452

collinjackson:
Landing on red to fix build

PR #1473, opened 2019-11-21
Update CONTRIBUTING.md to allow mockito

collinjackson:
Removes obsolete recommendations about removing mockito (which is now used in testing federated plugins).

## Related Issues

https://github.com/flutter/flutter/issues/34284


hterkelsen:
LGTM

PR #1481, opened 2019-11-22
[firebase_messaging] Separate onLaunch to an specific method to be ab…

jherencia:
Separate onLaunch to an specific method to be able to retrieve launch message in a synchronous way

## Description

When an application is closed an a user opens it by clicking a push notification we can retrieve the message in a synchronous way but the current architecture expects this to be asynchronous which may conflict with other processes run in parallel. This behaviour forces programmers to synchronize those processes and with the added problem that on of the async calls (onLaunch) is not called when the user opens the app in a normal way, so it is needed to add timeouts to limit the time we wait the callback.

Android and iOS let programmers access how the app has been opened in a synchronous way for Dynamic Links, Push notifications, etc. So this PR (and https://github.com/flutter/plugins/pull/1687) fixes this and handles this two situations in a consistent way.

## Checklist

- [x] I signed the [CLA].
- [ x I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md


jherencia:
```
  firebase_messaging:
    git:
      url: https://github.com/mrmilu/flutterfire.git
      path: packages/firebase_messaging
      ref: PR-1481-v1
```

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
```diff
@@ -153,8 +153,12 @@ public void onComplete(@NonNull Task<InstanceIdResult> task) {
                   channel.invokeMethod("onToken", task.getResult().getToken());
                 }
               });
+      result.success(null);
```
ened: This seems to be new behaviour, you probably added this to fix something else? Is it worth putting that change, along with a test driver test, in a separate PR?

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -137,6 +134,11 @@ class FirebaseMessaging {
     }
   }
 
+  Future<Map<dynamic, dynamic>> getLaunchMessage() async {
```
ened: Please document this method.

packages/firebase_messaging/test/firebase_messaging_test.dart
```diff
@@ -113,16 +145,11 @@ void main() {
         verify(mockChannel.setMethodCallHandler(captureAny)).captured.single;
 
     final Map<String, dynamic> onMessageMessage = <String, dynamic>{};
-    final Map<String, dynamic> onLaunchMessage = <String, dynamic>{};
     final Map<String, dynamic> onResumeMessage = <String, dynamic>{};
 
     await handler(MethodCall('onMessage', onMessageMessage));
     expect(await onMessage.future, onMessageMessage);
-    expect(onLaunch.isCompleted, isFalse);
-    expect(onResume.isCompleted, isFalse);
```
ened: I think we need to keep this one?

jherencia:
> Just a few minor comments. I wonder if we should keep `onLaunch` method (but mark it deprecated) and add the new method.
> That way it would not require a major version bump.

Thank you @ened! I marked `onLaunch` as deprecated and did all the changes you proposed.

```
  firebase_messaging:
    git:
      url: https://github.com/mrmilu/flutterfire.git
      path: packages/firebase_messaging
      ref: PR-1481-v2
```

PR #1485, opened 2019-11-23
[firebase_messaging] Fix negated @available method on iOS

mehmetf:
This is a minor fix and cleanup so that when we stop supporting iOS < 10, we can remove the code easier.


mehmetf:
This has already been fixed.

PR #1535, opened 2019-12-02
Avoid testing web plugins on stable

collinjackson:
Flutterfire equivalent for https://github.com/flutter/plugins/pull/2203

Avoids testing web plugins until support is available on the stable channel.


collinjackson:
Landing on red to fix build

Issue #349, opened 2019-10-13
[cloud_firestore] Xcode build extremely slow


kroikie:
----

**EDIT** (@Salakar): See https://github.com/FirebaseExtended/flutterfire/issues/2751 for a solution.

----

## Summary of Issue

Simply adding `cloud_firestore` to `pubspec.yaml` results in a **huge increase in iOS xcode build times**. The issue can be recreate with nothing more than a simple `flutter create` application and minor changes to `pubspec.yaml`. 

```
MacOS Mojave 10.14.6
Xcode 10.3
```

```
> flutter --version
Flutter 1.8.4-pre.91 • channel master • https://github.com/flutter/flutter.git
Framework • revision 8fdd759225 (11 hours ago) • 2019-08-04 00:56:23 -0400
Engine • revision c8f4f5d5cc
Tools • Dart 2.5.0 (build 2.5.0-dev.1.0 9bb446aae1)
```

```
> flutter create test_cloud_firestore
> cd test_cloud_firestore
```

## Without `cloud_firestore`

Just to show that it is specifically the `cloud_firestore` plugin that causes this issue I will add some other firebase plugins to `pubspec.yaml`:

```
firebase_core: ^0.4.0+1
firebase_analytics: ^4.0.2
```

Now we build the app:

```
> flutter packages get
> time flutter build ios
....
Xcode build done. 
real	0m41.014s
```

Okay that looks reasonable.

## With `cloud_firestore`

Simply add the latest version of `cloud_firestore` plugin to `pubspec.yaml` (no other changes):

```
cloud_firestore: ^0.12.9
```

Now we build the application again:

```
> flutter packages get
> time flutter build ios
....
Xcode build done. 
real	6m40.982s
```

Yes... that is right. A **six minute increase in iOS build times** just from adding the `cloud_firestore` plugin to `pubspec.yaml`. Again, no changes were made to the `flutter create` generated application, this is simply changes to `pubspec.yaml`.

## Logs

I did confirm that this issue persists on other `flutter` channels as well as other later versions of `cloud_firestore` plugin.

```
> flutter doctor -v
[✓] Flutter (Channel master, v1.8.4-pre.91, on Mac OS X 10.14.6 18G84, locale en-US)
    • Flutter version 1.8.4-pre.91 at /Users/taza/Applications/flutter
    • Framework revision 8fdd759225 (11 hours ago), 2019-08-04 00:56:23 -0400
    • Engine revision c8f4f5d5cc
[...]

PR #1563, opened 2019-12-05
[cloud_firestore] Update README.md

collinjackson:
Explains how speed up builds using incremental builds.

This is especially useful for cloud_firestore since it has a lot of code and dependencies.

## Related Issues

https://github.com/FirebaseExtended/flutterfire/issues/349

### DIFF
--- a/packages/cloud_firestore/cloud_firestore/README.md
+++ b/packages/cloud_firestore/cloud_firestore/README.md
@@ -99,7 +99,7 @@ Firestore.instance.runTransaction((Transaction tx) async {
 
 ## Improving iOS build times
 
-Cloud Firestore has dependencies on iOS that lead to slow builds (see [#349](https://github.com/FirebaseExtended/flutterfire/issues/349)). A solution to improve initial build times is in progress, but in the meantime, you can improve the performance of subsequent builds using the `cocoapods-binary` plugin.
+Cloud Firestore has dependencies on iOS that lead to slow builds (see [#349](https://github.com/FirebaseExtended/flutterfire/issues/349)). A solution is in the works to speed up initial builds (see https://github.com/firebase/firebase-ios-sdk/issues/4284), but in the meantime, you can improve the performance of subsequent builds using the `cocoapods-binary` plugin.
 
 To use, run `brew install cocoapods-binary` or `gem install cocoapods-binary` and add the following to your app's ios/Podfile.
 
--- a/packages/cloud_firestore/cloud_firestore/README.md
+++ b/packages/cloud_firestore/cloud_firestore/README.md
@@ -109,7 +109,7 @@ install! 'cocoapods',
    incremental_installation: true
 ```
 
-See https://github.com/flutter/flutter/issues/46198 for enabling incremental installation by default in the future.
+In the future, incremental installation may be enabled by default. Follow https://github.com/flutter/flutter/issues/46198 for updates.
 
 ## Getting Started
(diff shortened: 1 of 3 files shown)


packages/cloud_firestore/cloud_firestore/README.md
```diff
@@ -97,6 +97,20 @@ Firestore.instance.runTransaction((Transaction tx) async {
 });
 ```
 
+## Faster iOS builds
+
+Cloud Firestore has dependencies on iOS that lead to significant build times (see https://github.com/FirebaseExtended/flutterfire/issues/349 and https://github.com/firebase/firebase-ios-sdk/issues/4284). You can substantially reduce build times after the initial build using [incremental installation](https://medium.com/@corteggo/incremental-pod-installation-cocoapods-1-7-0-multiple-projects-302da0208af3).
+
+Add the following to your app's ios/Podfile:
```
kroikie: Could we give more context about where in the Podfile this should be added?

collinjackson: Done

MrAlek:
@collinjackson As mentioned in #349, I'm not so sure about this remedy. Incremental installation is supposed to speed up `pod install`, _not_ Xcode builds. Incremental builds work fine even with the standard Podfile.

In fact, splitting up into multiple Xcode projects could very well increase build times because of the many dependencies `cloud_firestore` has.

packages/cloud_firestore/cloud_firestore/README.md
```diff
@@ -97,6 +97,21 @@ Firestore.instance.runTransaction((Transaction tx) async {
...
+
+Cloud Firestore has dependencies on iOS that lead to significant build times (see [#349](https://github.com/FirebaseExtended/flutterfire/issues/349) and [https://github.com/firebase/firebase-ios-sdk/issues/4284](https://github.com/firebase/firebase-ios-sdk/issues/4284)). You can substantially reduce build times after the initial build using [incremental installation](https://medium.com/@corteggo/incremental-pod-installation-cocoapods-1-7-0-multiple-projects-302da0208af3).
+
+To use, ensure that `generate_multiple_pod_projects: true` and `incremental_installation: true` are included in the `install!` command in your `Podfile`. If there isn't one yet, you can use this:
+
+```
+install! 'cocoapods',
+   disable_input_output_paths: true,
+   generate_multiple_pod_projects: true,
+   incremental_installation: true
```
aaronraff: Just happened to see this PR, so I figured I would chime in with my recent experience.

After adding `incremental_installation: true` to my project, I was running into this issue https://github.com/CocoaPods/CocoaPods/issues/8604. I believe that the post_install hook may also need to be changed to address this. 

I ended up using `generated_projects` instead of `pods_project.targets`, but I admittedly have zero experience with iOS development, so there may be a better solution.

xmkevinchen: @aaronraff it's right

I've the Podfile setting like this

```ruby
install! 'cocoapods',
         :disable_input_output_paths => true,
         :generate_multiple_pod_projects => true,
         :incremental_installation => true

post_install do |installer|
  installer.generated_projects.each do |target|  ## Must use `installer.generated_projects` instead of `installer.pods_project.targets`
    target.build_configurations.each do |config|
      config.build_settings['ENABLE_BITCODE'] = 'NO'
    end
  end
end
```

PR #1585, opened 2019-12-09
[all] Migrate pubspec to platforms manifest

cyanglaz:
Remove author in the pubspec.
Use platform scheme in the pubspec.
update flutter min bound to 1.10.0.

disable CI on stable.
change publishable check to only on master.


.cirrus.yml
```diff
@@ -14,7 +14,10 @@ task:
   matrix:
     - name: publishable
       script:
-        - flutter channel stable
+        # Temporarily disabling CI on stable to allow using new Flutter features closely
+        # before a new release.
+        # TODO(cyanglaz): re-enable this
```
amirh: Add the issue link

packages/firebase_crashlytics/example/lib/main.dart
```diff
@@ -17,7 +17,7 @@ void main() {
   // Pass all uncaught errors to Crashlytics.
   FlutterError.onError = Crashlytics.instance.recordFlutterError;
 
-  runZoned<Future<void>>(() {
+  runZoned<void>(() {
```
amirh: Is this some Dart change?

bparrishMines: @kroikie was working on fixing this https://github.com/FirebaseExtended/flutterfire/pull/1589#event-2868300280.

I believe it's from the analysis changes.

amirh:
FWIW I'm not exactly sure why post-submits seems to be faster/more reliable than presubmits.
I've struggled with getting pre-submits to pass for the equivalent flutter/plugins PR, so ended up landing without presubmits passing, and waited for post-submits before publishing.

PR #1633, opened 2019-12-13
[WIP] Firestore platform interface (v2)

ditman:
## Description

Continuation of the WIP for the cloud_firestore federation; platform interface + core plugin changes.

This v2 attempts to address the concerns raised in https://github.com/FirebaseExtended/flutterfire/pull/1568. 

This is still incomplete, but I've implemented the minimal scaffolding code so I can run the firestore core plugin tests on top of the new implementation. I'm using that test output to track progress, and ensure feature parity with the old version.

### Missing things:

* Actual MethodChannel implementations for most of the functionality; for now:
```
00:01 +9 -23: Firestore Query document pagination FieldPath assertions [E]                                                                                                              
  UnimplementedError: DocumentReferencePlatform::get() is not implemented
  package:cloud_firestore_platform_interface/src/interfaces/document_reference_platform.dart 59:5  DocumentReferencePlatform.get
  package:cloud_firestore/src/document_reference.dart 77:90                                        DocumentReference.get
  cloud_firestore_test.dart 1113:67                                                                main.<fn>.<fn>.<fn>
```
* `00:02 +23 -23: Some tests failed.`
* Tests in the `cloud_firestore_platform_interface` package.
* Document all public members
* Actual dartfmt and flutter analyze

## Related Issues

* https://github.com/flutter/flutter/issues/45293

## Checklist

- [ ] If the pull request affects only one plugin, the PR title starts with the name of the plugin in brackets (e.g. [cloud_firestore])
- [ ] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [ ] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [ ] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].

## Breaking Change


packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
```diff
@@ -0,0 +1,32 @@
...
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:meta/meta.dart' show visibleForTesting;
+
+import '../interfaces/document_reference_platform.dart';
+
+class MethodChannelDocumentReference extends DocumentReferencePlatform {
+  MethodChannelDocumentReference() {
+    channel.setMethodCallHandler(_callHandler);
```
hterkelsen: it looks like you are calling `setMethodCallHandler` multiple times. I think this will cause the previous handler for that channel to be overwritten.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  @visibleForTesting
+  static const MethodChannel channel = MethodChannel(
```
hterkelsen: in the original implementation I see that the channel is initialized with `FirestoreMessageCodec`. You probably need to move `FirestoreMessageCodec` to this package and make sure the channels are initialized with it.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/types/write_batch.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+class PlatformWriteBatch {
```
hterkelsen: docs. Also I think it would be clearer if this was named `PlatformWriteBatchHandle`

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/types/query.dart
```diff
@@ -0,0 +1,27 @@
...
+  /// Causes Firestore to immediately return a value from the cache, ignoring the server completely
+  /// (implying that the returned value may be stale with respect to the value on the server). If
+  /// there is no data in the cache to satisfy the [get()] or [getDocuments()] call,
+  /// [DocumentReference.get()] will return an error and [Query.getDocuments()] will return an empty
+  /// [QuerySnapshot] with no documents.
+  cache,
+}
+
+// TODO: this needs to be more precise
+class PlatformQuerySnapshot {
```
hterkelsen: You should probably model this class on https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/QuerySnapshot

I would remove `asMap()` as well, and make the app-facing plugin use the typed data

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/interfaces/firestore_platform.dart
```diff
@@ -0,0 +1,84 @@
...
+import './write_batch_platform.dart';
+
+import '../implementations/method_channel_firestore.dart';
+
+/// The Firestore platform interface.
+abstract class FirestorePlatform extends PlatformInterface {
+  /// Constructor
+  FirestorePlatform() : super(token: _token);
+
+  static const Object _token = Object();
```
amirh: Make these final, this was an overlooked bug, see: https://github.com/flutter/plugins/pull/2417

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
ditman: You're right, though each instance of the platform has its own copy of the MethodChannel (I HOPE)

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
ditman: Thanks, I think I didn't realize this. I'll get this done!

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/interfaces/firestore_platform.dart
ditman: Ouch! Thanks for the heads up, I'll make all my `_token`s `final` instead of `const`

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/types/query.dart
ditman: Yeah, I've been avoiding this type for a while. I have a suspicion that the data that the backend sends back through the channel to the frontend is not *entirely* identical to the one documented, so I'll model this as I go (I might be wrong)

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/types/write_batch.dart
ditman: Yes, docs are missing pretty much everywhere. I'm giving myself some time before I migrate this to add *Handle at the end, but I agree it's clearer.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
ditman: Oh my, this codec complicates things 😭

ditman: Because of how the Codec operates (it basically uses the frontend of the plugin to decode some data types), I changed the initialization code of the MethodChannel Implementations to allow to pass a codec in; that way we can pass a codec with access to "userland" from the plugin, without having to migrate EVERY data type defined in the plugin (I tried, it was painful, and still didn't solve my issue).

This is [the problematic bit](https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_firestore/cloud_firestore/lib/src/firestore_message_codec.dart#L91-L96).

ditman:
Some more tests are passing now; after adding the Transaction implementation:

`00:01 +27 -19: Some tests failed.`

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
ditman: The code above probably explains why there's so much "static" code, and initialization guards in different places of the plugin; since we're creating new instances of the plugin all the time the message codec needs to hydrate a serialized DocumentReference!

Fixing this is *very* involved; we may be able to inject a factory function so the codec doesn't need to create instances in there, but maybe this particular part of the plugin could benefit from some extra design work (I didn't find any other place where layers are pierced like this, so far)

collinjackson: Is hydrating a serialized DocumentReference expensive? It's not clear to how much overhead is involved in that, compared to the general overhead of moving serialized snapshots in and out of the Dart VM.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
ditman: @hterkelsen you were right, even across different instances of a MethodChannel, only the LAST call to `setMethodCallHandler` seems to be called.

@mklim said: "There's only one handler per channel. A way to thin the Handler code is to actually move the implementation of the methods into other classes, and write the Handler itself just to plumb from method call and response to the business logic and back."

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/method_channel_document_reference.dart
ditman: Currently, it is *not* very expensive because:

* There's a bunch of init guards/static code that prevents multiple initialization
* We hydrate the moment we see a DcoumentReference that needs to be hydrated.

However, logic-wise, it **is** expensive, because the deserialization needs to be able to basically instantiate a whole new plugin. 

If we want to break that up, we'd probably need to hoist all the data types to the platform interface. At that moment, to maintain current feature parity, we'd need hydration on getters (or similar). The cost of that hydration would depend on the complexity of the returned data.

(Or maybe there's something I'm not seeing, there's a chance I've been focusing on this code for too long already :P)

ditman:
`00:02 +36 -10: Some tests failed.`

ditman:
Added DocumentReference implementation.

**`00:01 +46: All tests passed!`**

ditman:
(Currently fixing dartfmt and analyzer)

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/interfaces/query_platform.dart
```diff
@@ -0,0 +1,60 @@
...
+  static QueryPlatform _instance;
+
+  // TODO(amirh): Extract common platform interface logic.
+  // https://github.com/flutter/flutter/issues/43368
+  static set instance(QueryPlatform instance) {
+    PlatformInterface.verifyToken(instance, _token);
+    _instance = instance;
+  }
+
+  // Actual interface
```
collinjackson: I am leaning towards removing this non-Dartdoc comment (in all `PlatformInterface` subclasses) since people will be follow the example of this plugin and it clutters up the class a bit. If the code is self-documenting only Dartdoc comments will be necessary.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/implementations/multi_method_channel.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+
+typedef _MultiMethodChannelHandler(MethodCall call);
+
+class MultiMethodChannel extends MethodChannel {
```
collinjackson: Wondering if we should move this into a package (in the Flutter SDK? in the plugin_platform_interface package?) since it will be useful for firebase_database and firebase_storage. /cc @amirh

packages/cloud_firestore/cloud_firestore/pubspec.yaml
```diff
@@ -15,6 +15,8 @@ flutter:
         pluginClass: FLTCloudFirestorePlugin
 
 dependencies:
+  cloud_firestore_platform_interface:
+    path: ../cloud_firestore_platform_interface
```
collinjackson: Should eventually be pointing to a hosted version (probably by splitting this into two PRs)

packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
```diff
@@ -123,39 +84,41 @@ class Firestore {
...
-    _transactionHandlers[transactionId] = transactionHandler;
-    final Map<String, dynamic> result = await channel
-        .invokeMapMethod<String, dynamic>(
-            'Firestore#runTransaction', <String, dynamic>{
-      'app': app.name,
-      'transactionId': transactionId,
-      'transactionTimeout': timeout.inMilliseconds
-    });
+
+    // Wrap the user-supplied [TransactionHandler] into something that can be passed to the Platform implementation.
```
collinjackson:
```suggestion
    // Wrap the user-supplied [TransactionHandler] into something that can be passed to the platform implementation.
```

packages/cloud_firestore/cloud_firestore/test/cloud_firestore_test.dart
```diff
@@ -167,7 +169,14 @@ void main() {
           default:
             return null;
         }
-      });
+      };
+
+      // Ensure the MethodChannel platforms are initialized
+      Firestore.platform;
```
collinjackson: It feels unintuitive that this has side effects. What do you think about doing a method instead?

packages/cloud_firestore/cloud_firestore/test/cloud_firestore_test.dart
```diff
@@ -167,7 +169,14 @@ void main() {
...
+
+      // All the MethodChannel implementation share the same channel
```
collinjackson:
```suggestion
      // All the MethodChannel implementations share the same channel
```

packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
```diff
@@ -8,32 +8,11 @@ part of cloud_firestore;
...
-        return result;
-      }
-    });
-    _initialized = true;
-  }
+  Firestore({FirebaseApp app}) : app = app ?? FirebaseApp.instance;
+
+  /// The platform instance that talks to the native side of the plugin.
+  @visibleForTesting
+  static final FirestorePlatform platform = FirestorePlatform.instance ?? (FirestorePlatform.instance = MethodChannelFirestore(FirestoreMessageCodec()));
```
collinjackson: It seems a bit unintuitive to be doing the assignment like this.

ditman:
(Closing in favor of https://github.com/FirebaseExtended/flutterfire/pull/1670)

PR #1826, opened 2020-01-16
[cloud_functions] Prepare for web implementation

sbeitzel:
## Description

Based on the layout of the firebase_auth plugin, this change sets the stage for a web implementation of the cloud_functions plugin. It adds a new package, cloud_functions_platform_interface.

## Related Issues

This PR addresses issue flutter/flutter#45299

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
```diff
@@ -0,0 +1,82 @@
...
+/// `implements` this interface will be broken by newly added
+/// [CloudFunctionsPlatform] methods.
+abstract class CloudFunctionsPlatform {
+  /// Only mock implementations should set this to `true`.
+  ///
+  /// Mockito mocks implement this class with `implements` which is forbidden
+  /// (see class docs). This property provides a backdoor for mocks to skip the
+  /// verification that the class isn't implemented with `implements`.
+  @visibleForTesting
+  bool get isMock => false;
```
collinjackson: This is a deprecated way of doing federation. You can find the new approach here: https://github.com/FirebaseExtended/flutterfire/pull/1670/files#diff-0f5e1382550b037f55f569cbfe8b11adR45

See also: https://pub.dev/packages/plugin_platform_interface

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
```diff
@@ -0,0 +1,82 @@
...
+  /// This method ensures that [CloudFunctionsPlatform] isn't implemented with `implements`.
+  ///
+  /// See class docs for more details on why using `implements` to implement
+  /// [CloudFunctionsPlatform] is forbidden.
+  ///
+  /// This private method is called by the [instance] setter, which should fail
+  /// if the provided instance is a class implemented with `implements`.
+  void _verifyProvidesDefaultImplementations() {}
+
+  /// Gets an instance of a Callable HTTPS trigger in Cloud Functions.
```
hterkelsen: From looking at the `cloud_functions` package, I think the correct platform interface is actually:

```dart
dynamic callCloudFunction({
  @required String appName,
  @required String functionName, 
  String region,
  String origin,
  Duration timeout,
  dynamic parameters,
}) { ... }
```

Platform interfaces should strive to be very simple abstraction of the `MethodChannel` calls that already exist in the current plugin unless the plugin is very complex. The idea is that this is the minimal interface that a platform would need to implement to support this plugin. By pulling in the abstractions from the existing plugin (`HttpsCallable`, `HttpsCallableResult`) you are making `package:cloud_functions` less flexible.

If you just have `dynamic callCloudFunction(...)` then in `package:cloud_functions` you just replace the method call with `CloudFunctionPlatform.instance.callCloudFunction(...)` and the rest of the plugin stays the same. The plugin is also free to change its `HttpsCallable` abstraction however it wants, since the platform interface offers the basic functionality.

sbeitzel: Okay, that makes sense. However, wouldn't it also be good to include the `useFunctionsEmulator` call? After all, the Firebase CLI includes a local emulator for functions specifically for development; wouldn't it make sense for the platform interface to expose a control to allow code to specify whether to use the emulator or the real functions, no matter what platform implementation is being used? Or does this somehow belong on the `cloud_functions` plugin?

sbeitzel: I guess my concern is that this would be a breaking change. Maybe that's okay, given that the point here is to prepare for flutter_web. Do you reckon that's acceptable?

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
```diff
@@ -0,0 +1,82 @@
...
+
+/// The interface that implementations of `cloud_functions` must extend.
+///
+/// Platform implementations should extend this class rather than implement it
+/// as `cloud_functions` does not consider newly added methods to be breaking
+/// changes. Extending this class (using `extends`) ensures that the subclass
+/// will get the default implementation, while platform implementations that
+/// `implements` this interface will be broken by newly added
+/// [CloudFunctionsPlatform] methods.
+abstract class CloudFunctionsPlatform {
```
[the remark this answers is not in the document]
sbeitzel: Absolutely, will do. Thank you for pointing me in the right direction!

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
hterkelsen: It wouldn't be a breaking change if you leave `HttpsCallable`, `HttpsCallableResult`, etc in `package:cloud_functions`. As for `useFunctionsEmulator`, you can document that setting an `origin` will change it to use a functions emulator. See the Android code here: https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_functions/android/src/main/java/io/flutter/plugins/firebase/cloudfunctions/CloudFunctionsPlugin.java#L49

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
```diff
@@ -0,0 +1,72 @@
...
+/// Platform implementations should extend this class rather than implement it
+/// as `cloud_functions` does not consider newly added methods to be breaking
+/// changes. Extending this class (using `extends`) ensures that the subclass
+/// will get the default implementation, while platform implementations that
+/// `implements` this interface will be broken by newly added
+/// [CloudFunctionsPlatform] methods.
+abstract class CloudFunctionsPlatform extends PlatformInterface {
+  static final Object _token = Object();
+
+  /// Constructs a CloudFunctionsPlatform
```
hterkelsen: nit: end doc comments with a period

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
```diff
@@ -0,0 +1,72 @@
...
+  ///
+  /// The data passed into the cloud function can be any of the following types:
+  ///
+  /// `null`
+  /// `String`
+  /// `num`
+  /// [List], where the contained objects are also one of these types.
+  /// [Map], where the values are also one of these types.
+  ///
+  /// @param appName name of the Firebase application to which the function belongs
```
hterkelsen: we don't use the JavaDoc style with `@param` sections for each parameter. The common style in Dartdocs is to write a paragraph (or paragraphs) describing the parameters and wrapping the parameter name in square brackets so they are linked. For example:

```
/// The [appName] is the name of the Firebase app to which the function belongs. ...
```

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
```diff
@@ -0,0 +1,63 @@
...
+    Duration timeout,
+    dynamic parameters,
+  }) async {
+    try {
+      final dynamic response = await channel
+          .invokeMethod<dynamic>('CloudFunctions#call', <String, dynamic>{
+        'app': appName,
+        'region': region,
+        'origin': origin,
+        'timeoutMicroseconds': timeout?.inMicroseconds,
```
hterkelsen: I think this should be `'timeoutMilliseconds': timeout?.inMilliseconds,` based on https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_functions/android/src/main/java/io/flutter/plugins/firebase/cloudfunctions/CloudFunctionsPlugin.java#L52

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
```diff
@@ -0,0 +1,98 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+part of cloud_functions_platform_interface;
+
+class CloudFunctionsException implements Exception {
```
hterkelsen: let's leave `CloudFunctionsException` in `package:cloud_functions` and not catch the `PlatformException` below. This allows `package:cloud_functions` to be more flexible about handling exceptions (and doesn't limit `CloudFunctionsException` to just the method channel implementation

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
```diff
@@ -0,0 +1,98 @@
...
+  CloudFunctionsException._(this.code, this.message, this.details);
+
+  final String code;
+  final String message;
+  final dynamic details;
+}
+
+class MethodChannelCloudFunctions extends CloudFunctionsPlatform {
```
hterkelsen: add doc comment saying this is the platform implementation backed by a `MethodChannel`

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
sbeitzel: Does that mean that the existing implementation is a bug?
https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_functions/lib/src/https_callable.dart#L38

I mean, I'm happy to make the change if it's the right thing to do, but I don't want to break something that works, currently.

sbeitzel: It seems that the Android code is the odd one out. iOS is also using microseconds:

`NSNumber *timeoutMicroseconds = call.arguments[@"timeoutMicroseconds"];`

https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_functions/ios/Classes/CloudFunctionsPlugin.m#L48

hterkelsen: Yes, it looks like a bug in the Android implementation

packages/cloud_functions/cloud_functions_platform_interface/LICENSE
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017-2020 The Chromium Authors. All rights reserved.
```
hterkelsen: I am not a lawyer but I think it would be best if this LICENSE file copied directly from the one in `packages/cloud_functions/cloud_functions/LICENSE` and had the Copyright year set to 2018

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
```diff
@@ -0,0 +1,75 @@
...
+  /// This can be used to send requests to a local emulator.
+  ///
+  /// The data passed into the cloud function via [parameters] can be any of the following types:
+  ///
+  /// `null`
+  /// `String`
+  /// `num`
+  /// [List], where the contained objects are also one of these types.
+  /// [Map], where the values are also one of these types.
+  dynamic callCloudFunction({
```
hterkelsen: this should return `Future<dynamic>`

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
```diff
@@ -0,0 +1,54 @@
...
+  ///
+  /// The data passed into the cloud function via [parameters] can be any of the following types:
+  ///
+  /// `null`
+  /// `String`
+  /// `num`
+  /// [List], where the contained objects are also one of these types.
+  /// [Map], where the values are also one of these types.
+  @override
+  dynamic callCloudFunction({
```
hterkelsen: This should return `Future<dynamic>`

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
```diff
@@ -0,0 +1,54 @@
...
+    @required String appName,
+    @required String functionName,
+    String region,
+    String origin,
+    Duration timeout,
+    dynamic parameters,
+  }) async {
+    final dynamic response = await channel
```
hterkelsen: nit: you can change this entire method body to:

```dart
@override
Future<dynamic> callCloudFunction(
  ...
) => channel.invokeMethod<dynamic>('CloudFunctions#call', <String, dynamic>{ ... });
```

Since `channel.invokeMethod<dynamic>` already returns a `Future<dynamic>`

packages/cloud_functions/cloud_functions_platform_interface/LICENSE
sbeitzel: Because of the way copyright works, one wants to claim copyright as soon as a work is published and then restate the claim any time new work is performed. Otherwise, the copyright will lapse early.

https://epiphany.law/articles/copyright/copyright-dates-single-year-or-range/

packages/cloud_functions/cloud_functions_platform_interface/lib/cloud_functions_platform_interface.dart
sbeitzel: Will do

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
sbeitzel: Will do.

packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
sbeitzel: Absolutely.

hterkelsen:
Thanks, @sbeitzel ! Just merged this and uploaded `cloud_functions_platform_interface` to pub

Issue #95, opened 2019-09-06
[cloud_functions] Timeout doesn't work on android


noelmansour:
The android code references `timeoutMilliseconds`: 

https://github.com/FirebaseExtended/flutterfire/blob/bfc666daa89698c9fdc342e7ae521ec5e4e8ec5b/packages/cloud_functions/android/src/main/java/io/flutter/plugins/firebase/cloudfunctions/CloudFunctionsPlugin.java#L52

but the dart code passes `timeoutMicroseconds`:

https://github.com/FirebaseExtended/flutterfire/blob/bfc666daa89698c9fdc342e7ae521ec5e4e8ec5b/packages/cloud_functions/lib/src/https_callable.dart#L38

The android plugin should use the parameter `timeoutMicroseconds` and should also use `TimeUnit.MICROSECONDS` instead of `TimeUnit.MILLISECONDS`.

noelmansour:
Actually, it looks like the `setTimeout()` functionality was broken in the firebase-functions android lib 17.0.0 (the version cloud_functions uses). See firebase/firebase-android-sdk#604

In addition to fixing the parameter issue mentioned above, the android library should be upgraded to 19.0.1.

I'd be happy to open a PR for this, but not sure if library upgrades require a special review.

JonasJuss:
I'm experiencing timeouts in Flutter after 5-10 seonds, with the DEADLINE_EXCEEDED error, even though the Firebase Function completes successfully and reports no error.

I came across this Stack Overflow post, reporting the problem in the Android functions library:
https://stackoverflow.com/questions/56977242/firebase-functions-gives-timeout-on-client-side-after-10s

It seems that an upgrade to the newest Android functions library would solve this problem.

JonasJuss:
@noelmansour do you know of any workaround for this, while we wait for the update?

noelmansour:
You'd have to fork this repo and make the changes I described in the comments. Then use that as the cloud_functions plugin for your project.

Kretin1:
> 
> 
> I'm experiencing timeouts in Flutter after 5-10 seonds, with the DEADLINE_EXCEEDED error, even though the Firebase Function completes successfully and reports no error.
> 
> I came across this Stack Overflow post, reporting the problem in the Android functions library:
> https://stackoverflow.com/questions/56977242/firebase-functions-gives-timeout-on-client-side-after-10s
> 
> It seems that an upgrade to the newest Android functions library would solve this problem.

[...]

PR #1854, opened 2020-01-22
[cloud_functions] Update cloud_functions to use platform interface

sbeitzel:
## Description

This change updates the existing `cloud_functions` plugin to use `cloud_functions_platform_interface`.

## Related Issues

This addresses issue flutter/flutter#45299
It's also a drive-by fix for issue #95 

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/cloud_functions/cloud_functions/android/src/main/java/io/flutter/plugins/firebase/cloudfunctions/CloudFunctionsPlugin.java
```diff
@@ -49,9 +49,9 @@ public void onMethodCall(MethodCall call, final Result result) {
           functions.useFunctionsEmulator(origin);
         }
         HttpsCallableReference httpsCallableReference = functions.getHttpsCallable(functionName);
-        Number timeoutMilliseconds = call.argument("timeoutMilliseconds");
```
sbeitzel: This change comes as a result of discussion in the [first PR](https://github.com/FirebaseExtended/flutterfire/pull/1826) in this series, where we discovered that the iOS plugin and the Dart code were using the named parameter `timeoutMicroseconds`. As a result, the Android plugin would always use the default timeout even if a timeout was specified, since looking for parameter `timeoutMilliseconds` would always return `null`. I haven't yet searched to see if there's already a bug filed for this behavior. If there is, then this change will also address that bug.

packages/cloud_functions/cloud_functions/lib/src/https_callable.dart
```diff
@@ -29,17 +29,16 @@ class HttpsCallable {
...
-          .invokeMethod<dynamic>('CloudFunctions#call', <String, dynamic>{
-        'app': _cloudFunctions._app.name,
-        'region': _cloudFunctions._region,
-        'origin': _cloudFunctions._origin,
-        'timeoutMicroseconds': timeout?.inMicroseconds,
-        'functionName': _functionName,
-        'parameters': parameters,
-      });
-      return HttpsCallableResult._(response);
+      return CloudFunctionsPlatform.instance
```
hterkelsen: You can either modify this slightly to use `async`/`await` as it is originally or you can get rid of the `async` in the function header.

sbeitzel: Oh, gosh. Good point.

packages/cloud_functions/cloud_functions/android/src/main/java/io/flutter/plugins/firebase/cloudfunctions/CloudFunctionsPlugin.java
sbeitzel: There is already an issue filed for this: #95

packages/cloud_functions/cloud_functions/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.4.2
```
collinjackson: nit: since this doesn't introduce any new Dart API surface area I think it should be 0.4.1+7.

sbeitzel: Okay, that's an easy change to make. I'll upload it shortly.

KevinTheGray:
@sbeitzel could you give me a quick rundown on/if there's anything to do for the following issue I had when upgrading to this version (this is on an iOS simulator): 
Screen Shot 2020-01-23 at 10 58 09 AM


I haven't dug too deeply, and just downgraded for now as I have to get a quick build out.  

this is the call I was making 
```
HttpsCallableResult callableResult = await CloudFunctions.instance
          .getHttpsCallable(
            functionName: functionName,
          )
          .call(parameters);
```
I can dig a little deeper later today if this is an issue on my end.

sbeitzel:
I think it was an oversight on my end. Looking at the MethodChannel being registered in the new version versus the old version, I think I copy/pasted from the wrong place. The error, I suspect, lies in `method_channel_cloud_functions.dart` -- in the platform interface library rather than the cloud_functions library. I have to run to a meeting right now but I will prepare a patch later today.

KevinTheGray:
Sounds good, thanks!

sbeitzel:
[Here's the PR](https://github.com/FirebaseExtended/flutterfire/pull/1862) to fix the problem.

packages/cloud_functions/cloud_functions/lib/src/https_callable.dart
```diff
@@ -27,19 +27,18 @@ class HttpsCallable {
...
+          .callCloudFunction(
+            appName: _cloudFunctions._app.name,
+            region: _cloudFunctions._region,
+            origin: _cloudFunctions._origin,
+            timeout: timeout,
+            functionName: _functionName,
+            parameters: parameters,
+          )
+          .then((response) => HttpsCallableResult._(response));
     } on PlatformException catch (e) {
```
cacianokroth, line 42: Why does this catch remain, if an error does not fall into it?

PR #1862, opened 2020-01-23
[cloud_functions] Fix method channel name

sbeitzel:
## Description

In implementing the `MethodChannelCloudFunctions` I assigned the wrong name to the channel. This meant that the tests passed (because the channel was being invoked) but that when the cloud_functions plugin with iOS or Android implementations was used, the invocation would fail because the channel wouldn't be found.

This change fixes the method channel name so that it matches the name being registered by the iOS and Android plugins.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
+++ b/packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
@@ -9,7 +9,7 @@ class MethodChannelCloudFunctions extends CloudFunctionsPlatform {
   /// The [MethodChannel] to which calls will be delegated.
   @visibleForTesting
   static const MethodChannel channel = MethodChannel(
-    'plugins.flutter.io/cloud_functions',
+    'cloud_functions',
   );
 
   /// Invokes the specified cloud function.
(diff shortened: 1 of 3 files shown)


packages/cloud_functions/cloud_functions_platform_interface/lib/src/method_channel_cloud_functions.dart
```diff
@@ -7,9 +7,11 @@ part of cloud_functions_platform_interface;
 /// [CloudFunctionsPlatform] implementation that delegates to a [MethodChannel].
 class MethodChannelCloudFunctions extends CloudFunctionsPlatform {
   /// The [MethodChannel] to which calls will be delegated.
+  /// Note that this name must match the name used in the platform (iOS/Android)
+  /// plugin.
   @visibleForTesting
   static const MethodChannel channel = MethodChannel(
-    'plugins.flutter.io/cloud_functions',
```
sbeitzel: When I created this file, I was copying and modifying a different plugin and didn't understand exactly how `MethodChannel` names worked. The plugin I was cribbing from used a different naming convention, and I didn't catch the mistake. The unit tests only check to see that the `MethodChannel` gets invoked, not what the channel name is. Moreover, the actual iOS and Android implementations are in the `cloud_functions` plugin, not this platform interface plugin, meaning integration testing is problematic. This feels to me like an architecture problem -- like maybe the method channel implementation should be in the plugin with the platform implementations -- but there are ramifications that get complicated, and the "right thing" might wind up being a breaking change.

collinjackson:
I believe that https://github.com/FirebaseExtended/flutterfire/issues/1866 was been fixed by changing the channel name to match the existing channel name on iOS/Android in https://github.com/FirebaseExtended/flutterfire/pull/1868

So I think this PR can be closed. If I missed something let me know!

Issue #1682, opened 2019-12-19
[firebase_auth_web] time to implement fetchSignInMethodsForEmail


azol:
`firebase_auth_web.dart` contains unimplemented `fetchSignInMethodsForEmail` waiting for resolving https://github.com/FirebaseExtended/firebase-dart/issues/272.

https://github.com/FirebaseExtended/firebase-dart/issues/272 is resoled so it's time to implement :)

Ehesp:
Support has landed for this in the latest release.

PR #1698, opened 2019-12-22
[firebase_auth_web] WIP mapping package:firebase FirebaseError to PlatformException

GregorySech:
## Description

I'm mapping the known firebasejs errors (received in package:firebase) to Platform exception as described in https://github.com/FirebaseExtended/flutterfire/issues/1559.

It's still a work in progress, I'll try to add some tests and update documentation when I've got some free time.

## Related Issues

* Closes https://github.com/FirebaseExtended/flutterfire/issues/1559
## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] If the pull request affects only one plugin, the PR title starts with the name of the plugin in brackets (e.g. [cloud_firestore])
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [ ] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [ ] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/FirebaseExtended/flutterfire/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #1864, opened 2020-01-24
[firebase_auth] implement missing auth functionality for web

CDDelta:
## Description
Implements the missing auth behavior for web except for  `verifyPhoneNumber`.

## Related Issues
#1682 

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_auth/firebase_auth_web/CHANGELOG.md
+++ b/packages/firebase_auth/firebase_auth_web/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.1.2
 
-* Implement `fetchSignInMethodsForEmail`, `isSignInWithEmailLink`, and `sendLinkToEmail`. 
+* Implement `fetchSignInMethodsForEmail`, `isSignInWithEmailLink`, `signInWithEmailAndLink`, and `sendLinkToEmail`. 
 
 ## 0.1.1+4
 
--- a/packages/firebase_auth/firebase_auth_web/CHANGELOG.md
+++ b/packages/firebase_auth/firebase_auth_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.2
+
+* Implement `fetchSignInMethodsForEmail`, `isSignInWithEmailLink`, and `sendLinkToEmail`. 
+
 ## 0.1.1+4
 
 * Prevent `null` users (unauthenticated) from breaking the `onAuthStateChanged` Stream.
--- a/packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
+++ b/packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
@@ -168,9 +168,9 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
   }
 
   @override
-  Future<bool> isSignInWithEmailLink(String app, String link) async {
+  Future<bool> isSignInWithEmailLink(String app, String link) {
     final firebase.Auth auth = _getAuth(app);
-    return auth.isSignInWithEmailLink(link);
+    return Future.value(auth.isSignInWithEmailLink(link));
   }
 
   @override
--- a/packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
+++ b/packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
@@ -147,10 +147,8 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
 
   @override
   Future<List<String>> fetchSignInMethodsForEmail(String app, String email) {
-    // TODO(hterkelsen): Use `fetchSignInMethodsForEmail` once
-    // https://github.com/FirebaseExtended/firebase-dart/issues/272
-    // is resolved.
-    throw UnimplementedError('fetchSignInMethodsForEmail');
+    final firebase.Auth auth = _getAuth(app);
+    return auth.fetchSignInMethodsForEmail(email);
   }
 
   @override
@@ -162,8 +160,6 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
 
   @override
   Future<PlatformIdTokenResult> getIdToken(String app, bool refresh) async {
-    // TODO(hterkelsen): `package:firebase` added `getIdTokenResult` in
-    // version 7.0.0. Use it here once that is published.
     final firebase.Auth auth = _getAuth(app);
     final firebase.User currentUser = auth.currentUser;
     final firebase.IdTokenResult idTokenResult =
@@ -172,11 +168,9 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
   }
 
   @override
-  Future<bool> isSignInWithEmailLink(String app, String link) {
-    // TODO(hterkelsen): Implement this once
-    // https://github.com/FirebaseExtended/firebase-dart/issues/273
-    // is resolved.
-    throw UnimplementedError('isSignInWithEmailLink');
+  Future<bool> isSignInWithEmailLink(String app, String link) async {
+    final firebase.Auth auth = _getAuth(app);
+    return auth.isSignInWithEmailLink(link);
   }
 
   @override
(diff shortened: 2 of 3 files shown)


packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -172,11 +168,9 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
   }
 
   @override
-  Future<bool> isSignInWithEmailLink(String app, String link) {
-    // TODO(hterkelsen): Implement this once
-    // https://github.com/FirebaseExtended/firebase-dart/issues/273
-    // is resolved.
-    throw UnimplementedError('isSignInWithEmailLink');
+  Future<bool> isSignInWithEmailLink(String app, String link) async {
```
ditman: You're not `await`ing for anything in this method, so it doesn't need to be marked as `async`, right?

_(If this **does** need to be `async`, the [platform interface](https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_auth/firebase_auth_platform_interface/lib/firebase_auth_platform_interface.dart#L110-L112) may need to be updated as well!)_

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -232,9 +226,20 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
...
+      url: url,
+      handleCodeInApp: handleCodeInApp,
+      iOS: firebase.IosSettings(
+        bundleId: iOSBundleID,
+      ),
+      android: firebase.AndroidSettings(
+        packageName: androidPackageName,
+        installApp: androidInstallIfNotAvailable,
+        minimumVersion: androidMinimumVersion,
+      ),
```
ditman: I was reading the [documentation](https://firebase.google.com/docs/auth/web/passing-state-in-email-actions#passing_statecontinue_url_in_email_actions) and... do we want to kick the user to the Android/iOS app from the web? This change kind of adds a requirement of setting up an iOSBundleID and an androidPackageName on a potentially web-only app.

Can these attributes be passed to the ActionCodeSettings conditionally? Does this handle `null` values gracefully, or does it cause an error?

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -280,9 +285,10 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
   @override
   Future<PlatformAuthResult> signInWithEmailAndLink(
       String app, String email, String link) async {
-    // TODO(hterkelsen): Use signInWithEmailLink once 7.0.0 of package:firebase
-    // is released.
-    throw UnimplementedError('signInWithEmailAndLink');
+    final firebase.Auth auth = _getAuth(app);
+    final firebase.UserCredential userCredential =
+        await auth.signInWithEmailLink(email, link);
```
ditman: This might _"Fail with an error if the email address is invalid or OTP in email link expires."_. [Docs](https://firebase.google.com/docs/reference/js/firebase.auth.Auth.html#signinwithemaillink).

Does that case need to be handled, or would that exception (?) propagate to the client app? _(I've seen other places in the web plugin with a similar pattern... How do the Android/iOS versions behave in that case?)_

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -162,8 +160,6 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
 
   @override
   Future<PlatformIdTokenResult> getIdToken(String app, bool refresh) async {
-    // TODO(hterkelsen): `package:firebase` added `getIdTokenResult` in
-    // version 7.0.0. Use it here once that is published.
```
ditman: Thanks!!

packages/firebase_auth/firebase_auth_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.1.2
+
+* Implement `fetchSignInMethodsForEmail`, `isSignInWithEmailLink`, and `sendLinkToEmail`.
```
ditman: Can you add some unit tests? At least, it'd be interesting to have a test of the sad path of `signInWithEmailLink`, that codifies whatever is decided in this PR (whether it throws or catch + return null... whatever)

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
CDDelta: It needs to be `async` as `auth.isSignInWithEmailLink(link)` returns a `bool` on web not a `Future<bool>` as the interface expects. I can change it to use `Future.value()` if that's preferred instead.

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
CDDelta: You're right, we shouldn't force users to set up those attributes.

I can change to make this optional but it would require modifying the plugin [to not assert that these attributes are non-null](https://github.com/FirebaseExtended/flutterfire/blob/36b3880179a0334e4d6b347ca6cd2dd0c2c1c600/packages/firebase_auth/firebase_auth/lib/src/firebase_auth.dart#L117) and [update the native code to gracefully handle non-null values](https://github.com/FirebaseExtended/flutterfire/blob/36b3880179a0334e4d6b347ca6cd2dd0c2c1c600/packages/firebase_auth/firebase_auth/android/src/main/java/io/flutter/plugins/firebaseauth/FirebaseAuthPlugin.java#L344) too to keep the api behavior consistent on all platforms.

Is that the route we want to go down or is there an alternative?

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
CDDelta: Right now it looks like the function will throw a `FirebaseError` instance to the client which we will want to map to a `PlatformException` but that would require #1698 to land first.

packages/firebase_auth/firebase_auth_web/CHANGELOG.md
CDDelta: Do we need to do that? I believe the `firebase-dart` package will properly throw the expected `FirebaseError` instances which #1698 will map to our expected `PlatformException`s.

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
ditman: Yes, please, I think you should return a Future.value, instead of changing the interface here.

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
ditman: I'm deferring this one to @hterkelsen, he wrote the original implementation. IMO all these should be optional. I'm guessing in a web-only flow, those are never expected to be used, right?

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
ditman: Ah, yes, that 1698 is going to be handy. Thanks for pointing it out!

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
hterkelsen: Since these are already required parameters in `package:firebase_auth` I think it makes sense to use them like this. All this is saying is if the Android or iOS app is installed then the sign in link will attempt to open the app,

CDDelta: Maybe this is out of the scope of this PR but there is currently no way for users to not include an Android `packageName` or iOS `bundleId`.

Passing a null will cause the method to assert and passing an empty string will cause Firebase to throw an argument error. A possible hack might be to pass in a random string but that seems fragile and possibly dangerous.

I think it would be a good idea to allow users to opt out of certain platforms by passing null.

CDDelta:
@hterkelsen what are the things I need to do to get this change merged? :)

hterkelsen:
Never mind. The failure is unrelated and is being fixed in a separate PR.

PR #1884, opened 2020-01-27
Fix fallthrough bug

mehmetf:
This was caught during google3 roll. When `confirmPasswordReset` is called, it will throw a `notImplemented` error since there's no break in the case statement.


mehmetf:
I need to submit this. It is a very simple fix and is blocking google3 rolls.

PR #1936, opened 2020-02-05
[cloud_firestore] Update README.md with solution for multidex issue

collinjackson:
Fixes https://github.com/FirebaseExtended/flutterfire/issues/1914

### DIFF
--- a/packages/cloud_firestore/cloud_firestore/README.md
+++ b/packages/cloud_firestore/cloud_firestore/README.md
@@ -23,7 +23,7 @@ with Xcode, and within Xcode place the file inside ios/Runner. Don't follow the 
 
 ## Building for Android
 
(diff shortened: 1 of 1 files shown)


ashwinraghav:
LGTM

collinjackson:
> Hi
> 
> This solution does not completely satisfy the issue.
> Changing `minSdkVersion` is kind of a business choice and developer might not want to increase it to 21 (Since the users with Android 4.4 or lower will lose the use of app)
> 
> So better to add full explanation on how to fix multiDex issue for all sdk versions.
> 
> Simon has offered some [suggestion](https://github.com/flutter/flutter/issues/21009#issue-353951634) to the flutter which can be used until the issue becomes resolved in the flutter framework itself.

Thanks for the advice! I'll add a link to that issue.

xster:
ya, this isn't right, even if your minsdk is 21+, you still need to do https://developer.android.com/studio/build/multidex#mdex-gradle. The dependency is just the runtime, you still need to use the runtime.

Issue #195, opened 2019-09-21
Bad state: Future already completed


earyzhe:
When initilising remote config  this error is thrown.

### The code
```
remoteConfig = await RemoteConfig.instance;
			remoteConfig.setConfigSettings(RemoteConfigSettings(debugMode: true));
			await remoteConfig.fetch(expiration: const Duration(seconds: 1));
			await remoteConfig.activateFetched();
			String stingified = remoteConfig.getString("app_questions");
```

###  The Error

Bad state: Future already completed

###  Output Flutter -v: 

```
✓] Flutter (Channel stable, v1.9.1+hotfix.2, on Mac OS X 10.14.6 18G95, locale en-GB)
    • Flutter version 1.9.1+hotfix.2 at /Users/earyzhe/dev/SDKs/flutter
    • Framework revision 2d2a1ffec9 (2 weeks ago), 2019-09-06 18:39:49 -0700
    • Engine revision b863200c37
    • Dart version 2.5.0

 
[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /Users/earyzhe/Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-28, build-tools 28.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_202-release-1483-b49-5587405)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 11.0)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 11.0, Build version 11A420a
    • CocoaPods version 1.7.5

[✓] Android Studio (version 3.5)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 39.0.3
    • Dart plugin version 191.8423
    • Java version OpenJDK Runtime Environment (build 1.8.0_202-release-1483-b49-5587405)

[✓] IntelliJ IDEA Community Edition (version 2018.3.4)
    • IntelliJ at /Applications/IntelliJ IDEA CE.app
    • Flutter plugin version 33.2.2
    • Dart plugin version 183.5901

[✓] VS Code (version 1.38.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.4.1

[✓] Connected device (3 available)
[...]

PR #28, opened 2019-08-25
[firebase_remote_config] Fix Future already completed RemoteConfig.instance

collinjackson:
Migrates flutter/plugins#1966 by @IchordeDionysos

I ended up rewriting it a bit.

Note: It seems the plugin was working fine and the behavior is covered by a test, but there was a spurious error message in the logs. This PR is removing the error message in the logs, and I don't think I can easily cover that in a test.

Fixes flutter/flutter#38060

PR #2061, opened 2020-02-26
[firebase_remote_config] Remote config/double instance error

IchordeDionysos:
## Description

This pull request addresses an issue with a test where two instances were created shortly after another leading to a `Future already completed` `StateError`.
This was caused due to side-effects of the tests where all tests passed when they were executed at once and the `doubleInstance` test failed when executed alone.

Also in this Pull Request is a fix for the `doubleInstance` error. Realized with a try/catch for only this exact error. I've used this fix because to me it feels much more elegant compared to a fix where I would basically would have to check for the same shortly after.

Alternative solution:

```dart
if (!instanceCompleter.isCompleted) {
  final _remoteConfigInstance = await _getRemoteConfigInstance();
  if (!instanceCompleter.isCompleted) {
    instanceCompleter.complete(_remoteConfigInstance);
  }
}
```

## Related Issues

#195
flutter/flutter#38060
#28

## Checklist

- [x] I signed the [CLA].

## Breaking Change


IchordeDionysos:
Is the failing check known to behave flakily? 🤔 
I can't figure out what code changes the check failure would have triggered.
Also, it says `fatal: repository 'https://cdn.cocoapods.org/' not found`...

IchordeDionysos:
Also, I'm open to alternative solutions to fix the error "Future already completed", 

I just wanted to fix the tests 
and get the bug fix going, 
because when I attach a debugger it always stops first at the exception/error and not my breakpoint, which is annoying.

Interestingly, this does work and gives back a `RemoteConfig` instance it's just that the error still occurs.

packages/firebase_remote_config/lib/src/remote_config.dart
```diff
@@ -25,17 +25,24 @@ class RemoteConfig extends ChangeNotifier {
...
 
   /// Gets the instance of RemoteConfig for the default Firebase app.
   static Future<RemoteConfig> get instance async {
-    if (!_instanceCompleter.isCompleted) {
-      _instanceCompleter.complete(await _getRemoteConfigInstance());
+    if (!instanceCompleter.isCompleted) {
+      try {
+        instanceCompleter.complete(await _getRemoteConfigInstance());
+      } on StateError catch (error) {
+        if (error.message != 'Future already completed') rethrow;
```
collinjackson: It seems like we could just check if the `instanceCompleter` is complete at this point, instead of checking the exception in a fragile way.

packages/firebase_remote_config/lib/src/remote_config.dart
```diff
@@ -25,17 +25,24 @@ class RemoteConfig extends ChangeNotifier {
   RemoteConfigSettings _remoteConfigSettings;
 
   DateTime get lastFetchTime => _lastFetchTime;
+
   LastFetchStatus get lastFetchStatus => _lastFetchStatus;
+
   RemoteConfigSettings get remoteConfigSettings => _remoteConfigSettings;
 
-  static Completer<RemoteConfig> _instanceCompleter = Completer<RemoteConfig>();
+  @visibleForTesting
```
collinjackson: We're generally trying to avoid @visibleForTesting APIs since they clutter the public logs, though I don't see another solution at this point.

packages/firebase_remote_config/lib/src/remote_config.dart
IchordeDionysos: I have rewritten the code to use an if statement.

packages/firebase_remote_config/lib/src/remote_config.dart
IchordeDionysos: Totally agree, I tried to find an alternative first before using it, but have to change it back somehow...

IchordeDionysos:
@collinjackson @kroikie Any updates on this?
Would it be okay to use @visibleForTesting in this case?

packages/firebase_remote_config/lib/src/remote_config.dart
IchordeDionysos: @FirebaseExtended/invertase @collinjackson is there any other solution to be able to test for this edge case?

IchordeDionysos:
@FirebaseExtended/invertase @Ehesp

It would be cool to have this fix landed. I've heard a couple of times in our company alone, that people spent quite some time getting to fix this issue, just because it sends a false-positive.

This adds a missing test case, where Remote Config is initialized twice in parallel, which leads to an error.

The current status of this PR is that @collinjackson wasn't too happy about using `@visibleForTesting`, but we found no other way to do so.
To fix the now failing test, I made some adjustments that there's no `Future already completed` error anymore.

I'll be happy to update the CHANGELOG.md if you are happy with the current implementation of the tests.

IchordeDionysos:
Closing in favor of #4186 which makes this PR obsolete!

PR #2099, opened 2020-03-02
[all] fix for UserAgent.h compilation failures

collinjackson:
Right now the development cocoapod is named the same thing on iOS and MacOS. I think this may be why the podspec's `prepare_command` is not running consistently for both iOS and MacOS.

The attached change points both platforms at a pregenerated shared UserAgent.h, ensuring that it always exists. With this change, the version should always be set but we'll be able to detect if it's not by an "unknown" version showing up as the version on the server side.

Fixes https://github.com/FirebaseExtended/flutterfire/issues/2087
Fixes https://github.com/FirebaseExtended/flutterfire/issues/1858
Fixes https://github.com/FirebaseExtended/flutterfire/issues/2084


collinjackson:
@franciscojma86 @jmagman while I think this fix will work for now, I'm wondering if Flutter plugins should be using a different pod name for the iOS and Mac development pods.

packages/cloud_firestore/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.13.4+1
+
+* Fix for missing UserAgent.h compliation failures.
```
danilofuchs:
```suggestion
* Fix for missing UserAgent.h compilation failures.
```

collinjackson: good catch, thanks

jmagman:
https://github.com/FirebaseExtended/flutterfire/issues/1858#issuecomment-577448620

> Those missing files should be coming from the [firebase* podspec `prepare_command`](https://github.com/FirebaseExtended/flutterfire/blob/7153c28f7517518157008e5410057205c6420860/packages/firebase_dynamic_links/ios/firebase_dynamic_links.podspec#L28) which only get run when the pod is first downloaded. I don't know how this works when the pod isn't being downloaded via `pod` but instead via a symlink to the pub-cache...
> I was able to reproduce this by adding this to a pubspec:
> 
> ```
>   firebase_auth: 
>   firebase_storage: 
>   firebase_core: 
> ```
> 
> Running `flutter pub get` then running:
> 
> ```
> flutter pub cache repair
> ```
> 
> which deleted the UserAgent.h files.
> 
> Essentially: `s.prepare_command` shouldn't be relied on in a podspec that is distributed via pub.

https://github.com/FirebaseExtended/flutterfire/blob/00c0c8707fcced86faec0d752b6c8bf43b3fcd7c/packages/cloud_firestore/cloud_firestore/ios/cloud_firestore.podspec#L28

Instead of generating a header file to define these macros instead you can use `GCC_PREPROCESSOR_DEFINITIONS` to define them (I didn't test this, should work in theory)  

```ruby
s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => ' $(inherited) LIBRARY_VERSION= "unknown" LIBRARY_NAME=flutter-fire-fst }
```
See https://github.com/FirebaseExtended/flutterfire/issues/1858#issuecomment-579456890.

collinjackson:
Thanks! I'll update the PR.

packages/cloud_firestore/cloud_firestore/ios/cloud_firestore.podspec
```diff
@@ -24,9 +24,5 @@ Firestore plugin for Flutter.
   s.dependency 'Firebase/Firestore', '~> 6.0'
   s.static_framework = true
 
-  s.prepare_command = <<-CMD
-      echo // Generated file, do not edit > Classes/UserAgent.h
-      echo "#define LIBRARY_VERSION @\\"#{libraryVersion}\\"" >> Classes/UserAgent.h
-      echo "#define LIBRARY_NAME @\\"flutter-fire-fst\\"" >> Classes/UserAgent.h
-    CMD
+  s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) LIBRARY_VERSION=\@\"unknown\" LIBRARY_NAME=\@\"flutter-fire-fst\"' }
```
jmagman: What's with the `\@`?

franciscojma86: a typo :) I was escaping only the "

packages/cloud_functions/cloud_functions/ios/cloud_functions.podspec
```diff
@@ -25,10 +25,6 @@ A new flutter plugin project.
   s.dependency 'Firebase/Functions', '~> 6.0'
   s.static_framework = true
 
-  s.prepare_command = <<-CMD
-      echo // Generated file, do not edit > Classes/UserAgent.h
-      echo "#define LIBRARY_VERSION @\\"#{libraryVersion}\\"" >> Classes/UserAgent.h
-      echo "#define LIBRARY_NAME @\\"flutter-fire-fn\\"" >> Classes/UserAgent.h
-    CMD
+  s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) LIBRARY_VERSION=@\"unknown\" LIBRARY_NAME=@\"flutter-fire-fn\"' }
```
collinjackson: Is it possible to use #{libraryVersion} here instead of "unknown"?

franciscojma86: Done. It wasn't converting the `#{libraryVersion} ` to the actual value so I concatenated the string.

collinjackson:
@franciscojma86 LGTM with your version (I can't review since I initially authored the PR).

Thanks @jmagman for the suggestion!

packages/cloud_firestore/cloud_firestore/ios/cloud_firestore.podspec
```diff
@@ -24,9 +24,5 @@ Firestore plugin for Flutter.
   s.dependency 'Firebase/Firestore', '~> 6.0'
   s.static_framework = true
 
-  s.prepare_command = <<-CMD
-      echo // Generated file, do not edit > Classes/UserAgent.h
-      echo "#define LIBRARY_VERSION @\\"#{libraryVersion}\\"" >> Classes/UserAgent.h
-      echo "#define LIBRARY_NAME @\\"flutter-fire-fst\\"" >> Classes/UserAgent.h
-    CMD
+  s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) LIBRARY_VERSION=@\"' + libraryVersion + '\" LIBRARY_NAME=@\"flutter-fire-fst\"' }
```
jmagman: Does this actually work?  I wouldn't have expected any @ signs everywhere.

jmagman: I was expecting something like (use Ruby string interpolation):
```
s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) LIBRARY_VERSION="#{libraryVersion}" LIBRARY_NAME=flutter-fire-fst' }
```

jmagman: Oh I see
> It wasn't converting the `#{libraryVersion} ` to the actual value so I concatenated the string.

If I recall it only works in double quotes:
```ruby
"'$(inherited) LIBRARY_VERSION=\"#{libraryVersion}\" LIBRARY_NAME=flutter-fire-fst"
```

jmagman: Need the @ to psudeo string-ize the the macro

franciscojma86: We both tried a bunch of options offline, but it's still giving some errors. We'll continue on Monday.

franciscojma86:
Those are existing failures :( I believe @collinjackson is looking into those, but we don't think those are related to this PR.

Ahmadre:
Any news? we need this asap

collinjackson:
@jmagman @franciscojma86 Thank you both SO much for improving this PR -- this approach is simpler, more reliable, and easy to maintain.

collinjackson:
The CI issue can probably be fixed with a change similar to https://github.com/flutter/plugins/pull/2592 to remove this file:

https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_crashlytics/example/ios/Runner.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings

I'll open a PR for that now.

danysz:
Someone can please fix the CLA of google and merge this issue. Is a huge waiting time this UserAgent.h missing

cbenhagen:
@danysz while you are waiting you can use this branch by adding entries like the following to your pubspec.yaml for every package you need:

```yaml
dependency_overrides:
  firebase_core:
    git:
      url: https://github.com/collinjackson/flutterfire
      path: packages/firebase_core/firebase_core
      ref: user_agent_mac
```

Or you can use a local git checkout where you merge the PRs you are waiting for.

jmagman:
This PR got messy with multiple usernames and a merge conflict.  Closing in favor of https://github.com/FirebaseExtended/flutterfire/pull/2169.

jmagman:
There are still UserAgent headers in firebase_remote_config, firebase_database, firebase_dynamic_links, firebase_messaging, firebase_ml_vision, firebase_analytics, firebase_storage, firebase_performance.

@collinjackson @franciscojma86 is there a follow up PR for them?

danysz:
> There are still UserAgent headers in firebase_remote_config, firebase_database, firebase_dynamic_links, firebase_messaging, firebase_ml_vision, firebase_analytics, firebase_storage, firebase_performance.
> 
> @collinjackson @franciscojma86 is there a follow up PR for them?

Yes...looks that things are moving really very slow here, way too slow

PR #2409, opened 2020-04-21
[firebase_analytics] add instance getter

lukepighetti:
The purpose for this change is to make it clear to the consumer of `firebase_analytics` that they don't need to maintain control over the `FirebaseAnalytics` class instance. That they can rely instead on a built in singleton instance that matches the pattern of all other `flutterfire` packages.

### DIFF
--- a/packages/firebase_analytics/firebase_analytics/lib/firebase_analytics.dart
+++ b/packages/firebase_analytics/firebase_analytics/lib/firebase_analytics.dart
@@ -13,6 +13,9 @@ const MethodChannel firebaseChannel =
 
 /// Firebase Analytics API.
 class FirebaseAnalytics {
+  FirebaseAnalytics._();
+  static final instance = FirebaseAnalytics._();
+    
   final MethodChannel _channel = firebaseChannel;
 
   /// Namespace for analytics API available on Android only.


packages/firebase_analytics/firebase_analytics/lib/firebase_analytics.dart
```diff
@@ -13,6 +13,9 @@ const MethodChannel firebaseChannel =
 
 /// Firebase Analytics API.
 class FirebaseAnalytics {
+  FirebaseAnalytics._();
+  static final instance = FirebaseAnalytics._();
```
domesticmouse, line 17: Consider using a [factory to return the singleton instance](https://stackoverflow.com/questions/12649573/how-do-you-build-a-singleton-in-dart), and thus not break current code.

lukepighetti: The intent isn't to make the default constructor return a singleton, but to match the semantics of other firebase packages by providing a static instance getter.

domesticmouse: No, I’d say I’m the one missing context. I was trying to avoid breaking APIs, but consistency is also important.

Issue #2526, opened 2020-05-10
cloud_firestore : issues causing a significant number of crashes


niypoo:
**Firebase CrashLytics  Report**
Emerging issuesNew issues causing a significant number of crashes
--
CloudFirestorePlugin.java line 106io.flutter.plugins.firebase.cloudfirestore.CloudFirestorePlugin$DocumentObserver.a

**Describe the bug**
I'm facing weird issue that make app crashing in production mode and it's happened once time and don't happened if I try same action I have did , so when try add item in firebase it's crashing and if return try same item in same condition maybe didn't crashed but I have cashed these issues

**Flutter doctor**
Running flutter doctor...
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel master, 1.19.0-1.0.pre.22, on Mac OS X 10.15.3 19D76, locale en-EG)
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 11.4)
[✓] Android Studio (installed)
[✓] Connected device (1 available)

```
CloudFirestorePlugin.java line 106
io.flutter.plugins.firebase.cloudfirestore.CloudFirestorePlugin$DocumentObserver.a

Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'void j.b.a.a.n.a(java.lang.String, java.lang.Object)' on a null object reference
io.flutter.plugins.firebase.cloudfirestore.CloudFirestorePlugin$DocumentObserver.a (CloudFirestorePlugin.java:106)
io.flutter.plugins.firebase.cloudfirestore.CloudFirestorePlugin$DocumentObserver.a (CloudFirestorePlugin.java:2)
com.google.firebase.firestore.DocumentReference.a (DocumentReference.java:87)
com.google.firebase.firestore.DocumentReference$$Lambda$3.a (DocumentReference.java:6)
com.google.firebase.firestore.core.AsyncEventListener.a (AsyncEventListener.java:6)
com.google.firebase.firestore.core.AsyncEventListener$$Lambda$1.run (AsyncEventListener.java:6)
android.os.Handler.handleCallback (Handler.java:873)
android.os.Handler.dispatchMessage (Handler.java:99)
com.google.android.gms.internal.tasks.zzb.dispatchMessage (zzb.java)
android.os.Looper.loop (Looper.java:214)
android.app.ActivityThread.main (ActivityThread.java:7156)
[...]

PR #2605, opened 2020-05-21
[firebase_auth] Clean up listener when detached and channel is null

gching:
## Description

* Remove / clean up listener when `onDetachedFromEngine` is called

### Background
* This happens when `onDetachedFromEngine` is called and when we have a at least one listener in `authStateListeners`, cause we do not remove / clean up listeners when detached, and therefore, when the listener is called after being detached, it will be `null`
* Just to highlight, this only started happening since `0.16.0` cause the package was upgraded to Android v2 embedding.

### Implementation details
* I created `PluginAuthStateListener` that implements `FirebaseAuth.AuthStateListener` that includes a `remove` method that can be called when `onDetachedFromEngine` occurs. Previously, I opted to remove the listener in the listener itself, but I couldn't really test this
* There was no way for me to mock `FirebaseApp` so for tests I had to include `Roboletric`
* I've changed the access for `authStateListeners` to be package-private so it can be accessed for tests

## Related Issues
* Fixes (hopefully) #2444. It worked for me :> but lets get others to test. However, `channel` is only set to `null` when `onDetachedFromEngine` is called so this should be the only reason why it is `null`
* **IMPORTANT** Looking into it more, it seems like some plugins (`cloud_firestore`) has possibly the same issue (ex. #2517, #2526). I actually worked on this before seeing the comments in #2526, so let me know if this is the wrong way of implementing this or if not, I can make separate PRs to address the issue for different plugins. (cc @cyanglaz @lidongze91)

## Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] If the pull request affects only one plugin, the PR title starts with the name of the plugin in brackets (e.g. [cloud_firestore])
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[...]

PR #2650, opened 2020-05-27
fix crash where listeners are not removed when app quits

lidongze91:
This should solve #2526


lidongze91:
> LGTM, please update the version in pubspec and changelog

Updated.

lidongze91:
> Is there a way we can test this? Maybe in a Junit test?

Do we have existing junit test already? Or you would like me to add a new one.
(I don't see the test files within android dir)

lidongze91:
@gching Could you do a code review when you get the chance? Thanks.

lidongze91:
> LGTM

@bparrishMines Can you also merge it? I don't seem to have the access. Much Thanks.

gching:
@lidongze91 Did we decide not to write unit tests for this PR?

bparrishMines:
@cyanglaz  @kroikie I merged this without tests since its a trivial change and we would require java unit tests or expresso tests.

gching:
@bparrishMines If possible, could you look into my PR #2605 for `firebase_auth`, it is a similar change but I wrote it with tests

PR #2869, opened 2020-06-30
[firebase_ml] Core android functionality

PolinaGo:
## Description

Implemented core functionality of the plugin on android:
- downloading models
- checking if model is downloaded
- getting latest model file

Current example app downloads a TF Lite model from Firebase and performs inference using tflite 3p plugin.

Current limitations of implementation on android side (to be added later):
- no support for specified FirebaseApp
- no support for deleting custom models from device

Testing:
- Unit-testing of FirebaseModelManager
- Integration testing all core functions
- Instrumented testing on Android with e2e plugin

Current limitations of testing:
- No unit-testing on java code, on majority of dart code as per current practices

NOTE: Some checks are failing because tests are ran on both android and ios. I can explicitly run tests on android only to pass these checks in needed.


packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
+        remoteModel = new FirebaseCustomRemoteModel.Builder(modelName).build();
+
+        FirebaseModelManager.getInstance()
+            .isModelDownloaded(remoteModel)
+            .addOnCompleteListener(
+                new OnCompleteListener<Boolean>() {
+                  @Override
+                  public void onComplete(@NonNull Task<Boolean> task) {
+                    Boolean isModelDownloaded = task.getResult();
+                    if (isModelDownloaded != null) result.success(isModelDownloaded);
```
nauthadr, line 116: Must include a bracket even if single statement. This may apply in many places in this PR.

https://screenshot.googleplex.com/ZDyUV6pPXYg
go/java-style#s4.1.1-braces-always-used

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
     channel.setMethodCallHandler(this);
   }
 
+  @RequiresApi(api = Build.VERSION_CODES.N)
   @Override
-  public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
+  public void onMethodCall(@NonNull MethodCall call, @NonNull final Result result) {
+    final FirebaseCustomRemoteModel remoteModel;
+    String modelName;
```
nauthadr, line 57: Since these are not used in the wider scope it's preferred to declare separate local variables in limited scope as they are needed.

https://screenshot.googleplex.com/gTaOBnWeu8h
go/java-style#s4.8.2.2-variables-limited-scope

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
+            .getLatestModelFile(remoteModel)
+            .addOnCompleteListener(
+                new OnCompleteListener<File>() {
+                  @Override
+                  public void onComplete(@NonNull Task<File> task) {
+                    File modelFile = task.getResult();
+                    if (modelFile != null) result.success(modelFile.getAbsolutePath());
+                    else task.getException().printStackTrace();
+                  }
+                });
```
nauthadr, line 101: optional: I'm not sure if you are building on java 8, but if so you may optionally replace the anonymous inner class with a lambda.

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
+
+        FirebaseModelManager.getInstance()
+            .getLatestModelFile(remoteModel)
+            .addOnCompleteListener(
+                new OnCompleteListener<File>() {
+                  @Override
+                  public void onComplete(@NonNull Task<File> task) {
+                    File modelFile = task.getResult();
+                    if (modelFile != null) result.success(modelFile.getAbsolutePath());
+                    else task.getException().printStackTrace();
```
nauthadr, line 99: Should a failure like this be propagated through the channel via result.error? Also, is it possible for task.getResult to be null here (no model file) but there is no exception on the task? If so line 99 would throw a NullPointerException.

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
     channel.setMethodCallHandler(this);
   }
 
+  @RequiresApi(api = Build.VERSION_CODES.N)
   @Override
-  public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
+  public void onMethodCall(@NonNull MethodCall call, @NonNull final Result result) {
```
nauthadr, line 55: For a large method like this with isolated pieces of logic it is preferred to break it up via private helper methods.

https://screenshot.googleplex.com/JSV0MchbpZc
go/java-practices/methods#length

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
+        if (conditionsToMap.get("requireWifi")) conditionsBuilder.requireWifi();
+
+        FirebaseModelDownloadConditions conditions = conditionsBuilder.build();
+        FirebaseModelManager.getInstance()
+            .download(remoteModel, conditions)
+            .addOnCompleteListener(
+                new OnCompleteListener<Void>() {
+                  @Override
+                  public void onComplete(@NonNull Task<Void> task) {
+                    result.success(remoteModelToMap(remoteModel));
```
nauthadr, line 80: Is there any error handling necessary here for when task contains an exception?

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
+      case "FirebaseModelManager#download":
+        assert (call.argument("model") != null);
+        assert (call.argument("conditions") != null);
+
+        modelName = call.argument("model");
+        Map<String, Boolean> conditionsToMap = call.argument("conditions");
+        remoteModel = new FirebaseCustomRemoteModel.Builder(modelName).build();
+        FirebaseModelDownloadConditions.Builder conditionsBuilder =
+            new FirebaseModelDownloadConditions.Builder();
+        if (conditionsToMap.get("requireCharging")) conditionsBuilder.requireCharging();
```
nauthadr, line 69: @bparrishMines Just curious, for flutter channels is there a more type safe way of passing arguments and return values back (ie via protobufs). Or is it usually done with primitives like strings etc? Just curious about typical convention, and not a blocker for this PR. Thanks

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: https://firebase.google.com/docs/ml-kit/android/use-custom-models this documentation suggests there is no failure case.

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
```diff
@@ -40,21 +50,86 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
-  public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
+  public void onMethodCall(@NonNull MethodCall call, @NonNull final Result result) {
+    final FirebaseCustomRemoteModel remoteModel;
+    String modelName;
+
     switch (call.method) {
-      case "downloadRemoteModel":
-        result.success("Testing call. Will be removed shortly");
+      case "FirebaseModelManager#download":
+        assert (call.argument("model") != null);
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: It does throw me an error. I would prefer to keep it to be consistent with other implementations of plugins.

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: Done. In fact, new class ModelManager is cleated (similar to Detector in ML Vision) to handle these method calls.

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
PolinaGo: I am not sure if it fully answers your question, but

"The standard platform channels use a standard message codec that supports efficient binary serialization of simple JSON-like values, such as booleans, numbers, Strings, byte buffers, and Lists and Maps of these (see [StandardMessageCodec](https://api.flutter.dev/flutter/services/StandardMessageCodec-class.html)) for details). The serialization and deserialization of these values to and from messages happens automatically when you send and receive values."

The full guide on channels is [here](https://flutter.dev/docs/development/platform-integration/platform-channels).

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -18,32 +19,35 @@ import 'firebase_model_download_conditions.dart';
...
 
   /// Initiates the download of remoteModel if the download hasn't begun.
   Future<void> download(FirebaseRemoteModel model,
       FirebaseModelDownloadConditions conditions) async {
-    // Example call
-    await _channel.invokeMethod(
-        "FirebaseModelManager#download", {'model': model.toString()});
+    Map modelMap = await channel.invokeMethod("FirebaseModelManager#download",
+        {'model': model.modelName, 'conditions': conditions.toMap()});
+    model.modelHash = modelMap['modelHash'];
```
elvisun, line 38: It feels like this could be done inside the channel call?

I don't feel strongly about this but maybe document why we are manually adding hash to the model?

packages/firebase_ml/lib/src/model_management/firebase_remote_model.dart
```diff
@@ -10,10 +10,23 @@ abstract class FirebaseRemoteModel {
   /// Constructor for [FirebaseRemoteModel].
   ///
   /// Called only by classes that extend [FirebaseRemoteModel].
-  const FirebaseRemoteModel(this.modelName) : assert(modelName != null);
+  FirebaseRemoteModel(this.modelName) : assert(modelName != null);
 
   /// Name associated with remote model in the Firebase console.
   final String modelName;
+
+  /// Hash associated with remote model in the Firebase console.
```
elvisun, line 18: maybe document how the hash is used

packages/firebase_ml/lib/src/model_management/firebase_remote_model.dart
```diff
@@ -10,10 +10,23 @@ abstract class FirebaseRemoteModel {
...
+
+  /// Express download conditions via map.
+  ///
+  /// This method is used for ease of transfer via channel and printing.
+  Map toMap() {
+    var remoteModelToMap = Map();
+    remoteModelToMap['modelName'] = this.modelName;
+    remoteModelToMap['modelHash'] = this.modelHash;
+    return remoteModelToMap;
+  }
```
elvisun, line 29: same as the other `toMap` above

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -18,32 +19,35 @@ import 'firebase_model_download_conditions.dart';
...
+  static final FirebaseModelManager instance = FirebaseModelManager._();
 
   /// Initiates the download of remoteModel if the download hasn't begun.
   Future<void> download(FirebaseRemoteModel model,
       FirebaseModelDownloadConditions conditions) async {
-    // Example call
-    await _channel.invokeMethod(
-        "FirebaseModelManager#download", {'model': model.toString()});
+    Map modelMap = await channel.invokeMethod("FirebaseModelManager#download",
+        {'model': model.modelName, 'conditions': conditions.toMap()});
```
elvisun: we are only using `model.modelName`, should we just take in a `String` of modelName only?

That way we can make FirebaseRemoteModel immutable by making modelHash final, and we just return the new `FirebaseRemoteModel`?

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -18,32 +19,35 @@ import 'firebase_model_download_conditions.dart';
...
+    return File(modelPath);
   }
 
   /// Returns whether the given [FirebaseRemoteModel] is currently downloaded.
-  Future<bool> isModelDownloaded(
-      FirebaseRemoteModel firebaseRemoteModel) async {
-    return true;
+  Future<bool> isModelDownloaded(FirebaseRemoteModel model) async {
+    return await channel.invokeMethod(
+        "FirebaseModelManager#isModelDownloaded", {'model': model.modelName});
```
elvisun: Does this need to check against modelHash to make sure that version of the model is downloaded?

packages/firebase_ml/test/firebase_ml_test.dart
```diff
@@ -0,0 +1,89 @@
...
+    final FirebaseModelManager modelManager = FirebaseModelManager.instance;
+
+    setUp(() {
+      FirebaseModelManager.channel
+          .setMockMethodCallHandler((MethodCall methodCall) async {
+        log.add(methodCall);
+
+        switch (methodCall.method) {
+          case 'FirebaseModelManager#download':
+            var model = <String, String>{};
```
elvisun, line 26: same here 
`return {...}`

packages/firebase_ml/test/firebase_ml_test.dart
```diff
@@ -0,0 +1,89 @@
...
+      FirebaseModelManager.channel.setMockMethodCallHandler(null);
+    });
+
+    test('valid sequence of calls', () async {
+      expect(modelManager, isNotNull);
+      FirebaseCustomRemoteModel model = FirebaseCustomRemoteModel(MODEL_NAME);
+      expect(model.modelHash, isNull);
+      FirebaseModelDownloadConditions conditions =
+          FirebaseModelDownloadConditions(requireWifi: true);
+      expect(conditions, isNotNull);
```
elvisun, line 51: not really related to the flow of downloading a model, it's more of a test for making sure these classes work as intended, maybe pull them into separate tests?

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/FirebaseMLPlugin.java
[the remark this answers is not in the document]
PolinaGo: This was in the old version, doesn't exist after "Java restructuring" commit (5b4b0b2)

packages/firebase_ml/example/android/app/google-services.json
```diff
@@ -0,0 +1,235 @@
+{
```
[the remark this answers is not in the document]
PolinaGo: It's a fake file to compile

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
PolinaGo: This was in the old version, doesn't exist after "Java restructuring" commit (5b4b0b2)

packages/firebase_ml/test/firebase_ml_test.dart
PolinaGo: saw two examples of tests in different plugins, one doing altogether, another separately. This was you know which statement fails (because you follow expects along the way). I have no strong preference for either way, can change.

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
PolinaGo: It doesn't explicitly say in the documentation if the model is checked by the model hash. To me it seems like it's checked by the name only. Would be useful to check in the SDK. I can also pass hash just in case.
"Downloads the given model from the server to a local directory on the device. Use isModelDownloaded(_:) to check the download status for the model. If this method is invoked and the model has already been downloaded, a request is made to check if a newer version of the model is available for download. If available, the new version of the model is downloaded."
-- from IOS docs

packages/firebase_ml/android/user-agent.gradle
```diff
@@ -0,0 +1,22 @@
...
+import java.util.regex.Pattern
+
+String libraryVersionName = "UNKNOWN"
+String libraryName = "flutter-fire-ml"
+File pubspec = new File(project.projectDir.parentFile,  'pubspec.yaml')
+
+if (pubspec.exists()) {
+    String yaml = pubspec.text
+    // Using \s*['|"]?([^\n|'|"]*)['|"]? to extract version number.
+    Matcher versionMatcher = Pattern.compile("^version:\\s*['|\"]?([^\\n|'|\"]*)['|\"]?\$", Pattern.MULTILINE).matcher(yaml)
```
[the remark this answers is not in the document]
PolinaGo: This file exists in every plugin in the repo, and it has something to do with the proper connection to firebase I believe. I am not entirely sure what it does though.

packages/firebase_ml/android/user-agent.gradle
```diff
@@ -0,0 +1,22 @@
...
+    if (versionMatcher.find()) libraryVersionName = versionMatcher.group(1).replaceAll("\\+", "-")
```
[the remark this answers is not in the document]
PolinaGo: Same here:
This file exists in every plugin in the repo, and it has something to do with the proper connection to firebase I believe. I am not entirely sure what it does though.

packages/firebase_ml/test/firebase_ml_test.dart
PolinaGo: What do you mean by return {...}?

packages/firebase_ml/android/user-agent.gradle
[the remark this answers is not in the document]
elvisun: well in that case I guess it's ok if it's standard across all plugins

packages/firebase_ml/example/android/app/google-services.json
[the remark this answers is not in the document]
elvisun: ok sgtm

packages/firebase_ml/test/firebase_ml_test.dart
elvisun: return {'modelName': MODEL_NAME, ...}

packages/firebase_ml/test/firebase_ml_test.dart
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_remote_model.dart
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
 }
 
 class _MyAppState extends State<MyApp> {
-  String _pluginOutput = 'Unknown';
+  File _image;
+  List _labels;
+
+  Future predictImagePicker() async {
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/test/firebase_ml_test.dart
elvisun: If you break them down into smaller tests it's even easier to see what fails because you'll see it in the test description without even having to follow the expects

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
elvisun: from the documentation it doesn't look like modelHash is used, maybe worth double checking. I think the API would be nicer if we can just use `modelName` alone in the input params

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
 
-  Future<void> initPlatformState() async {
-    String pluginOutput;
-    try {
-      pluginOutput = await FirebaseML.doSomething;
-    } on PlatformException {
-      pluginOutput = 'Failed to get plugin working.';
-    }
-    if (!mounted) return;
+  Future loadModelFromFirebase() async {
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
+    FirebaseCustomRemoteModel model =
+        FirebaseCustomRemoteModel('image_classification');
```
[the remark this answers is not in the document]
PolinaGo: This is in the FirebaseCustomRemoteModel class description

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
```diff
@@ -0,0 +1,30 @@
...
+
+      await modelManager.download(model, conditions);
+      expect(model.modelHash, isNotNull);
+
+      var isModelDownloaded = await modelManager.isModelDownloaded(model);
+      expect(isModelDownloaded, isTrue);
+
+      var modelFile = await modelManager.getLatestModelFile(model);
+      expect(modelFile, isNotNull);
+      expect(modelFile.path.runtimeType, ''.runtimeType);
```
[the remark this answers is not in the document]
PolinaGo: Yes, but these are two separate checks.
If modelFile is null, it will fail on calling .path, something that I want to check with different expect.

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
PolinaGo: Hash gets populated by the SDK, and we agreed it should be present in a model. If it is added inside channel call, I will need to pass the whole model. Because only hash is changed, I decided to return only hash.

kroikie:
@PolinaGo @elvisun Since this would be a new plugin. Would it be possible for you to define the interface that will be presented to Flutter devs so that it can be reviewed by the Firebase API counsel?

PolinaGo:
I don't see any problem in defining the interface for the Firebase API counsel. I am currently updating the design doc (draft seen by @bparrishMines). After review from @nauthadr and @elvisun I can share it with you and the counsel.

packages/firebase_ml/test/firebase_ml_test.dart
PolinaGo: Done

packages/firebase_ml/test/firebase_ml_test.dart
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
PolinaGo: I would keep it as it is to be consistent with native APIs. I would really like to know why Java and Obj-C APIs require full model and not only the name.
If we end up changing it, it would make sense to change getLatestModelFile as well.
I would appreciate @bparrishMines and @nauthadr input on this.

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
 }
 
 class _MyAppState extends State<MyApp> {
-  String _pluginOutput = 'Unknown';
+  File _image;
+  List _labels;
+
+  Future predictImagePicker() async {
+    final image = await ImagePicker.pickImage(source: ImageSource.gallery);
+    if (image != null) {
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
+    FirebaseCustomRemoteModel model =
+        FirebaseCustomRemoteModel('image_classification');
+
+    FirebaseModelDownloadConditions conditions =
+        FirebaseModelDownloadConditions(requireWifi: true);
+    FirebaseModelManager modelManager = FirebaseModelManager.instance;
+
+    await modelManager.download(model, conditions);
+
+    var isModelDownloaded = await modelManager.isModelDownloaded(model);
```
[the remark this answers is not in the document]
PolinaGo: Yes, you can. Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
+    var modelFile = await modelManager.getLatestModelFile(model);
+    assert(modelFile != null);
+
+    var appDirectory = await getApplicationDocumentsDirectory();
+    var labelsData =
+        await rootBundle.load("assets/labels_mobilenet_v1_224.txt");
+    var labelsFile =
+        await File(appDirectory.path + "_labels_mobilenet_v1_224.txt")
+            .writeAsBytes(labelsData.buffer.asUint8List(
+                labelsData.offsetInBytes, labelsData.lengthInBytes));
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
+        body: Column(children: [
+          _image != null ? Image.file(_image) : Text('No image selected.'),
+          Column(
+            children: _labels != null
+                ? _labels.map((res) {
+                    return Text("${res["label"]}");
+                  }).toList()
+                : [],
+          ),
+        ]),
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
 }
 
 class _MyAppState extends State<MyApp> {
-  String _pluginOutput = 'Unknown';
+  File _image;
+  List _labels;
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
+        "success");
   }
 
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
       home: Scaffold(
         appBar: AppBar(
-          title: const Text('Plugin example app'),
+          title: const Text('FirebaseML example app'),
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/test_driver/firebase_ml_test.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_driver/flutter_driver.dart';
+
+void main() async {
+  final FlutterDriver driver = await FlutterDriver.connect();
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_download_conditions.dart
```diff
@@ -13,8 +13,19 @@ class FirebaseModelDownloadConditions {
...
-    this.requireWifi = false,
     this.requireDeviceIdle = false,
     this.requireCharging = false,
+    this.requireWifi = false,
   });
+
+  /// Express download conditions via map.
+  ///
+  /// This method is used for ease of transfer via channel and printing.
+  Map toMap() {
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_download_conditions.dart
```diff
@@ -13,8 +13,19 @@ class FirebaseModelDownloadConditions {
...
+    var conditionsToMap = Map();
+    conditionsToMap['requireCharging'] = this.requireCharging;
+    conditionsToMap['requireDeviceIdle'] = this.requireDeviceIdle;
+    conditionsToMap['requireWifi'] = this.requireWifi;
+    return conditionsToMap;
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_download_conditions.dart
```diff
@@ -13,8 +13,19 @@ class FirebaseModelDownloadConditions {
...
     this.requireCharging = false,
+    this.requireWifi = false,
   });
+
+  /// Express download conditions via map.
+  ///
+  /// This method is used for ease of transfer via channel and printing.
+  Map toMap() {
+    var conditionsToMap = Map();
+    conditionsToMap['requireCharging'] = this.requireCharging;
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -18,32 +19,35 @@ import 'firebase_model_download_conditions.dart';
 /// https://firebase.google.com/docs/reference/android/com/google/
 /// firebase/ml/common/modeldownload/FirebaseModelManager
 class FirebaseModelManager {
-  static const MethodChannel _channel =
+  FirebaseModelManager._();
+
+  @visibleForTesting
+
+  /// Means for communication with native platform code
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -18,32 +19,35 @@ import 'firebase_model_download_conditions.dart';
...
-  static final FirebaseModelManager instance = FirebaseModelManager();
+  static final FirebaseModelManager instance = FirebaseModelManager._();
 
   /// Initiates the download of remoteModel if the download hasn't begun.
   Future<void> download(FirebaseRemoteModel model,
       FirebaseModelDownloadConditions conditions) async {
-    // Example call
-    await _channel.invokeMethod(
-        "FirebaseModelManager#download", {'model': model.toString()});
+    Map modelMap = await channel.invokeMethod("FirebaseModelManager#download",
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/test_driver/firebase_ml.dart
```diff
@@ -0,0 +1,18 @@
...
+
+import 'dart:async';
+import 'package:flutter_driver/driver_extension.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:firebase_ml/firebase_ml.dart';
+
+part 'firebase_model_manager.dart';
+
+void main() {
+  final Completer<String> completer = Completer<String>();
```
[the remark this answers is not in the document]
PolinaGo: Done

kroikie:
Much appreciated, thanks!

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
nauthadr: I would be in favour of being consistent with the native APIs. Also, if we are undergoing a Firebase counsel API review may make sense to leave as is in this PR for now and then change accordingly based on feedback from the API review.

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
         ),
-        body: Center(
-          child: Text('Plugin output: $_pluginOutput\n'),
+        body: Column(children: [
+          _image != null ? Image.file(_image) : Text('No image selected.'),
+          Column(
+            children: _labels != null
+                ? _labels.map((res) {
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
+        _image = image;
+      });
+    }
+  }
 
   @override
   void initState() {
     super.initState();
-    initPlatformState();
+    loadModelFromFirebase();
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -18,32 +19,35 @@ import 'firebase_model_download_conditions.dart';
...
+        "FirebaseModelManager#getLatestModelFile", {'model': model.modelName});
+    return File(modelPath);
   }
 
   /// Returns whether the given [FirebaseRemoteModel] is currently downloaded.
-  Future<bool> isModelDownloaded(
-      FirebaseRemoteModel firebaseRemoteModel) async {
-    return true;
+  Future<bool> isModelDownloaded(FirebaseRemoteModel model) async {
+    return await channel.invokeMethod(
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
elvisun: +1 for waiting for the API review

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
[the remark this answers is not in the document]
elvisun: oh sorry I meant on line 27, can we do something like `expect(modelFile.path, 'abc')`? This will cover the runtimeType check as well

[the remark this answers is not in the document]
PolinaGo: Not really, because it is a real file that is downloaded, and, I am not sure how is path is generated. Previously it was abstracted away, and only now users can access the file with the latest getLatestFile API.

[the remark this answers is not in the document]
PolinaGo: I can be more thorough, however, on what it should contain, such as name of the model (at some point)

packages/firebase_ml/example/lib/main.dart
```diff
@@ -15,37 +18,80 @@ class MyApp extends StatefulWidget {
...
-        body: Center(
-          child: Text('Plugin output: $_pluginOutput\n'),
+        body: Column(children: [
+          _image != null ? Image.file(_image) : Text('No image selected.'),
+          Column(
+            children: _labels != null
+                ? _labels.map((res) {
+                    return Text("${res["label"]}");
+                  }).toList()
+                : [],
```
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/lib/src/model_management/firebase_remote_model.dart
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/ModelManager.java
```diff
@@ -0,0 +1,132 @@
...
+import com.google.firebase.ml.common.modeldownload.FirebaseModelDownloadConditions;
+import com.google.firebase.ml.common.modeldownload.FirebaseModelManager;
+import com.google.firebase.ml.custom.FirebaseCustomRemoteModel;
+import io.flutter.plugin.common.MethodCall;
+import io.flutter.plugin.common.MethodChannel.Result;
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ModelManager {
```
nauthadr, line 19: nit: For a utility class (where all members are static), it is recommended to add a private constructor to prevent subclassing and instantiating. You can also mark the class as final as well.

https://screenshot.googleplex.com/5MaeRMfBPHx

packages/firebase_ml/test/firebase_ml_test.dart
```diff
@@ -0,0 +1,109 @@
...
+        }
+      });
+      log.clear();
+    });
+
+    tearDown(() {
+      FirebaseModelManager.channel.setMockMethodCallHandler(null);
+    });
+
+    test('download', () async {
```
nauthadr, line 45: I noticed 2 of the tests here have the same name 'download'. I'm not sure the best practices for these tests specifically, but we have some guidelines around naming for dart tests. Please take a look here:

https://screenshot.googleplex.com/QxqE2XLAbJv

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/ModelManager.java
```diff
@@ -0,0 +1,132 @@
...
+                } else {
+                  result.error(
+                      "FirebaseModelManager", task.getException().getLocalizedMessage(), null);
+                }
+              }
+            });
+  }
+
+  private static Map<String, String> remoteModelToMap(FirebaseCustomRemoteModel model) {
+    Map remoteModelToMap = new HashMap<String, String>();
```
nauthadr, line 127: Should define the types of the Map's generics to avoid unchecked calls. Can also used inferred generics when instantiating. ie:
Map<String, String> remoteModelToMap = new HashMap<>();

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/ModelManager.java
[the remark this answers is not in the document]
PolinaGo: Done

packages/firebase_ml/test/firebase_ml_test.dart
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/ModelManager.java
PolinaGo: Done

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/ModelManager.java
PolinaGo: Done

packages/firebase_ml/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -38,10 +40,12 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
         </activity>
-        <!-- Don't delete the meta-data below.
```
bparrishMines, line 41: Is there a reason you're using the v1 embedding instead of v2?

PolinaGo: My understanding is that this way I  test v1 embedding’s compatibility with my plugin. Same thing is done in all other packages in the repo.

bparrishMines: I think you have to include both Activities in the `AndroidManifest` like in the `firebase_ml_vision` plugin: https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_ml_vision/example/android/app/src/main/AndroidManifest.xml#L20

PR #2929, opened 2020-07-10
fix(firestore): normalize Firestore pointer path

IchordeDionysos:
## Description

Equality checking of references is not user input safe as the internal `Pointer` check unnormalized paths. 

## Related Issues

#2928

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] If the pull request affects only one plugin, the PR title starts with the name of the plugin in brackets (e.g. [cloud_firestore])
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] NOT NEEDED: I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] NOT NEEDED: I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [ ] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/FirebaseExtended/flutterfire/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2931, opened 2020-07-10
fix(firestore): collection and document equality checking

creativecreatorormaybenot:
## Description

This expands on #2929: I think the collection reference was missing the `==` override and it also seems to make more sense to check if the pointers are equal rather than the paths, which are derived from the pointers. Essentially, this makes changing the `Pointer` behavior more easily.

Additionally, I added some tests that ensure that no regression occurs. The changelog entries that @IchordeDionysos added already reflect exactly what these changes do, so there is no need for another entry.

## Related Issues

#2929

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/cloud_firestore/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/cloud_firestore/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - **FIX**: Collection reference `==` operator.
 
-## [UNPUBLISHED]
+## 0.14.0-dev.1
 
 Along with the below changes, the plugin has undergone a quality of life update to better support exceptions thrown. Any Firestore specific errors now return a `FirebaseException`, allowing you to directly access the code (e.g. `permission-denied`) and message.
 
--- a/packages/cloud_firestore/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/cloud_firestore/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## [UNPUBLISHED]
 
-- **FIX**: Collection reference `==` operator.
-
+- **FIX**: Added `==` operator override to `CollectionReferencePlatform`.
 ## 0.14.0-dev.1
 
 Along with the below changes, the plugin has undergone a quality of life update to better support exceptions thrown. Any Firestore specific errors now return a `FirebaseException`, allowing you to directly access the code (e.g. `permission-denied`) and message.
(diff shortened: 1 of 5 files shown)


creativecreatorormaybenot:
@Salakar Is the `doc4` failing test caused by me? I do not exactly understand how it would cause that test to fail 😶

Salakar:
> @Salakar Is the `doc4` failing test caused by me? I do not exactly understand how it would cause that test to fail 😶

Don't think so, we're aware of a potential race condition on the web tests that we're looking into.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/platform_interface/platform_interface_collection_reference.dart
```diff
@@ -48,4 +48,16 @@ abstract class CollectionReferencePlatform extends QueryPlatform {
   DocumentReferencePlatform doc([String path]) {
     throw UnimplementedError("doc() is not implemented");
   }
+
+  @override
+  bool operator ==(dynamic o) =>
+      o is CollectionReferencePlatform &&
+      o.firestore == firestore &&
+      o._pointer == _pointer;
```
Ehesp, line 56: I think checking the path is better here, since it removes a dependency on the Pointer class and matches a property local to the current class.

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/platform_interface/platform_interface_collection_reference.dart
```diff
@@ -48,4 +48,16 @@ abstract class CollectionReferencePlatform extends QueryPlatform {
...
+
+  @override
+  int get hashCode => _pointer.hashCode;
```
Ehesp, line 59: Same as above, using `path.hashCode` seems like a better option to me since it removes depending on a different class.

creativecreatorormaybenot:
@Ehesp done.

packages/cloud_firestore/cloud_firestore/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## [UNPUBLISHED]
 
+- **FIX**: Collection reference `==` operator.
```
Ehesp, line 3: Could this be changed to the following please?

```
- **FIX**: `CollectionReferencePlatform` add "==" operator.
```

Just makes it more explicit it's on the platform interface. Thanks!

creativecreatorormaybenot: done

PR #2950, opened 2020-07-13
[firebase_ml] Added IOS support (part 1)

PolinaGo:
## Description

Added Obj-C and other supporting files for IOS support.

Please note that in current state apps on Android work fully.
IOS does not work because several interface changes are required to the dart side. IOS functionality is broken into 2 pull requests (Obj-C/Dart sides), and the next PR will make this plugin fully functional on IOS.


packages/firebase_ml/example/pubspec.yaml
```diff
@@ -18,7 +18,7 @@ dependencies:
   cupertino_icons: ^0.1.3
 
   image_picker: ^0.6.7
-  tflite: 1.0.6
+  tflite: 1.1.1
```
PolinaGo, line 21: Change required for the code to compile. For some reason 1.0.6 is totally broken on IOS

packages/firebase_ml/ios/Classes/FLTModelManager.m
```diff
@@ -0,0 +1,97 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FLTFirebaseMLPlugin.h"
+
+@implementation FLTModelManager
+
++ (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
```
kroikie, line 9: Is there any support for [deleteDownloadModel](https://firebase.google.com/docs/reference/ios/firebasemlcommon/api/reference/Classes/FIRModelManager#-deletedownloadedmodel:completion:)?

PolinaGo: As per design doc, deleteDownloadModel had a bug to be fixed before I could implemented. I know that bug was fixed, but I'm not sure if it is in the production yet. 

Either way Android side doesn't have this support at the moment. If we end up implementing deleteDownloadModel for flutter, it will be in a separate PR and will contain IOS, Android and Dart portions altogether.

PolinaGo: I just confirmed that the deleteDownloadModel native SDK fix push got delayed, and should be available early August.

kroikie: Sounds good, thanks for checking on that.

kroikie:
I'll merge this in and we can work on migrating it to `firebase_ml_custom` in a follow up PR.

PR #3010, opened 2020-07-23
[firebase_ml_vision] Bug fix: if confidence is 1 or 0 it is decoded as int

PolinaGo:
## Description

When `confidence` (used by `ImageLabel` and `TextContainer`) is 1 or 0 and the value is passed over the channel, on the Dart side it is decoded as `int`, which leads to the following error:
```
Unhandled Exception: type 'int' is not a subtype of type 'double'
```

This fix forces the value to be interpreted as `double`.

## Related Issues

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
+++ b/packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
@@ -795,7 +795,7 @@ void main() {
                 'languageCode': 'cd',
               }
             ],
-            'confidence': 0,
+            'confidence': 0.1,
           },
           <dynamic, dynamic>{
             'text': 'my',
@@ -808,7 +808,7 @@ void main() {
               <dynamic>[8.0, 7.0],
             ],
             'recognizedLanguages': <dynamic>[],
-            'confidence': 1,
+            'confidence': 0.2,
           },
         ];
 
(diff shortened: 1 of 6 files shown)


packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
```diff
@@ -795,7 +795,7 @@ void main() {
                 'languageCode': 'cd',
               }
             ],
-            'confidence': 0.1,
+            'confidence': 0,
```
kroikie, line 798: Should we still have tests of confidence being `0.1`?

packages/firebase_ml_vision/test/image_labeler_test.dart
```diff
@@ -33,7 +33,7 @@ void main() {
       test('processImage', () async {
         final List<dynamic> labelData = <dynamic>[
           <dynamic, dynamic>{
-            'confidence': 0.6,
+            'confidence': 1,
```
kroikie, line 36: Should we still have a test for `0.6` confidence?

packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
PolinaGo: Validity of confidence was checked multiple times with various non-corner case values.
However, it is probably better to add new test cases instead of modifying old ones to avoid introducing more errors by accident. 

Confidence here is now changed to the original value.

packages/firebase_ml_vision/test/image_labeler_test.dart
PolinaGo: Same as above, 

Confidence here is now changed to the original value.

packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
kroikie: Thanks for updating, feel free to remove tests that you believe are unnecessary or redundant in a follow up PR.

PR #3036, opened 2020-07-27
[firebase_ml_custom] Interface changes and cleanup

PolinaGo:
## Description

This PR addresses the latest changes to the interface of the plugin according to the design documentation (except final renaming to firebase-ml-custom) and inconsistencies between native platform implementations.

### Main changes:
- Added two new IOS download conditions and platform specific prefixes to to android download conditions in `FirebaseModelDownloadConditions`
- Removed hash from `FirebaseRemoteModel` class
- Added dart unit and integration testing to check for the missing behaviours 

**This version provides full intended functionality of the plugin. Consider this code a near final version of the plugin first release.**


packages/firebase_ml/example/ios/Podfile
```diff
@@ -64,5 +64,3 @@ post_install do |installer|
     end
   end
 end
-
-pod 'Firebase/MLModelInterpreter'
```
PolinaGo, line 68: Unnecessary dependency that was mistakenly added in the last commit.

packages/firebase_ml/example/lib/main.dart
```diff
@@ -35,6 +35,7 @@ class _MyAppState extends State<MyApp> {
     }
     var labels = List<Map>.from(await Tflite.runModelOnImage(
       path: image.path,
+      imageStd: 127.5,
```
PolinaGo, line 38: Provides more accurate image labels when the user explores example app.

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
```diff
@@ -7,17 +7,16 @@ part of 'firebase_ml.dart';
...
     final FirebaseModelManager modelManager = FirebaseModelManager.instance;
-    final String MODEL_NAME = 'image_classification';
+    final String MODEL_NAME = 'myModelName';
+    final String INVALID_MODEL_NAME = 'invalidModelName';
 
-    test('downloadModel', () async {
-      FirebaseCustomRemoteModel model = FirebaseCustomRemoteModel(MODEL_NAME);
-      assert(model != null);
+    test('downloadModel and get its file', () async {
+      var model = FirebaseCustomRemoteModel(MODEL_NAME);
```
kroikie, line 14: If these are not changing can we use `final` here and in other places instead of `var`?

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
```diff
@@ -7,17 +7,16 @@ part of 'firebase_ml.dart';
 void firebaseModelManagerTest() {
   group('$FirebaseModelManager', () {
     final FirebaseModelManager modelManager = FirebaseModelManager.instance;
-    final String MODEL_NAME = 'image_classification';
+    final String MODEL_NAME = 'myModelName';
```
elvisun, line 10: nit: `String` is not needed since it's initialized

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
```diff
@@ -26,5 +25,21 @@ void firebaseModelManagerTest() {
...
       expect(modelFile.path.contains(MODEL_NAME), isTrue);
     });
+
+    test('throw an error when model is not downloaded', () async {
+      var model = FirebaseCustomRemoteModel(INVALID_MODEL_NAME);
+
+      var conditions = FirebaseModelDownloadConditions();
+
+      expect(() => modelManager.download(model, conditions),
+          throwsA(isA<PlatformException>()));
```
elvisun, line 35: are we able to expect on the error message too?

packages/firebase_ml/lib/src/model_management/firebase_model_download_conditions.dart
```diff
@@ -1,31 +1,46 @@
 /// Conditions to download remote models.
 class FirebaseModelDownloadConditions {
-  /// Boolean value that indicates if wifi is required.
-  final bool requireWifi;
+  /// Android: Boolean that indicates if wifi is required.
```
elvisun, line 3: most of this can be inferred from the actual code, I'd change it to something like this or maybe better just merge all of them into one saying they are for android only

```
/// Android only: if wifi is required to download model.
```

elvisun: https://screenshot.googleplex.com/5MackpyCuR1
from 
https://www.investigatii.md/uploads/resurse/Clean_Code.pdf

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
```diff
@@ -30,24 +30,31 @@ class FirebaseModelManager {
   static final FirebaseModelManager instance = FirebaseModelManager._();
 
   /// Initiates the download of remoteModel if the download hasn't begun.
+  ///
+  /// If the model is already downloaded to the device, and there is no update, the task will immediately succeed.
+  /// If the model is already downloaded to the device, and there is update, a download for the updated version will be attempted.
```
elvisun, line 35: maybe also document what happens if updated version download fails?

packages/firebase_ml/lib/src/model_management/firebase_model_download_conditions.dart
PolinaGo: Thank you. 
In flutter all public members have to be documented, so I will not merge all of these comments into one saying. 
I will follow your suggestion on removing redundancy in each line.

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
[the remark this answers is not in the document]
PolinaGo: Here is [Java documentation](https://firebase.google.com/docs/reference/android/com/google/firebase/ml/common/modeldownload/FirebaseModelManager) for this method, from where I took these lines.

I believe the model refers to `remoteModel` from the previous sentence, therefore **the** is used.

I agree with this logic.

PolinaGo: Flutter team tries to stay consistent with native implementations and documentation.
Current native (java/obj-c) documentation does not explicitly specify what happens if updated version download fails. 
I don't think we should do it either. 

@kroikie may provide more input on whether we should diverge from native documentation.

packages/firebase_ml/test/firebase_ml_test.dart
```diff
@@ -8,102 +8,257 @@ import 'package:firebase_ml/firebase_ml.dart';
...
-              'modelName': MODEL_NAME,
-            },
-          ),
-        ],
-      );
+          () => FirebaseModelDownloadConditions(iosAllowCellularAccess: null),
+          throwsA(isA<AssertionError>()));
+    });
+
+    test('conditions are passed as a map correctly', () {
```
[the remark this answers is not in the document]
PolinaGo: This is testing toMap function. This method does not differentiate between default and non-default values. Assignment of values has already happened, and we know they are not null. Testing how constructor works is done in the previous tests.

packages/firebase_ml/test/firebase_ml_test.dart
```diff
@@ -8,102 +8,257 @@ import 'package:firebase_ml/firebase_ml.dart';
...
+      final model = FirebaseCustomRemoteModel(MODEL_NAME);
+
+      expect(model, isA<FirebaseRemoteModel>());
+      expect(model, isNotNull);
+      expect(model.modelName, MODEL_NAME);
     });
 
-    tearDown(() {
-      FirebaseModelManager.channel.setMockMethodCallHandler(null);
+    test('constructor throws exception when name is null', () {
```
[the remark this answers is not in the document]
PolinaGo: Assert is a general error that indicates that some expression is not satisfied. It explicitly states the expression providing an actionable item for the user. I don't see the need to check how correctly assertion message works as long as we check for this specific type of error.

Alternatively, each assert statement can have a specific message provided by the developer. In that case I should check the messages.

packages/firebase_ml/android/src/main/java/io/flutter/plugins/firebaseml/ModelManager.java
```diff
@@ -95,7 +94,9 @@ public void onComplete(@NonNull Task<File> task) {
                   result.success(modelFile.getAbsolutePath());
                 } else {
                   result.error(
-                      "FirebaseModelManager", task.getException().getLocalizedMessage(), null);
+                      "FirebaseModelManager",
+                      "Please make sure your custom remote model is downloaded.",
```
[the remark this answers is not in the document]
PolinaGo: You are right, good catch!

packages/firebase_ml/test/firebase_ml_test.dart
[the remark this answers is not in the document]
nauthadr: Checking an Assertion is thrown does not guarantee the expected error was encountered especially if you have many possible reasons for that same exception to be thrown. Checking the message would ensure the right Assertion error is thrown with the right feedback for the developer.

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
```diff
@@ -7,24 +7,57 @@ part of 'firebase_ml.dart';
...
+            modelManager.download(model, conditions),
+            throwsA(isA<PlatformException>().having((e) => e.toString(),
+                'message', contains('Failed to schedule the download task'))));
+      }
+
+      if (Platform.isIOS) {
+        expect(
+            modelManager.download(model, conditions),
+            throwsA(isA<PlatformException>().having((e) => e.toString(),
+                'message', contains('The operation couldn’t be completed'))));
```
kroikie, line 47: Is this the error message that the iOS SDK provides? I'm wondering if it could be more informative.

packages/firebase_ml/lib/src/model_management/firebase_model_manager.dart
kroikie: I would feel free to add more context to documentation. Especially in cases where the native documentation may be less than detailed :)

packages/firebase_ml/example/test_driver/firebase_model_manager.dart
PolinaGo: Yes, it is. Unfortunately, unless we wrap it with our own error message (which I have not seen in other plugins), that's what we get.

PR #3044, opened 2020-07-28
[firebse_auth] Remove invalid override on toMap in UserInfo

kroikie:
Remove invalid override annotation in `user_info.dart`. Maybe next time I'll believe the checks before merging :)


kroikie:
This is already fixed.

PR #3094, opened 2020-08-04
[Firebase_auth] Text says register instead of sign in auth example

mosesweb:
Should say sign in here for more clarity


mosesweb:
@googlebot I signed it!

PR #3107, opened 2020-08-05
Use firebase hosting

kroikie:
## Description

Setup a workflow to deploy documentation to Firebase hosting when the `docs` directory is updated.


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,43 @@
+name: publish_docs
+
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - "docs/**"
```
Salakar, line 8: Late to the party but could we also include `website/**` here 😅

Issue #3236, opened 2020-08-19
[firebase_admob] Prepare your apps for iOS 14 issue


YazeedAlKhalaf:
We need to add this to the package `firebase_admob`.

The issue is in Admob's dashboard.

```
Prepare your apps for iOS 14 Apple announced the new AppTrackingTransparency framework, which requires changes to your iOS apps. Implement the GMA SDK 7.64.0 (or later) and set up consent messaging to help prevent a significant loss in ad revenue.
```

<img src="https://i.imgur.com/UN5q6Y2.png" />

Ehesp:
@kroikie one to take up internally 👍

ianpark:
This is urgently required. how do we get this done asap?

kroikie:
I believe this should now be resolved. The `firebase_admob` plugin now depends on required AdMob pod #3347

deniza:
Hey guys. I just published a package to solve this problem. Just use this package, build your apps using Xcode 12 and publish to an ios 14 device for a test drive.

https://pub.dev/packages/app_tracking_transparency

@kroikie I am afraid but updating admob framework does not solve this issue. First we should  build our app using latest Xcode 12 (currently in beta). Then we call ATTrackingManager functions to request tracking permissions. After these steps firebase_admob sdk can actually use tracking identifier to serve personalized ads.

PR #3347, opened 2020-08-25
feat(firebase_admob): integrate Admob with new core

### LINKED ISSUE #3209: [firebase_admob] version solving failed
**Describe the bug**
Today, I saw that many packages got updates. And I updated them. Unfortunately, I got below error on `firebase_admob`:

```
Because firebase_admob 0.9.3+4 depends on firebase_core ^0.4.2+1 and no versions of firebase_admob match >0.9.3+4 <0.10.0, firebase_admob ^0.9.3+4 requires firebase_core ^0.4.2+1.

So, because myapp depends on both firebase_core ^0.5.0 and firebase_admob ^0.9.3+4, version solving failed.
pub get failed (1; So, because myapp depends on both firebase_core ^0.5.0 and firebase_admob ^0.9.3+4, version solving failed.)
```

**To Reproduce**
pubspec.yaml:
```
firebase_core: ^0.5.0
firebase_admob: ^0.9.3+4
```

**Expected behavior**
The package `firebase_admob` should be updated to use new `firebase_core`...

**Flutter doctor**
```
[✓] Flutter (Channel dev, 1.21.0-9.0.pre, on Mac OS X 10.15.3 19D76, locale en-IN)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 11.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 4.0)
[✓] Connected device (2 available)
```

kroikie:
Depend on firebase_core v 0.5.0 and migrate plugin to use firebase_core native SDK versioning features;

Firebase iOS SDK versions are now locked to use the same version defined in firebase_core.
Firebase Android SDK versions are now using the Firebase Bill of Materials (BoM) to specify individual SDK versions. BoM version is also sourced from firebase_core.

Fixes #3209


Ehesp:
Is it worth adding in https://github.com/FirebaseExtended/flutterfire/issues/3236 as part of this too?

kroikie:
Seems urgent enough to add in.

Salakar:
Re: #3236 I don't think there's anything we need to do on the FlutterFire side other than make sure we're using the latest sdk, source: https://github.com/invertase/react-native-firebase/issues/4107

kroikie:
Makes sense, we will have to update this dependency outside of setting Firebase Core since Firebase/AdMob is no longer updated and Firebase docs recommends direct dependency.

Salakar:
@googlebot I consent.

Salakar:
Both Android & iOS should be working now, have ran both locally:

image

Salakar (approved):

john990:
> Makes sense, we will have to update this dependency outside of setting Firebase Core since Firebase/AdMob is no longer updated and Firebase docs recommends direct dependency.

Why Firebase/AdMob is no longer updated?!!!
Developing flutter applications can’t make money. Why do we use flutter?

Salakar:
@john990 Firebase/AdMob podspec is just a shell to bring in the underlying google ads pods, what's being said here is that we'd in future rely on the google ads pods directly instead. Only the shell podspec is what's not being updated

praveentcom:
When is the release planned for this PR?

ianpark:
I think this update is still in prerelease. When is it going to be published?

https://pub.dev/packages/firebase_admob

john990:
> @john990 Firebase/AdMob podspec is just a shell to bring in the underlying google ads pods, what's being said here is that we'd in future rely on the google ads pods directly instead. Only the shell podspec is what's not being updated

Wow, this is really good news, is it implemented entirely in Flutter?  Is there an approximate release date?  look forward to!

Issue #3273, opened 2020-08-20
[firebase_auth] Missing Exceptions


feinstein:
I am migrating my exceptions to the new ones and on `signInWithEmailAndPassword()` I can see some changed. These are the old:

```dart
  ///  * `ERROR_INVALID_EMAIL` - If the [email] address is malformed.
  ///  * `ERROR_WRONG_PASSWORD` - If the [password] is wrong.
  ///  * `ERROR_USER_NOT_FOUND` - If there is no user corresponding to the given [email] address, or if the user has been deleted.
  ///  * `ERROR_USER_DISABLED` - If the user has been disabled (for example, in the Firebase console)
  ///  * `ERROR_TOO_MANY_REQUESTS` - If there was too many attempts to sign in as this user.
  ///  * `ERROR_OPERATION_NOT_ALLOWED` - Indicates that Email & Password accounts are not enabled.
```

And these are the new:

```dart
  /// A [FirebaseAuthException] maybe thrown with the following error code:
  /// - **invalid-email**:
  ///  - Thrown if the email address is not valid.
  /// - **user-disabled**:
  ///  - Thrown if the user corresponding to the given email has been disabled.
  /// - **user-not-found**:
  ///  - Thrown if there is no user corresponding to the given email.
  /// - **wrong-password**:
  ///  - Thrown if the password is invalid for the given email, or the account
  ///    corresponding to the email does not have a password set.
```

So `ERROR_TOO_MANY_REQUESTS` and `ERROR_OPERATION_NOT_ALLOWED` aren't in the docs anymore.

Just to confirm, is this intentional or are they missing?

feinstein:
I can see the same for many other methods, like `reauthenticateWithCredential` and `updatePassword`.

Ehesp:
Hmm, those ones are not method specific and can happen across them all - maybe worth adding these to the FirebaseAuthException class..

KristianBalaj:
#3402 check out this PR. It could solve some exception codes struggles 👋

feinstein:
I am also finding other exceptions mismatches:

* `reauthenticateWithCredential`:
    * Old Exceptions missing:
        * `ERROR_USER_DISABLED`        
        * `ERROR_OPERATION_NOT_ALLOWED`
    
    * New Exceptions not matching old ones:
        * `user-mismatch`
        * `invalid-email`
        * `invalid-verification-code`
        * `invalid-verification-id`

* `updatePassword`:
[...]

PR #2963, opened 2020-07-15
chore: update contributing file

Ehesp:

PR #3402, opened 2020-08-30
[firebase_auth] Auth exception enum codes

KristianBalaj:
## Description

The `FirebaseAuthException` codes should be static typed to prevent any typo bugs in the client code. Therefore there should be an enum that ensures the AuthException type.

At the same time, this static typed code of the exception serves as an implicit documentation of the possible states of an `FirebaseAuthException`.

## Related Issues
[#3273](https://github.com/FirebaseExtended/flutterfire/issues/3273)

## Checklist
- [x] creating `AuthExceptionStatusCode` enum type.
- [x] add parsing of the exception string codes to the `FirebaseAuthException` class via the `statusCode` getter method.

## Breaking Change

### DIFF
--- a/packages/firebase_auth/firebase_auth_platform_interface/lib/src/auth_exception_status_code.dart
+++ b/packages/firebase_auth/firebase_auth_platform_interface/lib/src/auth_exception_status_code.dart
@@ -1,0 +1,26 @@
+/// The codes of the auth exceptions.
+enum AuthExceptionStatusCode {
+  /// Used if the email address is not valid.
+  invalidEmail,
+
+  /// Used if the user corresponding to the given email has been disabled.
+  userDisabled,
+
+  /// Used if there is no user corresponding to the given email.
+  userNotFound,
+
+  /// Used if the password is invalid for the given email, or the account corresponding to the email does not have a password set.
+  wrongPassword,
+
+  /// Used if the Firebase Authentication quota is reached.
+  tooManyRequests,
+
+  /// Used if specific auth provider is not enabled.
+  operationNotAllowed,
+
+  /// Used if the email exists for multiple Firebase user's providers.
+  accountExistsWithDifferentCredential,
+
+  /// Used if the status is undefined.
+  undefined
+}
--- a/packages/firebase_auth/firebase_auth_platform_interface/lib/src/auth_exception_status_code.dart
+++ b/packages/firebase_auth/firebase_auth_platform_interface/lib/src/auth_exception_status_code.dart
@@ -21,6 +21,9 @@ enum AuthExceptionStatusCode {
   /// Used if the email exists for multiple Firebase user's providers.
   accountExistsWithDifferentCredential,
 
+  /// Used if the request failed due to network issues.
+  networkRequestFailed,
+
   /// Used if the status is undefined.
   undefined
 }
--- a/packages/firebase_auth/firebase_auth_platform_interface/lib/src/firebase_auth_exception.dart
+++ b/packages/firebase_auth/firebase_auth_platform_interface/lib/src/firebase_auth_exception.dart
@@ -56,6 +56,8 @@ class FirebaseAuthException extends FirebaseException implements Exception {
         return AuthExceptionStatusCode.operationNotAllowed;
       case 'account-exists-with-different-credential':
         return AuthExceptionStatusCode.accountExistsWithDifferentCredential;
+      case 'network-request-failed':
+        return AuthExceptionStatusCode.networkRequestFailed;
       default:
         return AuthExceptionStatusCode.undefined;
     }
--- a/packages/firebase_auth/firebase_auth_platform_interface/lib/src/firebase_auth_exception.dart
+++ b/packages/firebase_auth/firebase_auth_platform_interface/lib/src/firebase_auth_exception.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:firebase_auth_platform_interface/src/auth_exception_status_code.dart';
 import 'package:firebase_core/firebase_core.dart';
 import 'package:meta/meta.dart';
 import 'auth_credential.dart';
@@ -37,4 +38,26 @@ class FirebaseAuthException extends FirebaseException implements Exception {
 
   /// The tenant ID being used for sign-in/linking.
   final String tenantId;
+
+  /// The exception code parsed into the [AuthExceptionStatusCode] enum.
+  AuthExceptionStatusCode get statusCode {
+    switch (code) {
+      case 'invalid-email':
+        return AuthExceptionStatusCode.invalidEmail;
+      case 'user-disabled':
+        return AuthExceptionStatusCode.userDisabled;
+      case 'user-not-found':
+        return AuthExceptionStatusCode.userNotFound;
+      case 'wrong-password':
+        return AuthExceptionStatusCode.wrongPassword;
+      case 'too-many-requests':
+        return AuthExceptionStatusCode.tooManyRequests;
+      case 'operation-not-allowed':
+        return AuthExceptionStatusCode.operationNotAllowed;
+      case 'account-exists-with-different-credential':
+        return AuthExceptionStatusCode.accountExistsWithDifferentCredential;
+      default:
+        return AuthExceptionStatusCode.undefined;
+    }
+  }
 }


KristianBalaj:
@googlebot I signed it!

helenaford:
Thanks for updating the class, it looks good although not sure if you would need e.code and e.statusCode (maybe you do). Would be good to see an example of how a user would use this class, assuming it would be exposed to the client? I think there maybe some exceptions missing such as `invalid-verification-id`, `no-current-user`, `weak-password`, `email-already-in-use`, `invalid-phone-number`. And, the tests in instance-e2e.dart should be updated  to use checked to use `AuthExceptionStatusCode`.

KristianBalaj:
@helenaford Thanks for the response! 

Sure the tests. I was asking myself, the class isn't covered? I was working with the root as packages/firebase_auth/firebase_auth_platform_interface so when searching the references, nothing has been found in the tests locally.
I was reading the contribution guide from current version of master. After a bit of searching, I found the actual contribution guide #2963. (for now still only in draft).

I will also add the missing codes that I will find in the tests 👍

**The actual usage** of the enum instead of the string is the usage in type checked manner (eg. switch statement) which could be used in the catch block. Also the code could be forwarded to any other method, but still it is type checked and the **client code is aware of the possible codes/states**.

And of course I need to export the `AuthExceptionStatusCode` enum.

**I would propose** to move the `e.code` under the `@deprecated` attribute in favor of `e.statusCode`. In case of localization purposes of the code as a key the usage could be like `e.statusCode.toString()`.

Ehesp:
I'm in two minds about this one. The problem we have is currently there is 3 platforms (and potentially more in the future) all using shared codes and their own platforms ones. 

Right now, we document common error codes returned and "stringify" the platform codes into a common format (e.g. "operation-not-allowed"). Generally if the underlying platform errors you can still read out the `.code` property and expect to handle it.

If we go down the route of adding enums, it means we have to add every single possibility for users to correctly handle them, and there is a LOT of codes. If Firebase update or add a new error code, we need also need to add this enum. If we add a new platform, we have to add every new error code (which isn't always obvious). 

There's also the case of some error codes not being applicable (e.g. `missing-phone-number` on web via `signInWithPhoneNumber`). We assert that there must be a phone number in Dart anyway, so this will never be returned.

TLDR; Right now we just let errors fall through to users in a formatted string, this change (or similar) would require us to know about every single error code and where they can pop up (which isn't easy to find and is error prone).

feinstein:
@Ehesp I completely understand your reasoning, and had some suspicions things were the way they are because of the underlining platforms/firebase libs, so I don't disagree with your.

I just wish things were different and properly documented, as in our App we like to give users meaningful and personal erros and alerts. We believe this improves user experience and reduces frustration when an error happens. 

Having just a catch-all "Ops, something went wrong, please contact customer service" isn't our goal, but unfortunately I see there's no escape from this, unless the underlying libraries provide better documentation.

leonardocustodio:
@Ehesp, @feinstein  Well I totally disagree.... Since this package is officially from Google I would expect that you guys at least had some kind of communication. Having error codes as plain string is horrible. Firebase changed the errors to lower case and my apps stopped catching the errors properly so it is not really acceptable when using an official package that your app will break that easy.

I would suggest at least if the error codes change in the server side and you guys don't know about it that returned some kind of "unknown" enum temporarily until changing all error codes to the right ones and updating in the package.

And not expect developers to keep looking if the error code haven't changed.

leonardocustodio:
@Ehesp, @feinstein Well I totally disagree.... Having error codes as plain string is horrible for people using the package. Firebase changed the errors to lower case and my apps stopped catching the errors properly and probably everybody else as well. It is way easier for people developing the package to know when error codes has changed because they are working in the package and more up to date with changes on firebase than developers that are just using the package.

I would suggest at least if the error codes change in the server side and you guys don't know about it that returned some kind of "unknown" enum temporarily until changing all error codes to the right ones and updating in the package.

I mean should the developers really keep checking from now and than if the error codes haven't changed and the app stopped working and than update all apps with new strings? Or just update a package when they change?

Ehesp:
@leonardocustodio the PR is still open and being discussed so you're welcome to an opinion. 

I personally think we should have both a string version and enum value, however if the string value doesn't match a handled enum, it returns `unknown`. I think removing the string value like this PR will mask any unhandled errors rather than leave them open to implementation.

Leaving this one open still, I'll get round to it at some point if no one carried it on.

leonardocustodio:
Maybe could we make an extra option? The enum for general case and leave the string for more specific ones? I don't know. Not perfect but is one way to go?

Ehesp:
I'm open to having:

`FirebaseAuthException.code` = 'permission-denied' 
+
`FirebaseAuthException.status` = `AuthExceptionStatusCode.permissionDenied`

leonardocustodio:
That would be nice. Or another option would be to have a helper that we could pass the value and it would return the enum.
So in the case the enum returned from the helper is unknown the person would still have the string in the first place.

Just a code snippet from RevenueCat library on how they threat the errors as example:

```
  static PurchasesErrorCode getErrorCode(PlatformException e) {
    var errorCode = int.parse(e.code);
    if (errorCode >= PurchasesErrorCode.values.length) {
      return PurchasesErrorCode.unknownError;
    }
    return PurchasesErrorCode.values[errorCode];
  }
```

But I think it is easier to have both values (code + status) like you said.

feinstein:
@leonardocustodio AFAIK the status codes changed after a new version, with breaking changes documented, was announced.

I always check the Changelog for breaking changes before updating any library references, so I wasn't caught by surprise that my exceptions stopped working. Still it was very frustrating as well and that's why I opened the issue linked with this PR.

I suggest you orient your team to only update libraries after checking the Changelogs, this helps a lot in keeping surprises to the minimum.

I also like the ideas of enums, but I also agree there has to be some "catch all" code since the folks at Firebase don't have this properly documented and set in stone, so these suggestions seem positive IMO.

ollydixon:
@feinstein not every developer checks the change logs. Infact I don't know any that do because usually we relay on 'deprecated flags'. This bug has actually cause a ton of silient errors across many Flutter apps using Firebase now; not good.

feinstein:
@ollydixon you can't put deprecated flags on exception strings or other values that just change the way they are returned and not their name. You can't put deprecated annotations on old behaviors, just on old names.

I know most people don't check the changelogs, but that's plain wrong. How else would they measure the impact of a change? You need to check for every major version change at least, by definition they might include breaking changes.

I am not arguing we don't need exception codes, just that "people don't check the changelogs" is an argument as strong as "people don't do unit tests". We can't guide our decisions based on bad practices that others are adopting.

ollydixon:
There are soltions for deprecation on this; I can think of a few that could have been implemented; perhaps they will be moving forwards with enums etc.
@feinstein try managing 31 projects. We don't have time to check the changelogs on 1000s of updates and tbh I'm sort of surprised you think anyone reads them! 🙃 on most of the packages though pub.dev there aren't even changelogs. :-)

feinstein:
All the packages that I use have changelogs.

If you have to manage 31 projects, than you are an exception on the field. Most teams won't have more than a handful of projects. I suggest you have a quality control were versions are vetted once and then propagated to your other projects. So your projects only use versions pre approved.

I do think changelogs are amazing, I can adapt my code to the latest additions, I learn a lot from what the package has to offer. I am surprised actually that people don't look at them. If they are so useless, why bother wasting time with them? Better remove them altogether.

Anyway, this is getting off topic and polluting the thread.

ollydixon:
@KristianBalaj, we implemented the new error codes across our projects. Will these be changed back, or are we safekeeping them? This is important, so customers do have a lousy experience again. We got a few 1-star reviews across our apps because of this issue.

KristianBalaj:
Hey, just checking and I need to consent to the cla not only sign it.

Also there are still some formatting issues needed to be fixed and I can't right now (check the checks).

KristianBalaj:
@googlebot I consent.

KristianBalaj:
Okay so obviously not my consent to googlebot :) @Ehesp

ollydixon:
@KristianBalaj does that mean the strings will be the same or there are enums?

KristianBalaj:
This PR is about adding the enums that will be unchangable - static.

And will be covered with tests, so any upcomming changes should be covered.

As far as this will be merged by the maintainers. I am not the maintainer.

ollydixon:
@KristianBalaj perfect! I think enums is the best solution.

Ehesp:
Yeah we're just trying to get messaging out of the door since we need to do some global version bumps. Once thats landed we'll get some PRs merged and released!

kroikie:
I would be in favor of having enums while still providing the underlying code from the platform as a string. Given the number of platforms there is no way to have enums covering all the necessary exceptions. A catch-all `unknown` with the underlying code as a string seems to me to be a good compromise.

russellwheatley:
Closing out as this isn't something we plan to implement ([See reasoning](https://github.com/firebase/flutterfire/pull/3402#issuecomment-685402151)) and has been stale for quite some time.

PR #3579, opened 2020-09-18
Initial test case for investigating web transaction runtime error

dackers86:
Initial test case for investigating transaction runtime errors with type mismatch


dackers86:
@Ehesp Is this worth closing. Not too sure if a working example would be useful given the issues encountered.

Happy to update once we have a solution?

Ehesp:
Yeah lets close, need to dig further.

PR #3851, opened 2020-10-12
fix(firebase_remote_config): Update plugin to no longer use deprecated APIs

karenyz:
Keeping public source in sync with internal copy


mehmetf (approved):

Salakar:
@karenyz thanks for this PR, apologies that it got stale - this was out of our control. Remote Config has since been re-written to use the latest APIs and I've confirmed I can't see any instances of these old APIs in the implementation anymore - so this PR is not longer required (sorry 🙈 )

Issue #4013, opened 2020-11-02
🐛 [firebase_storage 5.0.0-dev.4] Update SDK to 26


dvird:
In order to present percent progress, TaskState.running should send events after every few bytes.
Today its sending only 2 3 4 events, which is not enough for presenting well progress on our screen.

old SDK has StorageTaskEventType.progress which worked great, but since the refactor youve made its not working anymore

Ehesp:
This is the native SDK, we just relay the events on as they come through so it's nothing we've changed. The SDK versions have updated though.

TahaTesser:
Hi @dvird 
As mentioned above this is from native SDK, with no changes from the plugin itself
Thank you

dvird:
@TahaTesser @Ehesp firebase storage is now 26, but flutter use 25. i read somewhere they fixed it already

russellwheatley:
As mentioned, this isn't something we have control over. More over, we're now using `26.2.0` android BOM version which ought to fix the android sdk issue.

PR #4249, opened 2020-11-27
feat!: forward port to firebase-ios-sdk v7+ / firebase-android-sdk v26+

mikehardy:
## Description

This is a forward port of FlutterFire to firebase-ios-sdk v7+ and firebase-android-sdk v26+

Each commit may be viewed in isolation, and in fact that is recommended.

## Related Issues

Unsure if there are other issues to relate but:
#4013 - updating to SDK26 for firestore
#3969 - auth useEmulator discussion - it is gated on this but is trivial once the forward port is done

## Checklist

This is my first Dart code, first Flutter code, and first batch of FlutterFire PRs
I did [the same port for react-native-firebase](https://github.com/invertase/react-native-firebase/pull/4471) though, so I'm new here but have been in the area a lot. Be gentle ;-)

**I am not sure what should be done in the CHANGELOG or with regard to version numbers here.**

- [x] I signed the [CLA].

## Breaking Change

iOS minimum deployment target is moved to 10+
RemoteConfig debugMode no longer exists. minimumFetchTime / fetchTimeout replace it

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).


mikehardy:
@googlebot I signed it!

mikehardy:
@Salakar  / @Ehesp this is ready for review, I believe iOS and Android are both working now, should clear the way for auth emulator as a quick followon

I understand there may be concurrent work with regards remote-config port, happy to coordinate on that though I was very very light with regards to my forward-port treatment of it here and kept the commits pretty granular so it should be easy to deal de-conflict / integrate

mikehardy:
GitHub actions aren't wrapped in retry so the android gradle ones flake without recovering as they attempt to pull remote dependencies. Something like this would fix it though it will need to be flutter-ified https://github.com/invertase/react-native-firebase/pull/4084/commits/f1b9af98f26f6b1e7302035785fa2c3108497922


One of the storage tests flaked for ios, it's passed before during this PR development and passes locally so should be unrelated to the commit:

```

[firebase_storage_example]:   updateMetadata
[firebase_storage_example]:      ✔ updates metadata
[firebase_storage_example]:      ✘ errors if metadata update removes existing data
[firebase_storage_example]:        [E]: Expected: {'action': 'updateMetadata test'}
[firebase_storage_example]:               Actual: {'activity': 'test', 'action': 'updateMetadata test'}
```

If you retry both (I don't have permission) it should go green after one or two attempts, they work locally each time

packages/firebase_remote_config/example/test_driver/firebase_remote_config_e2e.dart
```diff
@@ -25,14 +26,18 @@ void main() {
       expect(lastFetchTime.isBefore(DateTime.now()), true);
       await remoteConfig.fetch(expiration: const Duration(seconds: 0));
       expect(remoteConfig.lastFetchStatus, LastFetchStatus.success);
-      await remoteConfig.activateFetched();
+      final activated = await remoteConfig.activateFetched();
+      expect(activated, true);
 
+      // TODO should verify that our config settings actually took
```
mikehardy, line 32: Just a note, there were problems here when I was testing and I could not figure them out. Apparently there was a remote-config bug in 7.1.0 (which I was testing) where remoteconfig was completely broken https://github.com/firebase/firebase-android-sdk/issues/2186 and remote settings did not come in. Bears a retest

packages/firebase_remote_config/ios/Classes/FirebaseRemoteConfigPlugin.m
```diff
@@ -97,12 +104,28 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
+                [error.userInfo[@"ActivationFailureReason"] containsString:@"already activated"]) {
+              newConfig = @false;
+            } else {
+              FlutterError *flutterError;
+              flutterError = [FlutterError errorWithCode:@"activateFailed"
+                                                 message:error.userInfo[@"ActivationFailureReason"]
+                                                 details:nil];
+              result(flutterError);
+            }
+          }
```
russellwheatley, line 121: I think there ought to be an `else` block wrapped around the rest of the `activateWithCompletion ` code block (i.e line 123-127).

packages/firebase_remote_config/ios/Classes/FirebaseRemoteConfigPlugin.m
```diff
@@ -97,12 +104,28 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
-    resultDict[@"newConfig"] = [NSNumber numberWithBool:newConfig];
-    resultDict[@"parameters"] = parameters;
-    result(resultDict);
+    [[FIRRemoteConfig remoteConfig]
+        activateWithCompletion:^(BOOL changed, NSError *_Nullable error) {
+          BOOL newConfig = @true;
+          if (error) {
+            if (error.userInfo && error.userInfo[@"ActivationFailureReason"] != nil &&
+                [error.userInfo[@"ActivationFailureReason"] containsString:@"already activated"]) {
+              newConfig = @false;
```
russellwheatley, line 113: should this `newConfig ` assignment be in the outer `if` block? assuming both errors cause config activation to fail?

packages/firebase_remote_config/example/test_driver/firebase_remote_config_e2e.dart
russellwheatley: I got a successful run locally 👍

russellwheatley (changes requested):
Ahoy @mikehardy, looking good, just a couple of minor points that I think might be an issue.

packages/firebase_remote_config/ios/Classes/FirebaseRemoteConfigPlugin.m
mikehardy: Interesting - will spiff this area up - you are right that is safer, however the else block calls `result(...)` which means (I think?) that the newConfig value will never be used as the statements where it would be assigned into a results dict and returned won't be hit. So I don't think there was an error currently but it's not great if that's not immediately clear on a read

Structurally for max safety, seems like I should start with it set to `@false` really, and then after the error handling block set it to `@true` with a little `// no errors, we have a new config` side comment

packages/firebase_remote_config/ios/Classes/FirebaseRemoteConfigPlugin.m
mikehardy: I think that's not exactly correct - there are two error cases, one which returns a regular result but noting it's "already activated" and one which is a real error type that falls into calling code error handling blocks

So the if (error) block is differentiating between those two, then the following block executes no matter what (assuming an error wasn't tossed out from the if(error && reallyWasAnError) case

Similar to the above note though, if that's not immediately clear, it's a defect. I'll think if there is a way to shape it up differently or at minimum add a quick comment explaining

russellwheatley: ah, I see. The reason why I raise that is the callback will be called twice if the error is not "ActivationFailureReason" if I'm not mistaken? This may result in another error/possible duplicate response in the dart code.  A `return;` after `result(flutterError);` I think ought to do the trick I think 🤔

mikehardy: ...and *that* is a real problem. Would have missed that even as I'm polishing it. I'll see if I can throw some test cover in there as well on error conditions to expose it since it's obvious in hindsight but was in my blind spot. Thanks

packages/firebase_remote_config/ios/Classes/FirebaseRemoteConfigPlugin.m
russellwheatley: I'm having a look at the RNFB implementation, and I see it mirrors that. To my mind, it doesn't seem right to return two different types (i.e an error and a successful resolution albeit with activated as `false`) for an error. But I'm more than likely missing something so I think this can be safely ignored 😄

mikehardy: Yeah - you will likely not be surprised to learn that I lifted this from the RNFB implementation :-), I agree it seems a little odd however when I took a little stab at doing it differently it meant changing up how the Dart code above this native layer worked since it already had the "newConfig" native toggle style and used the results dictionary to cache the values for later retrieval. It wasn't a huge change but meant the PR was less surgical then I wanted given the real point was just to ingest SDK API changes, not clean up how it worked. So I left it

packages/firebase_remote_config/ios/Classes/FirebaseRemoteConfigPlugin.m
mikehardy: I added explanatory comments, and fixed the "missing return in the error case" issue you found.
I explored adding cover but there were enough inconsistencies in how lastFetchTime and activation status were coming back that I feel it's best left as a minimal port per @Salakar direction, pending landing of #4186 which will otherwise be a massive collision with this

russellwheatley (approved):
LGTM

Salakar (approved):
LGTM

PR #4266, opened 2020-11-30
fix: issue when specifying a functions region #4226

### LINKED ISSUE #4226: 🐛 [cloud_functions_web] CloudFunctions calls the "us-central1" server even if the regions is set to "europe-west1"
## Bug report

When calling a "europe-west1" function, the url is looks the function in the "us-central1" server, and eventually crashes, even if setting the region to europe.

### Steps to reproduce

call a function and set the region, like this
`HttpsCallable callable = FirebaseFunctions.instanceFor(region: 'europe-west1').httpsCallable('function-name');`

### Expected behavior

The function must be called to the europe-west1 server.

## Additional context

Rolling back to 3.0.1 fixes the issue

wliumelb:
fix #4226


wliumelb:
@googlebot I signed it!

Salakar (approved):
Oh wow, no idea how the e2e region tests are even passing then 😞 I will take a look at the tests in another PR. Thanks for sorting this

PR #4277, opened 2020-12-01
fix: child() missing after ref()

subirats345:
Hi! Thanks for the great work. I was working with the `getDownloadUrl()` and I noticed that in the example above the ref() function have arguments and the child() function is missing. I tried the code and ref() expects 0 arguments. With  `ref().child(“string”).getDownloadUrl()` works perfectly. Thank you!


subirats345:
@googlebot I signed it!

Ehesp:
This shouldn't be the case: https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_storage/firebase_storage/lib/src/firebase_storage.dart#L111

Are you sure you're using the latest version?

PR #4383, opened 2020-12-11
feat(auth_web): constant initialization for auth on app(s) start up

russellwheatley:
## Description

Work in progress.

## Related Issues

To be confirmed.

## Checklist

- [ ] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [ ] I signed the [CLA].

## Breaking Change


packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -28,6 +29,29 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
...
+  /// Called during initializeApp() to instantiate plugin constants before the user's code
+  /// executes. We listen fr initial auth state event so we're sure web app
+  /// is initialized
+  static Future<Map<String, dynamic>> constantInitializor(app) async {
+    auth_interop.Auth authInstance =
+        auth_interop.getAuthInstance(core_interop.app(app.name));
+    StreamSubscription<auth_interop.User> stream;
+    Completer completer = Completer<Map<String, dynamic>>();
+
+    stream = authInstance.onAuthStateChanged.listen((event) {
```
Ehesp, line 46: It might be cleaner here to await a `first` event instead: https://api.dart.dev/stable/1.10.1/dart-async/Stream/first.html

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -28,6 +29,29 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
...
+      auth_interop.User user = authInstance.currentUser;
```
Ehesp, line 47: Does it not make more sense her to use the event provided by the stream? (user or null)

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
```diff
@@ -112,6 +143,11 @@ class FirebaseAuthWeb extends FirebaseAuthPlatform {
     return UserWeb(this, _webAuth.currentUser);
   }
 
+  @override
+  set currentUser(UserPlatform userPlatform) {
```
Ehesp, line 147: Since we've now set `this.currentUser`, the `get currentUser` method should probably use this value, I think. That also means, the event which returns the current user (`onAuthStateChanged`) should also be setting it? 

Right now, the current user is being taken from the `_webAuth` instance, rather than our own `currentUser` property.

packages/firebase_core/firebase_core_platform_interface/lib/src/platform_interface/platform_interface_firebase_plugin.dart
```diff
@@ -21,6 +21,14 @@ abstract class FirebasePluginPlatform extends PlatformInterface {
   /// requires that the value is synchronously available for use after initialization.
   static Map<dynamic, dynamic> _constantsForPluginApps = {};
 
+  static void setConstantsForPluginApps(String appName,
```
Ehesp, line 24: For consistency, the method channels should probably use this method instead of directly using the private static?

packages/firebase_core/firebase_core_web/lib/src/firebase_core_web.dart
```diff
@@ -81,6 +111,7 @@ class FirebaseCoreWeb extends FirebasePlatform {
         throw _catchJSError(e);
       }
     }
+    await _initializeCore();
```
Ehesp, line 114: Add a space and comment here just to explain what it's doing

Ehesp (changes requested):

packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart
russellwheatley: Updated the currentUser to get our private `_currentUser`. To your second point, if I understood you correctly;   `onAuthStateChanged` is called in a static method on initialisation so we can't set any instance variables. I don't think it should matter because on first use of the delegate, we `setInitialValues` which will set the user and languageCode anyway.

Ehesp: Right now though, the `_currentUser` is set by `setInitialValues`, which is fine. However once the web instance is initialised it'll also trigger an [onAuthStateChanged](https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart#L68) to pass in those events to the change listeners on the users end.

If the user logs out, the users will get the event through the Stream, but the `currentUser` won't be updated since it now comes from the `_currentUser` property, which hasn't been updated.

Therefore the `_currentUser` needs updating here: https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_auth/firebase_auth_web/lib/firebase_auth_web.dart#L75

packages/firebase_auth/firebase_auth_platform_interface/lib/firebase_auth_platform_interface.dart
```diff
@@ -31,4 +31,4 @@ export 'src/providers/phone_auth.dart';
 export 'src/providers/twitter_auth.dart';
 
 ///method channel name used by both delegates
```
Ehesp, line 33: Space / Capitalization - remember it's used for public API doc generation

lastthyme (approved):

Ehesp:
Going to close this one for the time being. 

We got around the issue via this PR: https://github.com/FirebaseExtended/flutterfire/pull/5766 - that being said, it only solves that specific issue, so this PR will still come in useful as reference when we need to deal with plugins such as remote config.

PR #4407, opened 2020-12-16
fix(firebase_crashlytics): Include obfuscated stack traces

### LINKED ISSUE #1150: [firebase_crashlytics] Support for deobfuscation of Dart reports
I can't even find any info on how to find dart obfuscation maps to deobfuscate reports manually

### LINKED ISSUE #2644: [firebase_crashlytics] No stack traces for issues on Crashlytics dashboard
After upgrading to Flutter SDK 1.17, where dart obfuscation is supported by default, **no stack traces appear for crashes that originate from dart code on the Crashlytics dashboard on the Firebase console**.
On earlier versions of the Flutter SDK they were reported with their obfuscated names.

untp:
## Description

No stack traces appear for crashes that from obfuscated dart code. Because current implementation can't parse obfuscated stack traces. This PR implementation sends unparsed stack traces if the stack trace is obfuscated.

Deobfuscating can be done like [this](https://github.com/FirebaseExtended/flutterfire/issues/2644#issuecomment-746217634). 

### Problems

1. This implementation removes warning line, PID and instructions lines, abs addresses. Because these variables are different between sessions, so the same error will be grouped in Crashlytics Console. (These variables are not important, only `_kDartIsolateSnapshotInstructions+0x` lines are needed for deobfuscating)

2. Crashlytics Console appends `(.java)` and prepends `.` every line. Developers need to trim these lines manually.

### Screenshots from Crashlytics Console

solved-problem
last-state

<details><summary>If warning, pid, instructions lines not removed: (all errors are grouped)</summary>
<p>

</p>
</details>

<details><summary>If warning, pid, instructions lines are removed, but abs is not removed: (These issues are same error with different sessions. Their virt addresses are same, but abs addresses are different. Abs addresses are must be removed for grouping same errors.)</summary>
<p>

problem

no-warning-obfuscated-stack-trace

</p>
</details>

## Related Issues

Fixes #1150
Fixes #2644

## Checklist

- [x] I signed the [CLA].

## Breaking Change


untp:
Currently this PR solves the issue and deobfuscating can be done like [this](https://github.com/FirebaseExtended/flutterfire/issues/2644#issuecomment-746217634). But it shows errors in firebase console like:


It appends (.java) and prepends . every line.
Also issues grouped like this:



We need to remove below lines from stack trace to ungroup issues. And these lines are not needed for deobfuscating the stack trace. We should send these as log, not as stack trace.
```
Warning: This VM has been configured to produce stack traces that violate the Dart standard.
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 28984, tid: 29024, name 1.ui
isolate_dso_base: 75f179900000, vm_dso_base: 75f179900000
isolate_instructions: 75f17990e000, vm_instructions: 75f179902000
```

I don't know how to prevent crashlytics to appending (.java).

Also this PR implementation checks obfuscated stack traces with that warning. But that warning can be removed future versions of dart-sdk. See dart-lang/sdk#43388. Checking with regex is better. We can use regex strings from [native_stack_traces/convert.dart](https://github.com/dart-lang/sdk/blob/master/pkg/native_stack_traces/lib/src/convert.dart) or adding [native_stack_traces](https://pub.dev/packages/native_stack_traces) as dependency.

untp:
I added another commit. Currently stack traces are parsed with `Trace.format(stack)` then calls `getStackTraceElements`, `getStackTraceElements` reparses formatted stack trace. We don't need to reparse. So I changed `Trace.format` to `Trace.from` and using `Trace` methods. I updated tests.

Instead checking obfuscated code with warning, this commit checks with regex. This removes Warning line, pid, instructions lines. So issues are now ungrouped.

I looked to how dart uses obfuscated stack traces. [This](https://github.com/dart-lang/sdk/blob/543364df65673d4c42d0dbbcea77d91415d9be08/pkg/native_stack_traces/lib/src/convert.dart#L111) file parses obfuscated stack traces. We can deobfuscate if we have only `_kDartIsolateSnapshotInstructions+0x`. Or we need both header (`isolate_instructions`, `vm_instructions` line) and `abs` address. We output `_kDartIsolateSnapshotInstructions` so we don't need to output header lines.

Now firebase console output is like this:
no-warning-obfuscated-stack-trace
problem

Issues are ungrouped now but these issues are same error with different sessions. Their virt addresses are same but abs addresses are different. We need to remove abs addresses to group same errors. We only need `_kDartIsolateSnapshotInstructions+0x` to deobfuscate, so we can remove them safely.

After solving this grouping issue, we need to prevent crashlytics to appending (.java) and prepending . every line.

untp:
I solved the grouping issue. Now same errors are grouped.
solved-problem

Now the only problem left is preventing crashlytics from appending (.java) and prepending . every line. It is not a requirement, developers can trim these lines manually.
last-state

Ehesp:
Wow nice one for this PR! Will get it reviewed.

talent-apps:
@Ehesp Please prioritize this issue.
It is crucial for Flutter production apps.
🙏

talent-apps:
@untp 
We tested the PR and we successfully managed to deobfuscate crashes from the Crashlytics reports.
One question remains: each report on the dashboard contains a reason field which is also obfuscated.
Example: Error thrown Instance of 'Ax'.
Is there a way to deobfuscate it? It could be very helpful for the analysis.

talent-apps:
@untp On iOS the stack trace lines have the following format:
`    #00 abs 000000011615a293 _kDartIsolateSnapshotInstructions+0x150293`
The current regexp doesn't pick those lines. Maybe we need a different regexp for iOS?

untp:
@talent-apps awesome, thanks for testing. I forgot to test on iOS. I fixed it with the above commit. If you catch any bugs, I would appreciate it if you let me know.

talent-apps:
@untp When testing the new RegExp, the app force quits upon dart non-fatal crash and we don't see any reports on the dashboard.
Any idea why?

untp:
@talent-apps thanks, I fixed it. The problem was file field of stack trace lines is not nullable on iOS.

talent-apps:
@untp We can confirm that the deobfuscation works on both platforms now. 🥳

dayron9110:
Any updates?

Vitor-Bukovitz:
@Ehesp  & @Salakar  can you guys take a look at this PR? This may solve the issue https://github.com/FirebaseExtended/flutterfire/issues/80

envercc:
This one would be really helpful, please take a look into it!

untp:
@rrousselGit @Salakar @ehesp
Sorry for spamming, but this PR is 3 months old, and it fixes a very important issue (#2644) which has `impact: crowd` tag. This issue is in the 1st page if issues are sorted to most reactions 👍. This issue is fixed with this PR, and it only needs to be reviewed.

Can you review this PR please?

packages/firebase_crashlytics/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -223,9 +221,10 @@ void main() {
 
       test('without character index', () async {
         final List<String> lines = <String>[
-          'package:flutter/src/widgets/framework.dart 3825  StatefulElement.build'
+          '#0      StatefulElement.build (package:flutter/src/widgets/framework.dart:3825:27)'
```
rrousselGit, line 224: I have some doubts about this change.
My understanding is that this test is supposed to check the case without the `:27`. Otherwise it is identical to `with character index`

Although to begin with, I'm not sure when this case happens. Any idea? I'm fine with just removing the test.

packages/firebase_crashlytics/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -237,16 +236,56 @@ void main() {
...
       });
+
+      test('with android obfuscated stack trace', () async {
+        final List<String> lines = <String>[
+          'Warning: This VM has been configured to produce stack traces that violate the Dart standard.',
+          '*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***',
+          'pid: 1357, tid: 1415, name 1.ui',
+          'isolate_dso_base: 75f178181000, vm_dso_base: 75f178181000',
+          'isolate_instructions: 75f17818f000, vm_instructions: 75f178183000',
+          '    #00 abs 000075f17833027b virt 00000000001af27b _kDartIsolateSnapshotInstructions+0x1a127b',
```
rrousselGit, line 258: Would it be possible to make this an integration test instead?
I'm skeptical about the format staying the same, which could break the regexp.

If it's too much work, just leave a todo with me assigned.

rrousselGit (approved):
LGTM

But I wonder why those `(.java)` are added. Thought @russellwheatley?

packages/firebase_crashlytics/firebase_crashlytics/test/firebase_crashlytics_test.dart
untp: Thanks for the review.

Column may not be included in stack trace in some cases. Because [there are if statements in stack trace body print function](https://github.com/dart-lang/sdk/blob/master/runtime/vm/object.cc#L24881).

The old implementation uses `Trace` API to parse stack trace but it converts to string then reparses it. This implementation uses `Trace` API directly.

Since all parsing is done with `Trace` API, parsing is no longer related to crashlytics package.

I will remove this test and replace `with character index` to `with symbolic stack trace`.

russellwheatley (approved):
@rrousselGit I don't know why the Crashyltics backend appends `.java` to each line, my best guess would be to indicate it came from an android app.

PR LGTM once Remi's test issues have been addressed.

packages/firebase_crashlytics/firebase_crashlytics/test/firebase_crashlytics_test.dart
untp: Yes, format can change. But tests are for checking a PR breaks the API design. If someone creates a PR and the tests fail, it should be the PR's problem. The PR must be handle that error or change the test.

If obfuscated stack trace format test included in this package, tests will fail in the future. PRs that unrelated to trace format will fail. They will probably comment out this test and add a todo about handling the new obfuscated format.

Instead we can watch `native_stack_traces` package or [`StackTrace::ToCString` function](https://github.com/dart-lang/sdk/blob/master/runtime/vm/object.cc#L24930). If there is a breaking change, a GitHub issue should be opened. This is better than adding the test.

Also, we need to change the CI workflow. We need to build 2 variants: not obfuscated and obfuscated. And test on both android and ios. Yes, this is too much work.

So what do you think?

untp:
The `file` field is an empty string, so it appears as `.java`. I tried with null, but it didn't work. This issue can only be fixed on the backend. Crashlytics backend should not append `.java` if `file` field is empty. Also on iOS it prepends `0  ???                            0x0`.

untp:
I marked this as draft. I tested now, it outputs the warning line. This shouldn't happen.
er

untp:
Now it works both obfuscated and non-obfuscated.
Working

russellwheatley (approved):
LGTM. What do you think, @rrousselGit?

PR #4541, opened 2021-01-01
docs(firebase_auth): Facebook Auth Example Code Fix

aadarshadhakalg:
FacebookAuth instance's login method is returning AccessToken but in the example code result was of type LoginResult which was causing error. Just changes it to AccessToken and made changes accordingly while creating credential also.


rrousselGit:
Thanks!

PR #4824, opened 2021-01-27
feat!: firebase_messaging nndb migration

Ehesp:
WIP

- [x] fix this test:
image
- [ ] Migrate example app


packages/firebase_messaging/firebase_messaging/lib/src/messaging.dart
```diff
@@ -126,15 +117,15 @@ class FirebaseMessaging extends FirebasePluginPlatform {
...
     return _delegate.getInitialMessage();
   }
 
   /// Removes access to an FCM token previously authorized with optional [senderId].
   ///
   /// Messages sent by the server to this token will fail.
-  Future<void> deleteToken() {
-    return _delegate.deleteToken();
+  Future<void> deleteToken({String? senderId}) {
+    return _delegate.deleteToken(senderId: senderId);
```
Ehesp, line 128: Note; this was actually missing anyway so added it in.

PR #4836, opened 2021-01-28
fix: Analysis error with firebase_core/web

jiahaog:
## Description

Fixes the following analysis warning:

```
firebase_core/web/lib/src/interop/utils/utils.dart:121
  The return type 'void' isn't assignable to 'FutureOr<Null>', as required by 'Future.catchError'. #return_type_invalid_for_catch_error
```

`return_type_invalid_for_catch_error` analysis option should probably be enabled to avoid this in the future

## Breaking Change

### DIFF
--- a/packages/firebase_core/firebase_core_web/lib/src/interop/utils/utils.dart
+++ b/packages/firebase_core/firebase_core_web/lib/src/interop/utils/utils.dart
@@ -113,11 +113,11 @@ PromiseJsImpl<S> handleFutureWithMapper<T, S>(
 ) {
   return PromiseJsImpl<S>(allowInterop((
     void Function(S) resolve,
-    Null Function(Object) reject,
+    void Function(Object) reject,
   ) {
     future.then((value) {
       var mappedValue = mapper(value);
       resolve(mappedValue);
-    }).catchError(reject);
+    }).catchError((error) => reject(error));
   }));
 }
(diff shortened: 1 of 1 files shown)


packages/firebase_core/firebase_core_web/lib/src/interop/utils/utils.dart
```diff
@@ -113,7 +113,7 @@ PromiseJsImpl<S> handleFutureWithMapper<T, S>(
 ) {
   return PromiseJsImpl<S>(allowInterop((
     void Function(S) resolve,
-    void Function(Object) reject,
+    Null Function(Object) reject,
```
Salakar, line 116:
```suggestion
    FutureOr<void> Function(Object) reject,
```

Shouldn't this be `FutureOr<void>`? Note `void` used instead of `Null` for [`prefer_void_to_null`](https://dart-lang.github.io/linter/lints/prefer_void_to_null.html)

Salakar:
Is `return_type_invalid_for_catch_error` the right option name, I can't find any information on this anywhere?

jiahaog:
Actually yeah, I couldn't find any documentation on `return_type_invalid_for_catch_error`, it seems like a thing that is inbuilt into the analyzer. Not sure if it has been set up correctly here. This is where the check is declared:
https://github.com/dart-lang/sdk/blob/019ce33409e246973a217a7f9655444bc13d2619/pkg/analyzer/lib/src/error/catch_error_verifier.dart#L71-L81

packages/firebase_core/firebase_core_web/lib/src/interop/utils/utils.dart
jiahaog: I thought about it a bit more, updated it to use a lambda instead so the type annotation can stay the same :)

jiahaog:
@cyanglaz can you please take a look at this?

cyanglaz (approved):
LGTM

Salakar (approved):

PR #4914, opened 2021-02-02
PluginRegistry cannot be converted to FlutterEngine

pratikbutani:
## Description

PluginRegistry cannot be converted to FlutterEngine.


russellwheatley:
I don't think this is correct and it is out of date.

PR #5241, opened 2021-03-05
feat:Android Intent broadcast to other plugins

tkeithblack:
## Description

Firebase_messaging consumes incoming push notifications and sends those messages to Flutter apps via callbacks. However, sometimes other packages depend on processing push notifications. For example VOIP packages. If a VOIP package cannot trap the push notification it cannot process an incoming call. 

Under iOS it is possible to trap VOIP messages, but under Android only the firebase_messaging package can see the push notifications.

Therefore, currently the only way a VOIP package can get these messages, particularly when the app is in the background, is if the app does not use firebase_messaging.

This Pull Request adds a new Dart entry point to firebase_messaging (Android only): 

`FirebaseMessaging.instance.registerAndroidMessageIntentListener("com.flutter.android.pluginname.message_interest");
`
After registering, each time firebase_messaging receives a push notification it will enumerate listeners and re-broadcast the push notification to the specified Intent(s). This allows any package to get push notifications even when the app is in the background.

Intents should be defined in App AndroidManifest.xml.

 ```
       <service
            android:name="com.developer.myplugin.fcm.NameOfServiceThatProcessesMyIntents"
            android:stopWithTask="false">
            <intent-filter>
                <action android:name="com.flutter.android.pluginname.message_interest" />
            </intent-filter>
        </service>

```

I offer this solution for consideration. Frankly I am not married to my solution, but package developers need a way to get direct access to incoming push notifications so they may co-exist with firebase_messaging.

## Checklist

- [X ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X ] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

None.

- [ X] No, this is *not* a breaking change.

### DIFF
--- a/packages/firebase_messaging/firebase_messaging/android/src/main/java/io/flutter/plugins/firebase/messaging/FlutterFirebaseMessagingBroadcaster.java
+++ b/packages/firebase_messaging/firebase_messaging/android/src/main/java/io/flutter/plugins/firebase/messaging/FlutterFirebaseMessagingBroadcaster.java
@@ -1,0 +1,78 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package io.flutter.plugins.firebase.messaging;
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.util.Log;
+import androidx.localbroadcastmanager.content.LocalBroadcastManager;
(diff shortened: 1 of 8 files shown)


tkeithblack:
@googlebot I signed it!

russellwheatley:
There hasn't been sufficient interest in this feature since its inception, and we feel this is outside the purview of this plugin. Thanks for the PR but we will be closing for now.

PR #5499, opened 2021-03-28
fix: Update README.md

Pingear:
The link does not work. It looks like an example is no longer available. 
Perhaps it can be replaced by this, at least until a more suitable one is found?


rrousselGit (approved):
LGTM!

PR #5765, opened 2021-04-14
fix(database): return ordered data

### LINKED ISSUE #753: Firebase Database on orderByChild / orderByKey returns in random order
`firebase_database: 1.0.2`

## Steps to Reproduce

Run a simple Flutter app:
https://gist.github.com/dotdoom/5180eb27084afb3b71db44f2201c6877

Most important part is (for history purposes):

```dart
(await FirebaseDatabase.instance
            .reference()
            .child(subpath)
            .orderByChild('year')
            .limitToFirst(3)
            .onValue
            .first)
        .snapshot
        .value
        .values
        .forEach((v) => debugPrint(v['year'].toString()));
```

The output is indeed the first 3 entries of a sorted list, but the entries themselves are not sorted:

```
I/flutter ( 6096): Sorted years:
I/flutter ( 6096): 1856
I/flutter ( 6096): 1867
I/flutter ( 6096): 1564
```

`limitToFirst` does not affect the order and is only to demonstrate that Firebase server gets the correct request.

I suspect there might be some issue when [transferring data between Java and Dart](https://github.com/flutter/plugins/blob/cd4829a6cf3afcf687fd064f18a028b519644d66/packages/firebase_database/android/src/main/java/io/flutter/plugins/firebase/database/FirebaseDatabasePlugin.java#L185) where the map ordering is lost. Unfortunately it's buried too deep for me to find quickly, so I hope someone with more experience in the MethodChannel mechanism can help me out. The type of the map in Flutter app is `_InternalLinkedHashMap` which IIUC is supposed to be ordered.

## Logs

```
[✓] Flutter (Channel beta, v0.5.1, on Linux, locale en_US.UTF-8)
    • Flutter version 0.5.1 at /home/artem/pkg/flutter
    • Framework revision c7ea3ca377 (7 weeks ago), 2018-05-29 21:07:33 +0200
    • Engine revision 1ed25ca7b7
    • Dart version 2.0.0-dev.58.0.flutter-f981f09760

[✓] Android toolchain - develop for Android devices (Android SDK 27.0.0)
    • Android SDK at /home/artem/Android/Sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-27, build-tools 27.0.0
    • Java binary at: /home/artem/pkg/android-studio/jre/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1024-b01)
    • All Android licenses accepted.

[✓] Android Studio (version 3.1)
    • Android Studio at /home/artem/pkg/android-studio
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1024-b01)

[!] VS Code (version 1.25.1)
    • VS Code at /usr/share/code
    • Flutter extension not installed; install from
      https://marketplace.visualstudio.com/items?itemName=Dart-Code.flutter

[✓] Connected devices (1 available)
    • Android SDK built for x86 • emulator-5554 • android-x86 • Android 8.1.0 (API 27) (emulator)

! Doctor found issues in 1 category.
```

russellwheatley:
## Description

Ensures data is ordered.

- [x] test to confirm implementation.

## Related Issues

fixes #753

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/firebase_database/lib/src/event.dart
```diff
@@ -17,26 +17,43 @@ enum _EventType {
...
+    Object value;
+
+    if (dataValue is Map<Object, Object>) {
+      value = {for (final key in childKeys) key: dataValue[key]};
+    } else if (dataValue is List<Object>) {
+      value = childKeys.map((key) => dataValue[int.parse(key)]).toList();
+    } else {
+      value = dataValue;
+    }
+    return DataSnapshot._(_data['key'], value);
```
Ehesp, line 49: This is technically breaking behaviour. If a dev is currently using the output of `value` without the expectation it will be ordered (unlikely but possible) it could cause some issues.

I think we should create a new method on the class, called `forEach` similar to how JS works (and support breaking out of the loop too).

Instead it'd look like this:

```
class DataSnapshot {
  DataSnapshot._(this._data);

  final Map<dynamic, dynamic> _data;

  /// The key of the location that generated this DataSnapshot.
  String get key => _data['key'];

  /// Returns the contents of this data snapshot as native types.
  dynamic get value => _data['value'];

  dynamic forEach(ForEachAction action) {
    // Order data into a List & apply to action
  }
}

```

What do you think?

russellwheatley: IMO, I don't think it does change the behaviour. The `childKeys` returned are unordered unless you use the `orderByChild` API which will return the `childKeys` ordered. I don't mind adding `forEach` but the existing bug will still need to be rectified anyway.

Ehesp: Would it make sense then to let "value" return the dynamic value as before, but add `forEach` which returns an iterator? At least then it allows users to keep existing behaviour but have a new feature which provides an actual typed Iterator?

russellwheatley: I may be misunderstanding, but the problem with leaving this as the same API:
```dart
 dynamic get value => _data['value'];
```

is because of the example that is illustrated in the [issue](#753):

```dart
(await FirebaseDatabase.instance
            .reference()
            .child(subpath)
            .orderByChild('year')
            .limitToFirst(3)
            .onValue
            .first)
        .snapshot
        .value
        .values
        .forEach((v) => debugPrint(v['year'].toString()));
```

This would still be unordered. I understand it's existing behaviour, but its tough to imagine a user calling `orderByChild()` and expecting an unordered list. 

On the other side, I can also appreciate that maybe somewhere out there is an existing code base  calling `orderByChild()`, perhaps accidentally, that is expecting the default query ordering. 

So I can get on board with either approach. Any thoughts @Salakar @rrousselGit?

rrousselGit: @Ehesp I don't think `forEach` should behave differently

It'd be very weird for `.value.forEach` to behave differently from `.forEach`

rrousselGit: I also do not think this is a breaking change.

Refactoring unordered to ordered is not breaking because the "unordered" order is not guaranteed. 
People shouldn't rely on the order of an unordered collection

packages/firebase_database/android/src/main/java/io/flutter/plugins/firebase/database/MethodCallHandlerImpl.java
```diff
@@ -25,6 +25,7 @@
 import com.google.firebase.database.ValueEventListener;
 import io.flutter.plugin.common.MethodCall;
 import io.flutter.plugin.common.MethodChannel;
+import java.util.ArrayList;
```
Duc56789, line 28: []()

packages/firebase_database/example/android/app/build.gradle
```diff
@@ -33,7 +33,7 @@ android {
   }
 
   defaultConfig {
-    applicationId "io.flutter.plugins.firebasedatabaseexample"
```
Duc56789, line 36: []()

packages/firebase_database/example/ios/Runner/GoogleService-Info.plist
```diff
@@ -2,39 +2,37 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>AD_UNIT_ID_FOR_BANNER_TEST</key>
-	<string>ca-app-pub-3940256099942544/2934735716</string>
-	<key>AD_UNIT_ID_FOR_INTERSTITIAL_TEST</key>
-	<string>ca-app-pub-3940256099942544/4411468910</string>
 	<key>CLIENT_ID</key>
-	<string>159623150305-1l2uq0jjcvricku3ah3fuav0v0nvrakp.apps.googleusercontent.com</string>
```
Duc56789, line 10: []()

packages/firebase_database/example/ios/Runner/GoogleService-Info.plist
```diff
@@ -2,39 +2,37 @@
...
+	<string>448618578101-28tsenal97nceuij1msj7iuqinv48t02.apps.googleusercontent.com</string>
 	<key>REVERSED_CLIENT_ID</key>
-	<string>com.googleusercontent.apps.159623150305-1l2uq0jjcvricku3ah3fuav0v0nvrakp</string>
```
Duc56789, line 12: []()

rrousselGit (approved):

PR #5867, opened 2021-04-21
fix: provide explicit types for StreamController(s).

scheglov:
IMHO it is better to have explicit types of fields, and it seems that other fields have them.
The real reason is that we are making a small change to the analyzer inference that makes it an error when there is a cycle in the references during inference. Specifically, here `_onMessageController` references itself in its initializer, so potentially its type depends on itself. It could be improved, but for now a less sophisticated way it is.


rrousselGit (approved):

rrousselGit:
Thanks!

PR #5938, opened 2021-04-26
feat: Add withConverter function to CollectionReference, DocumentReference and Query

rrousselGit:
## Description

`withConverter` transforms a reference to manipulate a custom object instead
 of a `Map<String, dynamic>`.

 This makes both read and write operations type-safe.

 ```dart
 final usersRef = FirebaseFirestore
     .instance
     .collection('users')
     .withConverter<User>(
       fromFirestore: (snapshot, _) => User.fromJson(snapshot.data()!),
       toFirestore: (user, _) => user.toJson(),
     );

 Future<void> main() async {
   // Writes now take a User as parameter instead of a Map
   await usersRef.add(User(name: 'John'));

   // Reads now return a User instead of a Map
   final User user = await usersRef.doc('123').get().then((s) => s.data());
 }
 ```

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/docs/firestore/usage.mdx
+++ b/docs/firestore/usage.mdx
@@ -412,7 +412,7 @@ Finally, we can use our new `moviesRef` variable to perform read and write opera
 Future<void> main() async {
   // Obtain science-fiction movies
   List<Movie> movies = await moviesRef
-      .where('genre', ['Sci-fi'])
+      .where('genre', isEqualTo: 'Sci-fi')
       .get()
       .then((snapshot) => snapshot.docs);
 
--- a/docs/firestore/usage.mdx
+++ b/docs/firestore/usage.mdx
@@ -368,6 +368,67 @@ database.
 For more information on what is possible when writing data to Firestore, please refer to
 this [documentation](https://firebase.google.com/docs/firestore/manage-data/add-data)
 
+## Typing CollectionReference and DocumentReference
+
+By default, Firestore references manipulate a `Map<String, dynamic>` object. The downside is that we lose type safety.  
+One solution is to use `withConverter`, which will modify methods like `CollectionReference.add` or `Query.where` to be type-safe.
+
+A common usage of `withConverter` is when combined with a serializable class, such as:
+
+```dart
+class Movie {
+  Movie({required this.title, required this.genre});
+
+  Movie.fromJson(Map<String, Object?> json)
+    : this(
+        title: json['title']! as String,
+        genre: json['genre']! as String,
+      );
+
+  final String title;
+  final String genre;
+
+  Map<String, Object?> toJson() {
+    return {
+      'title': title,
+      'genre': genre,
+    };
+  }
+}
+```
+
+We can then use `withConverter` to manipulate a collection of movies like so:
+
+```dart
+final moviesRef = Firestore.instance.collection('movies').withConverter(
+      fromFirestore: (snapshot, _) => Movie.fromJson(snapshot.data()!),
+      toFirestore: (movie, _) => movie.toJson(),
+    );
+```
+
+Finally, we can use our new `moviesRef` variable to perform read and write operations:
+
+```dart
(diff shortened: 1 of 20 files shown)


Salakar:
Shall we have a dedicated documentation page for this? Is there a specific third party library we can couple inside a complete documentation example/tutorial, e.g. `json_serializable`?

Salakar (approved):

Nolence:
This is so cool ! I'm really liking it so far :D

This might not be a common use case but I have one collection that is just composed of document Ids. That is, the documents are empty and I only use this list to index into other collections.

I was wondering how or if the `withConverter` could handle this situation -- converting to a List<String> for just document IDs. Something like have a second parameter (or a class with the values as fields) to hold the document reference/document id.

```dart
final freeQuestionIdsRef = _instance
    .collection('free_question_ids')
    .withConverter<String>(
      fromFirebase: (json, reference) => reference.id,
      // This might instead return a value key kind of thing where both might be optional ??? spitballing
      toFirebase: (String questionID) => MyMapEntry(key: questionID),
    );
```

I also imagine people will somehow want to attach their document reference to their model in some way so they can easily update that reference. I'd imagine someone just slapping them together in a tuple:

```dart
final usersRef = _instance
    .collection(_profiles)
    .withConverter<Tuple2<DocumentReference, Profile>>(
      fromFirebase: (json, reference) =>
          Tuple2(reference, Profile.fromJson(json)),
      toFirebase: (tupleData) => tupleData.item2.toJson(),
    );
```

Which of course would be further simplified when/if tuple native support lands. Also, would there be no way to make DocumentReference generic so its `get` and `set` might be typesafe as well?

Despite these suggestions, this is an awesome PR. Very much looking forward to it

rrousselGit:
> I was wondering how or if the `withConverter` could handle this situation

You should be able to do:

```dart
final freeQuestionIdsRef = _instance
    .collection('free_question_ids')
    .withConverter<String>(
      fromFirestore: (DocumentSnapshot snapshot, _) => snapshot.id,
      toFirestore: (String id, _) => const {},
    );
```

although I'm not sure that's really valuable

> Also, would there be no way to make DocumentReference generic so its get and set might be typesafe as well?

That would be a breaking change, as existing code would see their `get` called return `dynamic` instead of `Map<String, dynamic>` like it currently does.

It's possible (and preferable), but we try to avoid breaking changes

Nolence:
> although I'm not sure that's really valuable

I can see that, especially as it adds an extra an extra two parameters users will have to consider. I'll leave to maintainers' discretion. Just wanted to point out a possible use case. Thanks !

Nolence:
I imagine a lot of people will want to pass along their `WithConverterDocumentSnapshot<T>` down the widget tree so they can use the the snapshot to update the data `T`. I personally found myself creating a typedef just to shorten the name as its quite long

```dart
typedef DocSnap<T> = WithConverterQueryDocumentSnapshot<T>;
```

```dart
class Article {
  final String title;

  Article({required this.title});
}

class ArticleList extends StatelessWidget {
  final List<WithConverterQueryDocumentSnapshot<Article>> articleSnapshots;

  // Or
  final List<DocSnap<Article>> articleSnapshots;

  ArticleList(this.articleSnapshots);

  @override
  Widget build(BuildContext context) {
    return ListView.builder(
      itemCount: articleSnapshots.length,
      itemBuilder: (context, index) {
        final articleSnapshot = articleSnapshots[index];
        final article = articleSnapshot.data();

        return ListTile(
          title: Text(article.title),
          onTap: () {
            // Clear title
            articleSnapshot.reference.set(article.copyWith(title: ''));
          },
        );
      },
    );
  }
}
```

However typedefs for classes I think is only for sdk >=2.13.0. Not really offering any solution, just an observation I noticed

Salakar (approved):

PR #6046, opened 2021-05-06
feat: firebase_analytics plugin with AnalyticsWithoutAdIdSupport pod

softkot:
As a part of Analytics 7.11.0 SDK there is a special analytics pod to follow AppTrackingTransparency. As mentioned at https://github.com/FirebaseExtended/flutterfire/discussions/6030 there is no solution to use this pod in a flutter. 

This pull request involve the new FirebaseAnalyticsWithoutAdIdSupport variable and it can be used to ovveride the default pod the same way as FirebaseSDKVersion used to [override sdk version](https://firebase.flutter.dev/docs/overview/#iosmacos)

Just add following line somewere in the head of your ios/Podfile  and  got the flutter app with AnalyticsWithoutAdIdSupport

`$FirebaseAnalyticsWithoutAdIdSupport = true`

Closes #6030

### DIFF
--- a/packages/firebase_analytics/firebase_analytics/README.md
+++ b/packages/firebase_analytics/firebase_analytics/README.md
@@ -11,7 +11,7 @@ To use this plugin, add `firebase_analytics` as a [dependency in your pubspec.ya
 
 ## iOS [App Tracking Transparency](https://developer.apple.com/documentation/apptrackingtransparency)
 
-In order to avoid any user tracking you should add following line into ios/Podfile
+If you wish to use Firebase Analytics without IDFA collection capability, open your `/ios/Podfile` and add the following global variable to the top of the file:
 
 ```
 $FirebaseAnalyticsWithoutAdIdSupport = true
--- a/packages/firebase_analytics/firebase_analytics/README.md
+++ b/packages/firebase_analytics/firebase_analytics/README.md
@@ -9,6 +9,14 @@ For Flutter plugins for other Firebase products, see [README.md](https://github.
 ## Usage
 To use this plugin, add `firebase_analytics` as a [dependency in your pubspec.yaml file](https://flutter.io/platform-plugins/). You must also configure firebase analytics for each platform project: Android and iOS (see the example folder for details).
 
+## iOS [App Tracking Transparency](https://developer.apple.com/documentation/apptrackingtransparency)
+
+In order to avoid any user tracking you should add following line into ios/Podfile
+
+```
+$FirebaseAnalyticsWithoutAdIdSupport = true
+```
+
 ## Track PageRoute Transitions
 
 To track `PageRoute` transitions, add a `FirebaseAnalyticsObserver` to the list of `NavigatorObservers` on your
--- a/packages/firebase_analytics/firebase_analytics/README.md
+++ b/packages/firebase_analytics/firebase_analytics/README.md
@@ -13,7 +13,7 @@ To use this plugin, add `firebase_analytics` as a [dependency in your pubspec.ya
 
 If you wish to use Firebase Analytics without IDFA collection capability, open your `/ios/Podfile` and add the following global variable to the top of the file:
 
-```
+```ruby
 $FirebaseAnalyticsWithoutAdIdSupport = true
 ```
 
--- a/packages/firebase_analytics/firebase_analytics/ios/firebase_analytics.podspec
+++ b/packages/firebase_analytics/firebase_analytics/ios/firebase_analytics.podspec
@@ -2,6 +2,11 @@ require 'yaml'
 
 pubspec = YAML.load_file(File.join('..', 'pubspec.yaml'))
 library_version = pubspec['version'].gsub('+', '-')
+firebase_analytics = 'Firebase/Analytics'
+
+if defined?($FirebaseAnalytics)
+firebase_analytics = $FirebaseAnalytics
+end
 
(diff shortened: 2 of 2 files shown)


Salakar:
I would prefer if this was done the same way we've done on React Native Firebase; https://github.com/invertase/react-native-firebase/pull/5224 - A flag to switch to without Ad ID Support, rather than having to specify a full string.

Would you mind matching this change to how it's ben done on here: https://github.com/invertase/react-native-firebase/pull/5224

softkot:
@Salakar ok change it to bool $FirebaseAnalyticsWithoutAdIdSupport

Salakar:
> @Salakar ok change it to bool $FirebaseAnalyticsWithoutAdIdSupport

Of course

Salakar:
> @Salakar ok change it to bool $FirebaseAnalyticsWithoutAdIdSupport

Thank you! cc @mikehardy to make sure we've not missed anything here. Some documentation for this would be nice in the readme perhaps of `firebase_analytics/firebase_analytics`

mikehardy:
This seems correct, the proof is a podfile.lock with it set both ways as a test, should get the results expected based on my read of it

Dobbe (approved):

packages/firebase_analytics/firebase_analytics/README.md
```diff
@@ -9,6 +9,14 @@ For Flutter plugins for other Firebase products, see [README.md](https://github.
 ## Usage
 To use this plugin, add `firebase_analytics` as a [dependency in your pubspec.yaml file](https://flutter.io/platform-plugins/). You must also configure firebase analytics for each platform project: Android and iOS (see the example folder for details).
 
+## iOS [App Tracking Transparency](https://developer.apple.com/documentation/apptrackingtransparency)
+
+In order to avoid any user tracking you should add following line into ios/Podfile
```
russellwheatley, line 14:
```suggestion
If you wish to use Firebase Analytics without IDFA collection capability, open your `/ios/Podfile` and add the following global variable to the top of the file:
```

packages/firebase_analytics/firebase_analytics/README.md
```diff
@@ -9,6 +9,14 @@ For Flutter plugins for other Firebase products, see [README.md](https://github.
...
+
+```
```
russellwheatley, line 16:
```suggestion
```ruby
```

russellwheatley (changes requested):

mikehardy (approved):
LGTM and matches how RNFB does it, which results in the expected difference in Podfile.lock. Would still be nice to see the result of a `cat Podfile.lock |grep AdId` to verify.

Should look like 

```
mike@bistromath:~/work/Kullki/ksocialscore/packages/public-app (p2p-features *) % cat ios/Podfile.lock |grep AdId
  - Firebase/AnalyticsWithoutAdIdSupport (8.1.1):
    - FirebaseAnalytics/WithoutAdIdSupport (~> 8.0.0)
  - FirebaseAnalytics/WithoutAdIdSupport (8.0.0):
    - GoogleAppMeasurement/WithoutAdIdSupport (= 8.0.0)
  - GoogleAppMeasurement/WithoutAdIdSupport (8.0.0):
    - Firebase/AnalyticsWithoutAdIdSupport (= 8.1.1)

```

russellwheatley (approved):
LGTM

russellwheatley:
I decided to confirm to get it over the line. Here's a before (with the `$FirebaseAnalyticsWithoutAdIdSupport = true` flag) and after grep of `Podfile.lock`:
 
Screenshot 2021-07-01 at 09 33 51

kphilemon:
Hi, may I know when this will be released? I'd need this to pass the App Store review. 
I believe editing `Podfile.lock` directly is not the correct way to do so, right?

Thanks in advance!

softkot:
> Hi, may I know when this will be released? I'd need this to pass the App Store review.
> I believe editing `Podfile.lock` directly is not the correct way to do so, right?
> 
> Thanks in advance!

While it is not released you can override the dependency in pubspec.yaml from PR branch 

```
  firebase_analytics:
    git:
      url: https://github.com/softkot/flutterfire.git
      path: packages/firebase_analytics/firebase_analytics

```

masewo:
Seems not to work when also trying to use firebase_crashlytics in the same project.
Got this build failure: 'FirebaseCrashlytics/FIRAnalyticsInterop.h' file not found

mikehardy:
Interesting - I wonder if that is related to versioning - I use crashlytics and analytics-without-ad-id in a non-flutter project and the underlying pods work in that configuration. There may be something unclean in your local environment @masewo with regard to pod cache or build cache or similar

masewo:
@mikehardy I tried following cleaning steps but still get this error (by the way: I am building on a M1):
`rm -rf Pods
rm -rf Podfile.lock
rm -rf ~/.pub-cache/hosted/pub.dartlang.org/
rm -rf ~/Library/Developer/Xcode/DerivedData
pod cache clean --all
flutter pub cache repair
flutter clean
flutter pub get
pod repo update`

Dobbe:
@masewo I ran into the same problem with firebase_crashlytics 2.1.0. Didn't dig deeper into the problem but it might also be a problem with the Crashlytics library as I managed to get around it by downgrading the Crashlytics library to 2.0.7. 

If it seems to be a problem with the Crashlytics library, it might be worth it to file a bug there.

PR #6208, opened 2021-05-21
feat(firebase_database): Add support for RTDB get method.

kroikie:
## Description

Add support for RTDB `get` method

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_database/firebase_database/ios/Classes/FLTFirebaseDatabasePlugin.m
+++ b/packages/firebase_database/firebase_database/ios/Classes/FLTFirebaseDatabasePlugin.m
@@ -315,6 +315,16 @@ id roundDoubles(id value) {
   } else if ([@"OnDisconnect#cancel" isEqualToString:call.method]) {
     [getReference(database, call.arguments)
         cancelDisconnectOperationsWithCompletionBlock:defaultCompletionBlock];
+  } else if ([@"Query#get" isEqualToString:call.method]) {
+    [getReference(database, call.arguments) getDataWithCompletionBlock:^(NSError *error, FIRDataSnapshot *snapshot) {
+      result(@{
+          @"error": getDictionaryFromError(error) ?: [NSNull null],
+          @"snapshot": @{
+              @"key": snapshot.key ?: [NSNull null],
+              @"value": snapshot.value
+          }
+      });
+    }];
   } else if ([@"Query#observe" isEqualToString:call.method]) {
     FIRDataEventType eventType = parseEventType(call.arguments[@"eventType"]);
     __block FIRDatabaseHandle handle = [getDatabaseQuery(database, call.arguments)
--- a/packages/firebase_database/firebase_database/lib/src/query.dart
+++ b/packages/firebase_database/firebase_database/lib/src/query.dart
@@ -81,6 +81,30 @@ class Query {
   /// Listens for a single value event and then stops listening.
   Future<DataSnapshot> once() async => (await onValue.first).snapshot;
 
+  /// Gets the most up-to-date result for this query.
+  Future<DataSnapshot?> get() async {
+    final result = await _database._channel.invokeMethod<Map<dynamic, dynamic>>(
+      'Query#get',
+      <String, dynamic>{
+        'app': _database.app?.name,
+        'databaseURL': _database.databaseURL,
+        'path': path,
+      },
+    );
+    if (result == null || result is! Map) {
+      throw 'firebase_database get call failed from Dart.';
+    } else if (result.containsKey('error') && result['error'] != null) {
+      final errorMap = result['error'];
+      throw FirebaseException(
+          plugin: 'firebase_database',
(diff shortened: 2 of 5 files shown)


packages/firebase_database/firebase_database/lib/src/query.dart
```diff
@@ -81,6 +81,30 @@ class Query {
...
+      'Query#get',
+      <String, dynamic>{
+        'app': _database.app?.name,
+        'databaseURL': _database.databaseURL,
+        'path': path,
+      },
+    );
+    if (result == null || result is! Map) {
+      throw 'firebase_database get call failed from Dart.';
+    } else if (result.containsKey('error') && result['error'] != null) {
```
rrousselGit, line 96: Why is the `result['error']` needed? Is there a case where we would have `{'error': null}` send?

packages/firebase_database/firebase_database/lib/src/query.dart
```diff
@@ -81,6 +81,30 @@ class Query {
...
+  Future<DataSnapshot?> get() async {
+    final result = await _database._channel.invokeMethod<Map<dynamic, dynamic>>(
+      'Query#get',
+      <String, dynamic>{
+        'app': _database.app?.name,
+        'databaseURL': _database.databaseURL,
+        'path': path,
+      },
+    );
+    if (result == null || result is! Map) {
```
rrousselGit, line 94: `result is! Map` is never reached

I'm not a fan of this `if` + exception. Maybe we want `result!.containsKey('error')`?

rrousselGit (commented):
Left a few comments

packages/firebase_database/firebase_database/lib/src/query.dart
kroikie: Form the iOS side the returned object can will contain an `error` key even when the result is successful. On Android side there would be no `error` key at all.

packages/firebase_database/firebase_database/lib/src/query.dart
kroikie: Seems like I can depend on result not being null. So I'll just check for the key. Thanks!

kroikie: I was wrong added your suggestion thanks!

kroikie: There is still an instance of `if` + exception though. I'm happy to rework to avoid if you think it is needed.

packages/firebase_database/firebase_database/lib/src/query.dart
rrousselGit: But why do we need both `containsKey` and `result['error'] != null`?

rrousselGit (approved):
LGTM once `snapshot.value` is tested

packages/firebase_database/firebase_database/ios/Classes/FLTFirebaseDatabasePlugin.m
```diff
@@ -315,6 +315,16 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
   } else if ([@"OnDisconnect#cancel" isEqualToString:call.method]) {
     [getReference(database, call.arguments)
         cancelDisconnectOperationsWithCompletionBlock:defaultCompletionBlock];
+  } else if ([@"Query#get" isEqualToString:call.method]) {
+    [getReference(database, call.arguments)
+        getDataWithCompletionBlock:^(NSError *error, FIRDataSnapshot *snapshot) {
+          if (error) {
+          }
```
russellwheatley, line 322: Can remove obsolete if statement.

packages/firebase_database/firebase_database/ios/Classes/FLTFirebaseDatabasePlugin.m
```diff
@@ -315,6 +315,16 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
+          result(@{
+            @"error" : getDictionaryFromError(error) ?: [NSNull null],
+            @"snapshot" : @{@"key" : snapshot.key ?: [NSNull null], @"value" : snapshot.value}
```
russellwheatley, line 325: This should probably be:
```obj-c
@"snapshot" : @{@"key" : snapshot.key ?: [NSNull null], @"value" : snapshot.value ?: [NSNull null]}

```

russellwheatley (changes requested):

packages/firebase_database/firebase_database/lib/src/query.dart
russellwheatley: Yes, it appears iOS sends along `{'error': null}` even when successful.

russellwheatley:
@googlebot I consent.

Salakar (approved):

kroikie:
Thanks Russell!

PR #6443, opened 2021-06-22
docs(messaging): Improve Docs For FCM subscribeToTopic

### LINKED ISSUE #6434: FCM subscribeToTopic docs do not indicate if this is required to be called at app startup everytime.
Looking over these docs to learn how to subscribeToTopic...

https://firebase.flutter.dev/docs/messaging/usage/#topics

Although I assume this only needs to be called once for the lifetime of the app install, it would be nice if the docs could make clear whether this is the case, or if this needs to be called everytime on application startup.

Thanks

Patil2099:
Fixes #6434

### DIFF
--- a/docs/messaging/usage.mdx
+++ b/docs/messaging/usage.mdx
@@ -374,7 +374,7 @@ For Android, you can view Logcat logs which will give a descriptive message on w
 
 Topics are a mechanism which allow a device to subscribe and unsubscribe from named PubSub channels, all managed via FCM.
 Rather than sending a message to a specific device by FCM token, you can instead send a message to a topic and any devices
-subscribed to that topic will receive the message.
+subscribed to that topic will receive the message.It should be called everytime the app starts.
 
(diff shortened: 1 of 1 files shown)


docs/messaging/usage.mdx
```diff
@@ -374,7 +374,7 @@ For Android, you can view Logcat logs which will give a descriptive message on w
 
 Topics are a mechanism which allow a device to subscribe and unsubscribe from named PubSub channels, all managed via FCM.
 Rather than sending a message to a specific device by FCM token, you can instead send a message to a topic and any devices
-subscribed to that topic will receive the message.
+subscribed to that topic will receive the message.It should be called everytime the app starts.
```
russellwheatley, line 377: Remove "It should be called everytime the app starts." from here, and leave a code comment on line 394: 
`// subscribe to topic on each app start-up`

russellwheatley (changes requested):

PR #6819, opened 2021-08-13
feat(firestore): cache snapshot connections with underlying native listener

rrousselGit:
## Description

Currently, calling `.snapshots()` twice on the same reference leads to listening the same reference twice, which is a billable operation.

This is a common issue when using reference like so:

```dart
StreamBuilder(
  stream: collection.where(...).snapshots(),
  ...
)
```

where rebuilding the StreamBuilder with the exact same query leads to a "removeListener" followed by "addListener" on the query, which is expensive.

This PR caches the bridge with the native platform so that when rebuilding a StreamBuilder like previously mentioned, this will instead reuse the existing connection instead of opening a new one and closing the previous one.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/cloud_firestore/cloud_firestore/lib/src/snapshot_cache.dart
```diff
@@ -0,0 +1,147 @@
...
+    // StreamBuilder(stream: reference.snapshots(), ...)
+    //
+    // In this scenarion, StreamBuilder would cancel the previous stream
+    // subscription then listen to the new stream.
+    // This would call the previous controller's onCancel **before** the new
+    // controller's onListen. Which would clear the native bridge cache
+    // when the expected behaviour is to reuse the connection.
+    return Future(() {
+      connectionSub?.cancel();
+      // if (!controllerWithLastEvent.isClosed) {
```
Ehesp, line 106: Just double checking, is this if statement supposed to be commented out?

rrousselGit: Ah yes, it's just dead code

packages/cloud_firestore/cloud_firestore/lib/src/snapshot_cache.dart
```diff
@@ -0,0 +1,147 @@
...
+          },
+        );
+      };
+
+      nativeController.onCancel = () {
+        sub.cancel();
+        // When all listeners on the native bridge are removed, remove the
+        // entry from cache to avoid memory leaks
+        snapshotCache.remove(key);
+        // nativeController.close();
```
rrousselGit, line 65:
```suggestion
        snapshotCache.remove(key);
```

packages/cloud_firestore/cloud_firestore/lib/src/snapshot_cache.dart
```diff
@@ -0,0 +1,147 @@
...
+    // In this scenarion, StreamBuilder would cancel the previous stream
+    // subscription then listen to the new stream.
+    // This would call the previous controller's onCancel **before** the new
+    // controller's onListen. Which would clear the native bridge cache
+    // when the expected behaviour is to reuse the connection.
+    return Future(() {
+      connectionSub?.cancel();
+      // if (!controllerWithLastEvent.isClosed) {
+      //   controllerWithLastEvent.close();
+      // }
```
rrousselGit, line 108:
```suggestion
      connectionSub?.cancel();
```

lesnitsky:
does it make sense to add tests which verify an actual call count of the method channel?

russellwheatley (approved):
LGTM.

rrousselGit:
> does it make sense to add tests which verify an actual call count of the method channel?

That'd probably be redundant with the tests for `getCachedSnapshot` – which verify that the stream is cached already.

lesnitsky (approved):

Salakar (approved):

tcf909:
I do want to raise an issue this commit created.

You've changed the behavior for listeners of this now cached connection.

The first listener will receive a full list of docChanges (with every record being represented as added).

the N+1 listener will only see the most recent updated records in docChanges initially and instead will need to rely on just docs to get a full list of the documents in the snapshot.

Certainly a breaking change.  And goes against the documented behavior (that the initial stream listener event will contain all documents in the docChanges as added).

Salakar:
cc @rrousselGit

rrousselGit:
That's a good point.
I'm not too familiar with `docChanges`. Are we expecting the initial call to simply contains all the "docs" with nothing else?

In particular I'm wondering if we can get this behaviour back without reverting this change.

larssn:
I'm not sure they charge you for re-reads. [The official pricing docs](https://cloud.google.com/firestore/pricing) say this:

```
Listening to query results
Firestore allows you to listen to the results of a query and get realtime updates when the query results change.

When you listen to the results of a query, you are charged for a read each time a document in the result set is added or updated. You are also charged for a read when a document is removed from the result set because the document has changed. (In contrast, when a document is deleted, you are not charged for a read.)

Also, if the listener is disconnected for more than 30 minutes (for example, if the user goes offline), you will be charged for reads as if you had issued a brand-new query.
```

The last paragraph seem to indicate that the snapshot listener has to be disconnected for more than 30min to be charged again.

PR #6878, opened 2021-08-24
fix(messaging, iOS): revert onMessage event handler commit which causes another bug

helenaford:
Reverts FirebaseExtended/flutterfire#6838

From additional testing, fixes the duplication issue but causes another issue where apple throttle willPresentNotification with the presence of mutableContent in the payload

I will be pursuing other solutions as the duplicate notification bug causes a big issue in apps


russellwheatley (approved):
LGTM

PR #7050, opened 2021-09-23
fix(firebase_auth): allow setLanguage to accept null

pr-Mais:
Allow setLanguage to accept `null` as a value as per documentation.


russellwheatley (approved):
LGTM

Ehesp (approved):

PR #7111, opened 2021-10-01
refactor(firebase_auth): migrate example app to null-safety

pr-Mais:
Migrates the example app & its tests to null-safety.


russellwheatley:
Do the web tests also need to be addressed?

pr-Mais:
Yes, it's failing [here](https://github.com/FirebaseExtended/flutterfire/blob/c7a05ca48f338098a44869287e031f06f6f652fd/packages/firebase_auth/firebase_auth_web/lib/src/interop/auth.dart#L210), which I guess is another migration issue.

PR #7359, opened 2021-11-12
feat(web): automatically inject Firebase JS SDKs

Ehesp:
## Description

Automatically register plugins on web and inject scripts. See https://github.com/FirebaseExtended/flutterfire/pull/7358 for more information.


creativecreatorormaybenot:
That is a super cool feature @Ehesp 🔥 Thanks for the great work 💪

A small request: could you please consider #7365? Would be great to get your input on that.

Ehesp:
@creativecreatorormaybenot That's a tricky one - injecting the scripts is an async operation and the only async process is `initializeApp` where we have control.

I'll have a think - this change is backwards compatible though so you can continue to use the scripts in the html file yourself for now.

creativecreatorormaybenot:
@Ehesp Is there a way to disable the auto injection for a particular plugin? We could just disable it for Firebase Analytics only and continue to inject that via GTM.

Ehesp:
How would you inject it via GTM? Does it look for some context or do you manually invoke it? We have access to the window at the time of injection so we could do something similar to what the version override does.

creativecreatorormaybenot:
@Ehesp We have a **Custom HTML Tag**  that is fired based on a trigger that is based on a **Custom Event**, which is synced via Usercentrics. The custom HTML tag is just:

```html
<script src="https://www.gstatic.com/firebasejs/7.24.0/firebase-analytics.js"></script>
```

Essentially, it is invoked automatically and we have no idea when it is invoked because the user could give consent at an arbitrary point in time.

Ehesp:
I guess I could add support for something like:

```html
  <script>window.flutterfire_ignore_scripts = ['analytics'];</script>
```

Then if the array contains the registered service, it'll just skip automatically adding it. It's then up to the developer to ensure it's loaded into the page before usage.

Interestingly though, does your approach work on development? We had to have a workaround if `requirejs` is enabled (which it is on dev) since the Firebase scripts don't attach to the window if `requirejs` is available.

creativecreatorormaybenot:
@Ehesp We had to fight with making this work in development for a long time. What we ended up on was basically just:

```html
  <!--
    This script takes care of loading Firebase Analytics in the same way Google Tag Manager
    would, however, without checking user opt-in. The reason for this is that we cannot use
    Tag Manager on every domain (because it would not be feasible to add them all to the
    environments of the container). Essentially, this is for debugging purposes.
    This means that the following script only does something when the app is *not* executed on our
    official hosting domain.
  -->
  <script>
    // Let Google Tag Manager handle this on app.simpleclub.com.
    if (window.location.hostname !== '<production domain>') {
        var head = document.getElementsByTagName('head')[0];
        var script = document.createElement('script');
        script.src = 'https://www.gstatic.com/firebasejs/7.24.0/firebase-analytics.js';

        head.append(script);
    }
  </script>
```

Ehesp:
@creativecreatorormaybenot Just pushed up something to handle this, let me know what you think.

packages/firebase_core/firebase_core_web/lib/src/firebase_core_web.dart
```diff
@@ -85,6 +128,9 @@ class FirebaseCoreWeb extends FirebasePlatform {
     }
 
     String version = _firebaseSDKVersion;
+    List<String> ignored = _ignoredServiceScripts;
+
+    print(ignored);
```
creativecreatorormaybenot, line 133: This print should not stay, I suppose.

creativecreatorormaybenot (commented):
Thanks so much for this @Ehesp 🙏

PR #8043, opened 2022-02-08
feat: made option arg optional for convert fn

cedvdb:
Made the convert function option arguments optional. Not sure this is a valid change because it was made with VSCode remote repository. I assumed I could see that from the github workflow tests but I can't see any running. Maybe it requires approval to run ?

### DIFF
--- a/packages/cloud_firestore/cloud_firestore/lib/src/document_snapshot.dart
+++ b/packages/cloud_firestore/cloud_firestore/lib/src/document_snapshot.dart
@@ -5,11 +5,11 @@
 part of cloud_firestore;
 
 typedef FromFirestore<T> = T Function(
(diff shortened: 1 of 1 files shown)


russellwheatley:
Thanks for the PR. However, this isn't something we'll be considering I'm afraid. It would constitute a breaking change with the param input changes for the function types.

russellwheatley:
My dart analyzer flagged it as an issue when I updated the function type for my existing code.

cedvdb:
Okay, if I clone the repo and fix the issues will this be reconsidered or I should not bother ?

It's just annoying to have to pass null in unit tests for a parameter that is not even used, but it's really not important of a change, just quality of life thing.

PR #8214, opened 2022-03-03
docs(firebase_auth): Add content from firebase.google.com

kevinthecheung:


Salakar:
This looks great 🎉 

There's an issue asking for docs about `linkWithPopup` (web only) - do you think we should address this as part of this PR?

kroikie (approved):

kroikie:
Lets get this in and address the web only in a follow up.

PR #8235, opened 2022-03-08
feat(web): upgrade to support v9.8.1 Firebase JS SDK

Ehesp:
## Description

Upgrades the internal Firebase JS SDK to latest v9.

## Breaking Change


packages/firebase_database/firebase_database_web/lib/src/interop/app.dart
```diff
@@ -32,7 +32,7 @@ class App extends core_interop.JsObjectWrapper<AppJsImpl> {
   }
 
   /// Deletes the app and frees resources of all App's services.
-  Future delete() => core_interop.handleThenable(jsObject.delete());
+  Future delete() => core_interop.deleteApp(jsObject);
```
russellwheatley, line 35: I've reverted this change locally and pushed but it has remained 🤔.

packages/firebase_core/firebase_core_web/lib/src/firebase_core_web.dart
```diff
@@ -10,15 +10,20 @@ class FirebaseWebService {
   /// The name which matches the Firebase JS Web SDK postfix.
   String name;
 
+  /// Overrides the created window object.
+  String? override;
```
Lyokone, line 14: It's not obvious for me how to use this override from the comment. Can we add some details?

Lyokone (approved):

packages/firebase_core/firebase_core_web/lib/src/firebase_core_web.dart
```diff
@@ -75,118 +75,62 @@ class FirebaseCoreWeb extends FirebasePlatform {
 
   /// Injects a `script` with a `src` dynamically into the head of the current
   /// document.
-  Future<void> _injectSrcScript(String src) async {
+  Future<void> _injectSrcScript(String src, String windowVar) async {
     ScriptElement script = ScriptElement();
     script.type = 'text/javascript';
```
Salakar, line 80: Can we add `script.crossorigin = 'anonymous';` here as well. Without this currently unable to capture errors that originate from inside the JS, e.g. if I have some Browser/JS error reporting tool integrated into my website all errors that originate from inside a Firebase SDK will appear as  just `"Script error."` and no extra information

More context; https://blog.sentry.io/2016/05/17/what-is-script-error

Ping me internally if a repro is needed (I have one on an internal tool)

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/settings.dart
```diff
@@ -44,13 +45,20 @@ class Settings {
   /// and can be set to [Settings.CACHE_SIZE_UNLIMITED] to disable garbage collection.
   final int? cacheSizeBytes;
 
+  /// Whether to skip nested properties that are set to undefined during object serialization.
```
nilsreichardt, line 48:
```suggestion
  /// Whether to skip nested properties that are set to undefined during object serialization.
  ///
```

[Separate the first sentence of a doc comment into its own paragraph.](https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph)

packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/settings.dart
```diff
@@ -44,13 +45,20 @@ class Settings {
...
+
   /// Returns the settings as a [Map]
   Map<String, dynamic> get asMap {
     return {
       'persistenceEnabled': persistenceEnabled,
       'host': host,
       'sslEnabled': sslEnabled,
-      'cacheSizeBytes': cacheSizeBytes
+      'cacheSizeBytes': cacheSizeBytes,
+      if (ignoreUndefinedProperties) 'ignoreUndefinedProperties': true,
```
nilsreichardt, line 61: What is the reason to not do it like the other properties?
```dart
'ignoreUndefinedProperties': ignoreUndefinedProperties
```

russellwheatley: It's web only. My sense it would confuse users to keep it there when using Firestore for other platforms. Happy to reinstate if that's the consensus. Maybe even just show it if it is the web platform? Or just show it for all platforms.

russellwheatley:
Hey @jorgeroncero, I think something has messed up with the CLA agreement from this https://github.com/firebase/flutterfire/pull/8810. Do you mind signing it again for this? 😄

jorgeroncero:
Hi there @russellwheatley, my CLA is not the problem this time. If I go to the details of the cla/google step, it says it's @Lyokone who has the problem?

Lyokone:
Weird, should be good now 👍

russellwheatley (changes requested):

russellwheatley (approved):

LiangHuangBC:
query.startAfterDocument(lastDocSnapshot);  is not working some time on release mode only. In my app, there are five pages in my firestore can be fetched on dev mode correctly, but on release mode, there are only two pages.

PR #8236, opened 2022-03-08
feat(firebase_auth)!: upgrade auth web to v9 sdk

Ehesp:
## Description

Upgrades the Firebase JS SDK to v9 for auth.

## Breaking Changes

1. There is some very odd behaviour where the e2e tests for creating dummy accounts as they still appear to be "known" (cached presumably) by web SDK which is why there is a try/catch around those API setting up dummy accounts.
2.  Cannot set `photoUrl` or `displayName` to `null` on web anymore. Confirmation using Javascript:
Screenshot 2022-06-30 at 10 08 03
3. Credential is no longer supplied with any Firebase Error.

- [X] Yes, this is a breaking change.
- [] No, this is *not* a breaking change.


cedvdb:
Quick question, is the js tree shaked currently ? If not will it be with this ?

Ehesp:
Unfortunately not. The only way we're able to interact with the JavaScript SDKs at the moment is to interact with it via the window.

That means we have to download the entire module and attach it to the window.

Lyokone (approved):

Lyokone (approved):

Salakar (approved):

PR #8539, opened 2022-04-28
feat(ui): Added Flexibility to the TableBuilder

wer-mathurin:
ADD - Ability to provide your own cel builder
ADD - Ability to provide your own onEditCel
ADD - Ability to disable individual cel edition
FIX - Little optimization (do not call localization when not needed)

## Description
This widget is very cool to display table, but offer limited flexibility.  Without overthinking about all possibilities.
Just add the ability to be able to build your own widget for each cells.  This is a common scenario to have a complex object.

eg: 
```
Person{
final firstName;
final Address address;
}
```
By being able to change the cel display we can display the Address correclty.

Also adding the capability to change editCell.  Using the same example will help editing cels that containt complex objects.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


wer-mathurin:
@rrousselGit :  Is this the kind of PR that will go through?  Since I never contribute to this project, I don't know what to expect.

Salakar:
cc @lesnitsky & @rrousselGit for review, but my main takeaway is that it should be `Cell` not `Cel`

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -833,16 +865,19 @@ class _Source extends DataTableSource {
...
           DataCell(
-            _ValueView(data[head]),
-            onTap: () {
-              onEditItem(
-                doc,
-                data[head],
-                head,
-              );
-            },
+            builder?.call(data, colList.indexOf(head)) ??
```
lesnitsky, line 870: I'd rather pass an instance of the head directly instead of passing an index.
Alternatively, we could define a builder on a head instance.

@rrousselGit thoughts?

wer-mathurin: Using an index will allow this kind of logic:

```
(data, colIndex) {
        final person = Person.fromMap(data);
        if (colIndex == 0) {
          return Text(person.firstName);
        }

        if (colIndex == 1) {
          return Row(
            children: [
              Text(person.address.street),
              Text(person.address.city),
            ],
          );
        }
        return Container();
      }
```

I think relying on the header key will tie it to the header implementation (today relying on String key...tomorrow ?).

wer-mathurin: Make sense.  I didn't think of reordering column ! 🙌

Will do the change.

And will also take it into consideration for my rewrite proposal at: https://github.com/firebase/flutterfire/discussions/8569

packages/flutterfire_ui/pubspec.yaml
```diff
@@ -32,9 +32,12 @@ dependencies:
   twitter_login: ^4.2.2
 
 dev_dependencies:
+  fake_cloud_firestore: ^1.2.2
```
rrousselGit, line 35: Could you remove this dependency?

packages/flutterfire_ui/test/table_builder_test.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  testWidgets('TableBuilder WITHOUT CelBuilder is render as expected',
+      (WidgetTester tester) async {
+    // Create the widget by telling the tester to build it.
+
+    await tester.pumpWidget(_basic(collection, null));
+
+    await tester.pumpAndSettle();
+
+    // final cityFinder = find.text('FlutterFire City');
```
rrousselGit, line 28: Why is this code commented?

packages/flutterfire_ui/test/table_builder_test.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  final bob = Person(
+    firstName: 'Bob',
+    address: Address(street: 'Awesome Road', city: 'FlutterFire City'),
+  );
+  await collection.add(bob);
+
+  testWidgets('TableBuilder WITHOUT CelBuilder is render as expected',
+      (WidgetTester tester) async {
+    // Create the widget by telling the tester to build it.
```
rrousselGit, line 22: This comment seems redundant and is present twice in the file. I'd remove it

packages/flutterfire_ui/test/table_builder_test.dart
```diff
@@ -0,0 +1,182 @@
...
+  );
+}
+
+enum ColumnKey {
+  firstName,
+  address,
+}
+
+@immutable
+class Person {
```
rrousselGit, line 97: What is the purpose of using withConverter and writing such a lengthy class?

packages/flutterfire_ui/test/table_builder_test.dart
```diff
@@ -0,0 +1,182 @@
...
+    final streetFinder = find.text(bob.address.street);
+    final firstNameFinder = find.text(bob.firstName);
+
+    expect(cityFinder, findsOneWidget);
+    expect(streetFinder, findsOneWidget);
+    expect(firstNameFinder, findsOneWidget);
+  });
+}
+
+Widget _basic(
```
rrousselGit, line 75: I don't like `_basic` as a name. Maybe `dataTableBuilder`?

packages/flutterfire_ui/test/table_builder_test.dart
```diff
@@ -0,0 +1,182 @@
...
+  CollectionReference<Person> col,
```
rrousselGit, line 76: Use named parameters matching FirestoreDataTable

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -819,8 +846,9 @@ class _Source extends DataTableSource {
 
     final doc = _previousSnapshot!.docs[index];
     final data = doc.data();
-
-    return DataRow(
+    final colList = getHeaders().keys.toList();
```
rrousselGit, line 849: Could you rename colList -> columns?

rrousselGit: Although arguably I'm not sure what the value of this variable is. I prefer having this in the `for` like before – and the `toList` negatively impacts performances

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -819,8 +846,9 @@ class _Source extends DataTableSource {
...
+    return DataRow.byIndex(
```
rrousselGit, line 850: What is this change about?

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -757,15 +783,16 @@ class _Source extends DataTableSource {
...
 
-  final void Function(
-    QueryDocumentSnapshot<Map<String, Object?>> snapshot,
-    Object? value,
-    String propertyName,
-  ) onEditItem;
+  final CellBuilder? builder;
+
+  final bool canEditCell;
+  final OnEditItem onEditItem;
```
rrousselGit, line 795: Could you test those two properties? :smile:

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -58,12 +70,24 @@ class FirestoreDataTable extends StatefulWidget {
...
         ), // using an assert instead of a type because `<A, B>{}` types as `Map` but is an instance of `LinkedHashMap`
         super(key: key);
 
+  /// When specified, the builder will be use to disply your own widget for the cel
+  final CellBuilder? cellBuilder;
+
+  /// When set to false this allow row selection, defaults is set to true
+  final bool canEditCell;
+
+  /// When specified, this will be called when selecting a cel
```
rrousselGit, line 88:
```suggestion
  /// When specified, this will be called when selecting a cell
```

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -58,12 +70,24 @@ class FirestoreDataTable extends StatefulWidget {
...
+    this.cellBuilder,
+    this.canEditCell = true,
+    this.onEditItem,
   })  : assert(
           columnLabels is LinkedHashMap,
           'only LinkedHashMap are supported as header',
         ), // using an assert instead of a type because `<A, B>{}` types as `Map` but is an instance of `LinkedHashMap`
         super(key: key);
 
+  /// When specified, the builder will be use to disply your own widget for the cel
```
rrousselGit, line 82:
```suggestion
  /// When specified, the builder will be used to display your own widget for the cell
```

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -58,12 +70,24 @@ class FirestoreDataTable extends StatefulWidget {
...
+  final CellBuilder? cellBuilder;
+
+  /// When set to false this allow row selection, defaults is set to true
```
rrousselGit, line 85:
```suggestion
  /// When set to false this allows row selection, defaults to true
```

rrousselGit: Why is the comment about "selection" but the name is about edits?

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
rrousselGit: Same remark here. I don't think the word `selecting` makes sense here when considering that we're talking about edits

packages/flutterfire_ui/pubspec.yaml
wer-mathurin: This is needed in the tests.

packages/flutterfire_ui/test/table_builder_test.dart
wer-mathurin: No reason, I will clean up

packages/flutterfire_ui/test/table_builder_test.dart
wer-mathurin: That was in prep for FirestoreDataTable to support generics, so collection would need to be of the specified type.
But since I didn't included it in this commit this is fair to remove it and add it later on.

Since I'm using the DataGenerator shortcut, that was free, but I can understand that the purpose of a test is to understand the functionality and can add burden.  Will remove unused functionalities from the class.

packages/flutterfire_ui/test/table_builder_test.dart
wer-mathurin: Will do

packages/flutterfire_ui/test/table_builder_test.dart
wer-mathurin: Good idea

packages/flutterfire_ui/test/table_builder_test.dart
wer-mathurin: That make sense.

packages/flutterfire_ui/pubspec.yaml
rrousselGit: Right, but could you refactor the tests to remove the need for it?

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
wer-mathurin: I was needed in a previous commit, because I was passing the index instead of the column key name to cellBuilder.
I removed it

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
wer-mathurin: byIndex, will assign a unique index per row, so sorting will work as expected when implemented, and this is anyway a good practice to unique key per DataRow for helping the rendering behind the scene.

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
wer-mathurin: I had in mind desktop, since I'm using it on web :-)

What about ->

When specified, this will be called when tapping/selecting a cell

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
wer-mathurin: will do

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
wer-mathurin: I was trying to not break the current API and allow row selection without triggering editing a cell

I will change the name to reflect more what's happening.

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -820,29 +857,39 @@ class _Source extends DataTableSource {
...
       onSelectChanged: selectionEnabled
           ? (selected) {
               if (selected == null) return;
 
               if ((selected && _selectedRowIds.add(doc.id)) ||
                   (!selected && _selectedRowIds.remove(doc.id))) {
+                onSelectedRows?.call(
+                  _previousSnapshot!.docs
+                      .where((e) => _selectedRowIds.contains(e.id))
+                      .map((e) => e.data())
```
rrousselGit, line 872: I'd pass the snapshot itself, not the data

packages/flutterfire_ui/lib/src/firestore/table_builder.dart
```diff
@@ -35,6 +35,22 @@ import 'query_builder.dart';
 /// );
 /// ```
 /// {@endtemplate}
+///
+typedef CellBuilder = Widget Function(
+  Map<String, Object?> data,
```
rrousselGit, line 40: This probably should be the snapshot, not the data

rrousselGit (changes requested):
LGTM with the small changes I requested above

rrousselGit (approved):
LGTM. Could you fix the git conflict?

wer-mathurin:
> LGTM. Could you fix the git conflict?

DONE

wer-mathurin:
@russellwheatley I added the missing comma an run melos to see if there's any other issues, the only one I'm seeing are waring and not related to the changes I did, so I didn't change the source code

[flutterfire_ui]: warning - lib\src\auth\screens\email_verification_screen.dart:90:30 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion
[flutterfire_ui]:    info - lib\src\auth\screens\email_verification_screen.dart:73:10 - A value for optional parameter 'actionCodeSettings' isn't ever given. Try removing the unused parameter. - unused_element

russellwheatley (approved):

Issue #8548, opened 2022-04-29
🐛 [firebase_messaging]  Flutter foreground without data-notification iOS not working


arthas1888:
## Bug report

**Describe the bug**
I have a .Net Core Server it sends a msg through firebase using FirebaseAdmin.Messaging The way to send it is something like this:
               
               // Create a list containing up to 500 registration tokens.               
                var message = new MulticastMessage()
                { 
                    Tokens = registrationTokens,                       
                    Data = new Dictionary<string, string>()
                    {
                        { "class_name", request.ClassName.ToLower() },
                        { "action", request.Action.ToString() },
                        { "id", request.Id },
                        { "type", "update_model" },
                        { "all", all.ToString().ToLower() },
                        { "obj", string.IsNullOrEmpty(jsonToSend) ? "" : jsonToSend },
                    },
                };
                await FirebaseMessaging.DefaultInstance.SendMulticastAsync(message);

In flutter I receive the msg:

        FirebaseMessaging.onMessage.listen((RemoteMessage message) {
           generalLog.info('Got a message whilst in the foreground!');
          generalLog.info('Message data: ${message.data}');
          });

As you can see I send the msg without the param Notification, because I use just to update some process in flutter for instance fetch an object and update the local db in flutter for this reason I don't need to show the notification, in Android it works, but in iOS I never see this enter to the method FirebaseMessaging.onMessage.listen((RemoteMessage message)... I realize that if I put the param Notification in C# like this

            var message = new MulticastMessage()
            {
                Tokens = registrationTokens,
                Notification = new Notification
                {
                    Body = "background process",
                    Title = "background",
                },
                Data = new Dictionary<string, string> {
                    { "obj", msgToSend},
                    { "type", "chat" }
                },
            };

The msg gets in iOS, but I don't want this, because in background I receive a lot of notifications

### Steps to reproduce

My phone has the default configuration
IOS 15
iPhone 7 PLUS

### Sample project
the code is the same than this repository https://github.com/firebase/flutterfire/tree/master/packages/firebase_messaging/firebase_messaging/example
and follow this guide https://firebase.flutter.dev/docs/messaging/apple-integration

### Flutter dependencies

Run `flutter pub deps -- --style=compact` and paste the output below:

<details><summary>Click To Expand</summary>

```
     dependencies:
            flutter:
              sdk: flutter              
          
            # The following adds the Cupertino Icons font to your application.
            # Use with the CupertinoIcons class for iOS style icons.
[...]

PR #8635, opened 2022-05-11
fix(firebase_messaging): ensure silent foreground messages for iOS are called via event channel.

russellwheatley:
## Description

see title.

## Related Issues

fixes https://github.com/firebase/flutterfire/issues/8548

## Checklist

- [ ] I signed the [CLA].

## Breaking Change


Ehesp (approved):

hugovangalen:
Silent/data notifications while the app is foregrounded still did not work for me (tested with firebase_messaging 11.4.0).

I was able to trace the problem to the `didReceiveRemoteNotification` and it appears to me that the following line is flawed (line 489 in FLTFirebaseMessagingPlugin.m):

`if (notificationDict[@"aps"] != nil && notificationDict[@"aps"][@"alert"] == nil) {`

As I am logging the contents of the notiticationDict at that point, I see that it does not at all contain an `aps` property. So I wonder whether this is the intended logic there, as it wants to test for the existance of an `alert` property.

`if (notificationDict[@"aps"] == nil || notificationDict[@"aps"][@"alert"] == nil) {`

This seems to fix the issue for me.

PapyElGringo:
11.4.0 doesn't work for me neither

hugovangalen:
Indeed, hence the suggested change in my comment just above yours. https://github.com/firebase/flutterfire/pull/8635#issuecomment-1128724326

Try out that change and if it works, give it a thumbs up ;)

PapyElGringo:
@hugovangalen well I believe it's more of a data model issue than just a condition issue
Because if `notificationDict[@"aps"]` is never populated then your condition is equivalent as `if ( true || true )`

russellwheatley:
@PapyElGringo - How are you sending the message? (i.e. nodeJS admin SDK or REST API, etc)

russellwheatley:
> Silent/data notifications while the app is foregrounded still did not work for me (tested with firebase_messaging 11.4.0 that has this patch merged).
> 
> I was able to trace the problem to the `didReceiveRemoteNotification` and it appears to me that the following line is flawed (line 489 in FLTFirebaseMessagingPlugin.m):
> 
> `if (notificationDict[@"aps"] != nil && notificationDict[@"aps"][@"alert"] == nil) {`
> 
> As I am logging the contents of the notiticationDict at that point, I see that it does not at all contain an `aps` property. So I wonder whether this is the intended logic there, as it wants to test for the presence (or rather, lack) of an `alert` property.
> 
> `if (notificationDict[@"aps"] == nil || notificationDict[@"aps"][@"alert"] == nil) {`
> 
> This seems to fix the issue for me.

@hugovangalen - this won't work because it will fire twice on `onMessage` API if notification + data is present on the message. This code change is strictly firing for data only messages (i.e. when `alert` property is not present on `aps` dictionary).

hugovangalen:
> @hugovangalen - this won't work because it will fire twice on `onMessage` API if notification + data is present on the message. This code change is strictly firing for data only messages (i.e. when `alert` property is not present on `aps` dictionary).

Right okay, I see. I guess I got distracted by the comment above the condition that says "if alert is present, this will be called by the other "Messaging#onMessage" channel handler", which lead me to think that it only dives in this specific routine for data only messages. If the payload has a notification, it goes through another handler `userNotificationCenter` (where the 'aps' property is also missing, and the notification text data is in the "notification" property). I do not observe a double onMessage triggering though with my change.

Anyways, yeah, @PapyElGringo may be right in the 'aps' getting stripped somewhere down the line, but it's stripped before these native handlers get called so that may imply it's a FCM service issue that strips it way? (I also noticed that "to" field gets stripped from the "data" payload, but that is probably another issue altogether.)

hugovangalen:
> @hugovangalen well I believe it's more of a data model issue than just a condition issue Because if `notificationDict[@"aps"]` is never populated then your condition is equivalent as `if ( true || true )`

Yet if it only gets populated when an "alert", or other property is present, then this condition does make sense. 

But you are right, I did some more logging and the 'aps' property is not filled in anywhere when it arrives in the native handlers, so it is indeed a bigger problem that's possibly not fixable by altering the plugin code.

hugovangalen:
> @hugovangalen well I believe it's more of a data model issue than just a condition issue Because if `notificationDict[@"aps"]` is never populated then your condition is equivalent as `if ( true || true )`

I erroneously thought that the notificationDict is what gets delivered by the operating system to the app, yesterday I overlooked the call to [FLTFirebaseMessagingPlugin remoteMessageUserInfoToDict:userInfo] which is where the problem arises, 'aps' is present in userInfo, but not in the resulting dictionary.

In fact, in remoteMessageUserInfoToDict, this is happening:

```
    // build data dict from remaining keys but skip keys that shouldn't be included in data
    if ([key isEqualToString:@"aps"] || [key hasPrefix:@"gcm."] || [key hasPrefix:@"google."]) {
      continue;
    }
```

It seems that the 'aps' *intensionally* gets stripped. It kinda does make sense to strip it, as it is operating system specific data that doesn't need to end up in the Flutter app, but it does break the condition back in didReceiveRemoteNotification.

@russellwheatley I suspect that in stead of notificationDict, the userInfo should be tested in line 489:
`if (userInfo[@"aps"] != nil && userInfo[@"aps"][@"alert"] == nil) {`

PapyElGringo:
@russellwheatley I use `sendMulticast` API

hugovangalen:
@PapyElGringo Could you please cconfirm that for you, when you dump the `userInfo` dictionary, that *does* contain the 'aps' property?

PapyElGringo:
My background notifications are blocked because of quotas and priority

russellwheatley:
> > @hugovangalen well I believe it's more of a data model issue than just a condition issue Because if `notificationDict[@"aps"]` is never populated then your condition is equivalent as `if ( true || true )`
> 
> I erroneously thought that the notificationDict is what gets delivered by the operating system to the app, yesterday I overlooked the call to [FLTFirebaseMessagingPlugin remoteMessageUserInfoToDict:userInfo] which is where the problem arises, 'aps' is present in userInfo, but not in the resulting dictionary.
> 
> In fact, in remoteMessageUserInfoToDict, this is happening:
> 
> ```
>     // build data dict from remaining keys but skip keys that shouldn't be included in data
>     if ([key isEqualToString:@"aps"] || [key hasPrefix:@"gcm."] || [key hasPrefix:@"google."]) {
>       continue;
>     }
> ```
> 
> It seems that the 'aps' _intensionally_ gets stripped. It kinda does make sense to strip it, as it is operating system specific data that doesn't need to end up in the Flutter app, but it does break the condition back in didReceiveRemoteNotification.
> 
> @russellwheatley I suspect that in stead of notificationDict, the userInfo should be tested in line 489: `if (userInfo[@"aps"] != nil && userInfo[@"aps"][@"alert"] == nil) {`

@hugovangalen thanks for the investigation. I will double check and get back to you on this 🙏

Issue #8444, opened 2022-04-11
🐛 [firebase_auth] java.util.ConcurrentModificationException: null


stx:
## Bug report

We found this unhandled error on production. We have a standard implementation and make Firebase Auth calls sparingly.

```
java.util.ConcurrentModificationException: null
    at java.util.ArrayList$Itr.next(ArrayList.java:860)
    at io.flutter.plugins.firebase.auth.FlutterFirebaseAuthPlugin.parseUserInfoList(FlutterFirebaseAuthPlugin.java:2)
    at io.flutter.plugins.firebase.auth.FlutterFirebaseAuthPlugin.parseFirebaseUser(FlutterFirebaseAuthPlugin.java:14)
    at io.flutter.plugins.firebase.auth.IdTokenChannelStreamHandler.a(IdTokenChannelStreamHandler.java:5)
    at io.flutter.plugins.firebase.auth.m0.a
    at com.google.firebase.auth.p1.run(com.google.firebase:firebase-auth@@21.0.3:4)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:226)
    at android.os.Looper.loop(Looper.java:313)
    at android.app.ActivityThread.main(ActivityThread.java:8663)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:567)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
```

## Additional context

---

### Flutter doctor

Run `flutter doctor` and paste the output below:

<details><summary>Click To Expand</summary>

```
PASTE OUTPUT INSIDE HERE
```

</details>

---

### Flutter dependencies

Run `flutter pub deps -- --style=compact` and paste the output below:

<details><summary>Click To Expand</summary>

```
Dart SDK 2.16.2
Flutter SDK 2.10.4
tessie 4.4.3+251

dependencies:
- auto_size_text 3.0.0 [flutter]
- cached_network_image 3.2.0 [flutter flutter_cache_manager octo_image cached_network_image_platform_interface cached_network_image_web]
- clipboard 0.1.3 [flutter]
- cloud_firestore 3.1.11 [cloud_firestore_platform_interface cloud_firestore_web collection firebase_core firebase_core_platform_interface flutter meta]
[...]

PR #8967, opened 2022-06-26
fix(firebase_auth): java.util.ConcurrentModificationException

softkot:
try to fix https://github.com/firebase/flutterfire/issues/8444

### DIFF
--- a/packages/firebase_auth/firebase_auth/android/src/main/java/io/flutter/plugins/firebase/auth/FlutterFirebaseAuthPlugin.java
+++ b/packages/firebase_auth/firebase_auth/android/src/main/java/io/flutter/plugins/firebase/auth/FlutterFirebaseAuthPlugin.java
@@ -343,7 +343,7 @@ public class FlutterFirebaseAuthPlugin
       List<? extends UserInfo> userInfoList) {
     List<Map<String, Object>> output = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


mikehardy:
Nice! Thanks for giving this an attempt @softkot - @russellwheatley it looked like you were involved with this one from #8444 ? On that issue and the linked issue in firebase-android-sdk it appeared people were conflating an OutOfMemory error and this ConcurrentModificationException.

firebase-android-sdk 30.2.0 *allegedly* fixes the out of memory problem, but this ConcurrentModification thing was still happening. I proposed what I thought was the easy fix here of shallow-copying the provider list so we may safely iterate, but you may know about the technical requirements. Specifically, obviously the underlying collection was modified while the code iterated, is it okay to return stale data now that we shallow copy? This PR is fine if returning results of first fetch of provider data is fine even if obsolete. If not then I guess you have to catch ConcurrentModificationException and retry the iterator or ... something ? :-)

Hopefully this works though as it definitely is the quick + easy fix. Cheers

russellwheatley:
Hey @mikehardy, thanks for the heads up. I think the shallow copy implementation should be fine, we don't need that particular data to be that up-to-date, especially now it is causing this bug.

russellwheatley (approved):

Lyokone (approved):

Issue #8861, opened 2022-06-08
🐛 [firebase_analytics] Error when trying to send a list when logging event


victor-tinoco:
## Bug report

**Describe the bug**
A `PlatformException` is thrown when trying to call `logEvent` passing a `List` on `parameter`.

This problem seems to be related to #4353 and #2835 (which was already addressed), so I can't understand why it's reproducing on the latest version. I'm using Android platform for this scenario.

### Steps to reproduce
For example, when logging event:
```dart
FirebaseAnalytics.instance.logEvent(
  name: 'my_event_name',
  parameters: {
    'items': ['1', '2'],
  },
);
```
Logs the following on console:

<img width="876" alt="image" src="https://user-images.githubusercontent.com/49794086/172694387-c4734d8c-a6f1-4693-9200-6079dd92c2a9.png">

### Expected behavior

It should accept lists.

---

### Flutter doctor

Run `flutter doctor` and paste the output below:

<details><summary>Click To Expand</summary>

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel unknown, 3.0.1, on macOS 12.1 21C52 darwin-x64, locale en-BR)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 13.4)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
[✓] VS Code (version 1.67.2)
[✓] Connected device (3 available)
[✓] HTTP Host Availability

• No issues found!
```

</details>

---

### Flutter dependencies

Run `flutter pub deps -- --style=compact` and paste the output below:

<details><summary>Click To Expand</summary>

```
Dart SDK 2.17.1
Flutter SDK 3.0.1
trader 2.31.02+1000090

dependencies:
- async 2.8.2 [collection meta]
- cached_network_image 3.2.0 [flutter flutter_cache_manager octo_image cached_network_image_platform_interface cached_network_image_web]
- clock 1.1.0
- cloud_firestore 3.1.11 [cloud_firestore_platform_interface cloud_firestore_web collection firebase_core firebase_core_platform_interface flutter meta]
- collection 1.16.0
- copy_with_extension 4.0.0
- cpfcnpj 2.0.0
- datadog_flutter 2.0.0 [meta logging http flutter flutter_web_plugins js]
- decimal 2.1.0 [rational]
[...]

Issue #4353, opened 2020-12-08
🐛 [firebase_analytics] Cannot pass List inside logEvent parameters


monisnap-julien:
## Bug report

**Describe the bug**
When passing an object of dart type `List` to the `logEvent` function, it causes an error due to passing an object that is converted to `ArrayList` in Java (Android side). The following is logged in the console :

```
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): Failed to handle method call
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): java.lang.IllegalArgumentException: Unsupported value type: java.util.ArrayList
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at io.flutter.plugins.firebaseanalytics.FirebaseAnalyticsPlugin.createBundleFromMap(FirebaseAnalyticsPlugin.java:188)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at io.flutter.plugins.firebaseanalytics.FirebaseAnalyticsPlugin.handleLogEvent(FirebaseAnalyticsPlugin.java:118)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at io.flutter.plugins.firebaseanalytics.FirebaseAnalyticsPlugin.onMethodCall(FirebaseAnalyticsPlugin.java:84)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:233)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at io.flutter.embedding.engine.dart.DartMessenger.handleMessageFromDart(DartMessenger.java:85)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(FlutterJNI.java:692)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at android.os.MessageQueue.nativePollOnce(Native Method)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at android.os.MessageQueue.next(MessageQueue.java:326)
E/MethodChannel#plugins.flutter.io/firebase_analytics(20572): 	at android.os.Looper.loop(Looper.java:160)
[...]

Issue #2835, opened 2020-06-25
[firebase_analytics] List type paramters


najmsaqib:
Firebase Analytics documents says that "ITEMS" parameter supports "Parcelable[]" for android and "NSArray" for IOS but this feature is missing in Flutter firebase_analytics 5.0.15 and we can't pass List to ITEMS parameter.

Documents for Android:
https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event#ADD_TO_CART

Documents for IOS:
https://firebase.google.com/docs/reference/swift/firebaseanalytics/api/reference/Constants#/c:FIREventNames.h@kFIREventAddToCart

GursheeshSingh:
Even for logViewItemList

keobrien:
+1 this would be extremely useful for working around any events that are not yet implemented like the new purchase event that replaces the ecommerce_purchase event. Also, iOS seems to allow the event with the list to be sent but Android does not. Looks like the issue is in android/src/main/java/io/flutter/plugins/firebaseanalytics/FirebaseAnalyticsPlugin.java createBundleFromMap does not support an Array List.

Error message:
`java.lang.IllegalArgumentException: Unsupported value type: java.util.ArrayList`

CameliaTripon:
When can we expect this to be fixed? It is really an issue which postpones our release since the Enhanced Ecommerce bundle list is requested by our marketing team.

rubenvereecken:
We too need List type parameters.

CameliaTripon:
Hello. I saw that the PR with the fix was merged to master. Any idea when we will have a release with this?

PR #9520, opened 2022-09-12
feat(firebase_analytics): update `logEvent()` & `setDefaultParameters()` to assert input types. 

russellwheatley:
## Description

This PR is designed to avoid confusion about what values are accepted when logging a custom event via `logEvent()`.  I've implemented an assert on the values passed to `logEvent` which throws when the parameter values are not `String` or `num`. The same for `setDefaultEventParameters()` except it also allows parameter values to be `null` which resets that value as a default parameter.

The Firebase documentation states that the only values accepted are strings and numbers:

From android documentation (see bold):
> The map of event parameters. Passing null indicates that the event has no parameters. Parameter names can be up to 40 characters long and must start with an [alphabetic](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLetter-int-) character and contain only [alphanumeric](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLetterOrDigit-int-) characters and underscores. __String, long and double param types are supported__. String parameter values can be up to 100 characters long. The "firebase_", "google_" and "ga_" prefixes are reserved and should not be used for parameter names.

I've also tested on web, iOS and android:

__android__:
Successful custom event without `items` list:
android successful custom event  No item

Unsuccessful custom event including `items` list  (Note error code 21):
android unsuccessful event  with item

__iOS__:
Successful custom event without `items` list:
iOS successful custom event  no item

Unsuccessful custom event including `items` list (Note error code 21):
iOS unsuccessful custom event  with item

__Web__:
Web is slightly different as it uses a Chrome extension to debug. It does pass the items list in the request for a custom event, but it appears to stringify the value. See:
```js
logEvent(
  analytics, 
  'test_event',
  { some_values: [{some:'thing'}, 3, 'hello'] }
);
```
The result in the debugger console for `some_values` looks like this:

```
ep.some_values=%5Bobject%20Object%5D%2C3%2Chello&_et=3
```

Once decoded, it shows the value as one string (including the object within the array, hence `[object Object]`)
```
[object Object],3,hello&_et=3
```

## setDefaultEventParameters

I've also changed `Int` to `Long` for the event parameters bundle in android. The documentation for android says `Long` & `Double` only. If you try to add `Int` when running `setDefaultEventParameters()` you will get the following warning in the console logs:

```
W/FA: Invalid default event parameter type. Name, value: someInt, 400
W/FA: Invalid default event parameter type. Name, value: longNumber, 230485
W/FA: Invalid default event parameter type. Name, value: inter, 9
```

It also doesn't show up in the debugger console. 
Screenshot 2022-12-09 at 14 05 17

When I changed `putInt()` to `putLong()`, the warnings disappear and the default parameters appear in the console:
Screenshot 2022-12-09 at 14 07 46

## TODO
- [x] Use `EventParameters` for `setDefaultEventParameters()` as this only allows strings , null and numbers. See: https://github.com/firebase/flutterfire/issues/9698

## Related Issues
closes https://github.com/firebase/flutterfire/issues/8861, https://github.com/firebase/flutterfire/issues/9698

Other related issues which highlight confusion:
https://github.com/firebase/flutterfire/issues/4353
https://github.com/firebase/flutterfire/issues/2835

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

- [X] Yes, this is a breaking change.


packages/firebase_analytics/firebase_analytics/example/lib/main.dart
```diff
@@ -67,18 +67,18 @@ class _MyHomePageState extends State<MyHomePage> {
...
       name: 'test_event',
-      parameters: <String, dynamic>{
-        'string': 'string',
-        'int': 42,
-        'long': 12345678910,
-        'double': 42.0,
-        // Only strings and numbers (ints & doubles) are supported for GA custom event parameters:
-        // https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets#overview
-        'bool': true.toString(),
-        'items': [itemCreator()]
```
Lyokone, line 80: We don't have items anymore?

packages/firebase_analytics/firebase_analytics/example/lib/main.dart
```diff
@@ -67,18 +67,18 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 
   Future<void> _sendAnalyticsEvent() async {
+    EventParameters params = EventParameters();
```
Lyokone, line 70: Maybe we could have something like EventParameters.fromMap({...}) that does all the checks in one time? It would ease migration

Lyokone (changes requested):

packages/firebase_analytics/firebase_analytics/example/lib/main.dart
russellwheatley: Not for custom events (`logEvent()`). See the description for why 😄

packages/firebase_analytics/firebase_analytics/example/lib/main.dart
russellwheatley: Yeah, I was thinking of that. I'll add that method 👍

russellwheatley: I think this is something we should bring up in the meeting to be sure that items are not available for custom events as well 🤔. My own testing says no but to be 💯 sure.

packages/firebase_analytics/firebase_analytics_platform_interface/lib/src/event_parameters.dart
```diff
@@ -0,0 +1,43 @@
...
+    });
+    return EventParameters._(parameters);
+  }
+
+  EventParameters._(this._parameters);
+
+  Map<String, Object> _parameters = {};
+
+  // Can only pass either a String or a num value
+  EventParameters addParameter(String key, {String? string, num? number}) {
```
mateusfccp, line 24: IMO, this is not much better. We are allowing the use to pass both of them, and he's going to have a runtime exception if he does it. We could make a safe API that wouldn't allow the user to pass any kind of invalid data, so that, if it compiles, it works.

packages/firebase_analytics/firebase_analytics_platform_interface/lib/src/event_parameters.dart
```diff
@@ -0,0 +1,43 @@
+const String _exceptionMessage =
+    "'string' OR 'number' must be set as the value of the parameter";
+
+class EventParameters {
```
victor-tinoco, line 4: Isn't it, again, error-prone?

API-wise, the user is able to provide both `String` and `num`. Implementation-wise we throw an exception for a behavior already allowed by its API.

I was thinking about something like..

```dart
abstract class EventParameter {
  const factory EventParameter.bool(bool value) = EventParameterBool;
  const factory EventParameter.string(String value) = EventParameterString;
  const factory EventParameter.num(num value) = EventParameterNum;
  const factory EventParameter.map(Map<String, EventParameter> value) = EventParameterMap;

  Object get value;

  T when<T>(
    T Function(bool value) bool,
    T Function(String value) string,
    T Function(num value) num,
    T Function(Map<String, EventParameter> value) map,
  );
}

class EventParameterBool implements EventParameter {
  const EventParameterBool(this.value);

  @override
  final bool value;

  @override
  String toString() => 'EventParameterBool($value)';

  @override
  T when<T>(
    T Function(bool value) bool,
    T Function(String value) string,
    T Function(num value) num,
    T Function(Map<String, EventParameter> value) map,
  ) {
    return bool(value);
  }
}

class EventParameterString implements EventParameter {
  const EventParameterString(this.value);

  @override
  final String value;

  @override
  String toString() => 'EventParameterString($value)';

  @override
  T when<T>(
    T Function(bool value) bool,
    T Function(String value) string,
    T Function(num value) num,
    T Function(Map<String, EventParameter> value) map,
  ) {
    return string(value);
  }
}

class EventParameterNum implements EventParameter {
  const EventParameterNum(this.value);

  @override
  final num value;

  @override
  String toString() => 'EventParameterNum($value)';

  @override
  T when<T>(
    T Function(bool value) bool,
    T Function(String value) string,
    T Function(num value) num,
    T Function(Map<String, EventParameter> value) map,
  ) {
    return num(value);
  }
}

class EventParameterMap implements EventParameter {
  const EventParameterMap(this.value);

  @override
  final Map<String, EventParameter> value;

  @override
  String toString() => 'EventParameterMap($value)';

  @override
  T when<T>(
    T Function(bool value) bool,
    T Function(String value) string,
    T Function(num value) num,
    T Function(Map<String, EventParameter> value) map,
  ) {
    return map(value);
  }
}
```

victor-tinoco: Also, the `logEvent` API can have a `Map<String, CustomEventParameter> parameters` or something like, so we would not introduce a new object which just contains a `Map` and `asMap`.

victor-tinoco: Also, funny enough, if you think good we can provide an _iterable_ constructor, encoding to / decoding from a `String` internally. 😅

packages/firebase_analytics/firebase_analytics_platform_interface/lib/src/event_parameters.dart
russellwheatley: Fair point. I think I will change `addParameter` to `addString` & `addNumber` to improve it. 

I will be keeping the `fromMap()` function, as its main purpose is to help transition to the new `EventParameters` API, and it will still throw assert on incorrect property values before it ever makes it to the native side.

packages/firebase_analytics/firebase_analytics_platform_interface/lib/src/event_parameters.dart
russellwheatley: Hey @victor-tinoco, thanks for your update. I'll be updating original implementation as mentioned [here](https://github.com/firebase/flutterfire/pull/9520#discussion_r970711696)

packages/firebase_analytics/firebase_analytics/example/lib/main.dart
```diff
@@ -67,18 +67,18 @@ class _MyHomePageState extends State<MyHomePage> {
...
     await widget.analytics.logEvent(
       name: 'test_event',
-      parameters: <String, dynamic>{
-        'string': 'string',
-        'int': 42,
-        'long': 12345678910,
-        'double': 42.0,
-        // Only strings and numbers (ints & doubles) are supported for GA custom event parameters:
-        // https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets#overview
-        'bool': true.toString(),
```
victor-tinoco, line 79: WDYT about introduce an `addBool` method, which behind the scenes saves it as `value.toString()`?

russellwheatley: I would prefer not as it isn't stored as a boolean and hence misleading, `true.toString()` is painless enough.

pr-Mais (approved):

Lyokone (approved):

mockturtl:
> From android documentation (see bold):
> > The map of event parameters. Passing null indicates that the event has no parameters. Parameter names can be up to 40 characters long and must start with an [alphabetic](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLetter-int-) character and contain only [alphanumeric](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLetterOrDigit-int-) characters and underscores. String, long and double param types are supported. String parameter values can be up to 100 characters long. The "firebase_", "google_" and "ga_" prefixes are reserved and should not be used for parameter names.

Source: https://developers.google.com/android/reference/com/google/firebase/analytics/FirebaseAnalytics#logEvent(java.lang.String,%20android.os.Bundle)

Issue #10541, opened 2023-02-28
🐛 [cloud_functions] Can't call 2nd gen functions due to missing `httpsCallableFromURL` method


jhuleatt:
## Bug report

**Describe the bug**
2nd gen callable functions don't yet have stable URLs, so [an extra argument is needed to get the callable instance](https://firebase.google.com/docs/functions/beta/callable#call_the_function). Since FlutterFire doesn't support this, it isn't currently possible to call 2nd gen callable functions using FlutterFire.

### Expected behavior

FlutterFire offers a function that takes a callable function's URL as an argument, similar to iOS' `httpsCallable(URL)` overload, the JS SDK's `httpsCallableFromURL`, and Android's `getHttpsCallableFromUrl`.

## Additional context

1st gen functions have stable, guessable `cloudfunctions.net` URLs, so the client SDKs don't need to be told what a function's URL is. We're working on adding stable URLs to 2nd gen functions, but until they're available, URLs look like `https://<functionName>-<randomHash>-<region>.a.run.app/addMessage` and aren't guessable.

jbaptisteroesch:
Hi,
I tested this solution for Gen2 functions. It appears that depending on the region, it may or may not work. I tested it on eu-west-3 and it did not work, but on eu-west-1 it did work. I'm not sure what the cause of this discrepancy is, but [this article led me to wonder if the problem might be related to the location](https://cloud.google.com/run/docs/mapping-custom-domains). It's possible that the location isn't related at all, but it's worth considering.

Lyokone:
Hi, 
You can find the list of supported region here: https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.md#supportedregion

It doesn't appear that eu-west-3 is part of this list.

Lyokone:
I'm closing this issue since Cloud Function Gen 2 has landed on the SDK.

PR #10545, opened 2023-03-01
feat(functions): add support for 2nd gen functions

Lyokone:
## Description

There are now 2 new functions, 
httpsCallableFromUrl and httpsCallableFromUri that can take a String Url or a parsed Uri to use a 2nd gen Cloud Function.

## Related Issues

#10541 

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/cloud_functions/cloud_functions/example/lib/main.dart
+++ b/packages/cloud_functions/cloud_functions/example/lib/main.dart
@@ -62,7 +62,7 @@ class _MyAppState extends State<MyApp> {
               children: [
                 FloatingActionButton.extended(
                   onPressed: () async {
-                    // See index.js in .github/workflows/scripts for the example function we
+                    // See .github/workflows/scripts/functions/src/index.ts for the example function we
                     // are using for this example
                     HttpsCallable callable =
                         FirebaseFunctions.instance.httpsCallable(
@@ -81,7 +81,7 @@ class _MyAppState extends State<MyApp> {
                 const SizedBox(height: 10),
                 FloatingActionButton.extended(
                   onPressed: () async {
-                    // See index.js in .github/workflows/scripts for the example function we
+                    // See .github/workflows/scripts/functions/src/index.ts for the example function we
                     // are using for this example
                     HttpsCallable callable =
                         FirebaseFunctions.instance.httpsCallable(
--- a/packages/cloud_functions/cloud_functions/example/lib/main.dart
+++ b/packages/cloud_functions/cloud_functions/example/lib/main.dart
@@ -49,40 +49,75 @@ class _MyAppState extends State<MyApp> {
             },
           ),
         ),
-        floatingActionButton: Builder(
-          builder: (context) => FloatingActionButton.extended(
-            onPressed: () async {
-              // See index.js in .github/workflows/scripts for the example function we
-              // are using for this example
-              HttpsCallable callable = FirebaseFunctions.instance.httpsCallable(
-                'listFruit',
-                options: HttpsCallableOptions(
-                  timeout: const Duration(seconds: 5),
-                ),
-              );
-
-              try {
-                final result = await callable();
-                setState(() {
-                  fruit.clear();
-                  result.data.forEach((f) {
-                    fruit.add(f);
-                  });
-                });
-              } catch (e) {
-                ScaffoldMessenger.of(context).showSnackBar(
-                  SnackBar(
-                    content: Text('ERROR: $e'),
+        floatingActionButton: Column(
+          mainAxisSize: MainAxisSize.min,
+          crossAxisAlignment: CrossAxisAlignment.end,
+          children: [
+            FloatingActionButton.extended(
+              onPressed: () async {
+                // See index.js in .github/workflows/scripts for the example function we
+                // are using for this example
+                HttpsCallable callable =
+                    FirebaseFunctions.instance.httpsCallable(
+                  Uri.parse(
+                    'http://localhost:5001/flutterfire-e2e-tests/us-central1/listfruits2ndgen',
+                  ),
+                  options: HttpsCallableOptions(
+                    timeout: const Duration(seconds: 5),
                   ),
                 );
-              }
-            },
-            label: const Text('Call Function'),
-            icon: const Icon(Icons.cloud),
-            backgroundColor: Colors.deepOrange,
-          ),
+
+                await callingFunction(callable, context);
+              },
+              label: const Text('Call Function'),
+              icon: const Icon(Icons.cloud),
+              backgroundColor: Colors.deepOrange,
+            ),
+            const SizedBox(height: 10),
+            FloatingActionButton.extended(
+              onPressed: () async {
+                // See index.js in .github/workflows/scripts for the example function we
+                // are using for this example
+                HttpsCallable callable =
+                    FirebaseFunctions.instance.httpsCallable(
+                  Uri.parse(
+                    'http://localhost:5001/flutterfire-e2e-tests/us-central1/listfruits2ndgen',
+                  ),
+                  options: HttpsCallableOptions(
+                    timeout: const Duration(seconds: 5),
+                  ),
+                );
+
+                await callingFunction(callable, context);
+              },
+              label: const Text('Call 2nd Gen Function'),
+              icon: const Icon(Icons.cloud),
+              backgroundColor: Colors.deepOrange,
+            ),
+          ],
         ),
       ),
     );
+  }
+
+  Future<void> callingFunction(
+    HttpsCallable callable,
+    BuildContext context,
+  ) async {
+    try {
+      final result = await callable();
+      setState(() {
+        fruit.clear();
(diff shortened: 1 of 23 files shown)


packages/cloud_functions/cloud_functions/example/lib/main.dart
```diff
@@ -50,39 +55,76 @@ class _MyAppState extends State<MyApp> {
...
-                options: HttpsCallableOptions(
-                  timeout: const Duration(seconds: 5),
+          builder: (context) {
+            return Column(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.end,
+              children: [
+                FloatingActionButton.extended(
+                  onPressed: () async {
+                    // See index.js in .github/workflows/scripts for the example function we
```
russellwheatley, line 65:
```suggestion
                    // See .github/workflows/scripts/functions/src/index.ts for the example function we
```

packages/cloud_functions/cloud_functions/example/lib/main.dart
```diff
@@ -50,39 +55,76 @@ class _MyAppState extends State<MyApp> {
...
+                  },
+                  label: const Text('Call Function'),
+                  icon: const Icon(Icons.cloud),
+                  backgroundColor: Colors.deepOrange,
                 ),
-              );
+                const SizedBox(height: 10),
+                FloatingActionButton.extended(
+                  onPressed: () async {
+                    // See index.js in .github/workflows/scripts for the example function we
```
russellwheatley, line 84:
```suggestion
                    // See .github/workflows/scripts/functions/src/index.ts for the example function we
```

tests/integration_test/cloud_functions/cloud_functions_e2e_test.dart
```diff
@@ -120,6 +122,20 @@ void main() {
           expect(result.data, isA<Map<String, dynamic>>());
         },
       );
+
+      test('can be called using an Uri', () async {
+        final localhostMapped =
+            kIsWeb || !Platform.isAndroid ? 'localhost' : '10.0.2.2';
```
russellwheatley, line 128: Is this needed: `kIsWeb ||`? If it is not android, it will hit `localhost`, right?

tests/integration_test/e2e_test.dart
```diff
@@ -32,17 +32,17 @@ void main() {
...
-    firebase_ml_model_downloader.main();
-    firebase_remote_config.main();
-    firebase_storage.main();
+    // firebase_app_check.main();
+    // firebase_app_installations.main();
+    // firebase_dynamic_links.main();
+    // firebase_messaging.main();
+    // firebase_ml_model_downloader.main();
+    // firebase_remote_config.main();
+    // firebase_storage.main();
```
russellwheatley: reinstate

russellwheatley (changes requested):

tests/integration_test/cloud_functions/cloud_functions_e2e_test.dart
Lyokone: Platform is not available on Web, so you have to use kIsWeb to remove the "Platform" when on Web during compilation :)

tests/integration_test/e2e_test.dart
Lyokone: Yes I've pushed them back on the last commit, 👍

tests/integration_test/cloud_functions/cloud_functions_e2e_test.dart
russellwheatley: ah, yes, I had a vague memory of something like that 👍

packages/cloud_functions/cloud_functions/ios/Classes/FLTFirebaseFunctionsPlugin.m
```diff
@@ -93,7 +94,17 @@ - (void)httpsFunctionCall:(id)arguments withMethodCallResult:(FLTFirebaseMethodC
...
 
-  FIRHTTPSCallable *function = [functions HTTPSCallableWithName:functionName];
+  FIRHTTPSCallable *function;
+
+  if (![functionName isEqual:[NSNull null]]) {
+    function = [functions HTTPSCallableWithName:functionName];
+  } else if (![functionUri isEqual:[NSNull null]]) {
+    function = [functions HTTPSCallableWithURL:[NSURL URLWithString:functionUri]];
+  } else {
+    result.error(@"WRONG_ARGUMENT", @"Either functionName or functionUri should be provided", nil,
```
russellwheatley, line 104: Sorry, I just noticed this error code: `@"WRONG_ARGUMENT"`. Perhaps we should have it matching on android as well? Also - probably make it lowercase I think to match other error code cases.

Lyokone: I've updated with `IllegalArgumentException` to match Android :)

packages/cloud_functions/cloud_functions/ios/Classes/FLTFirebaseFunctionsPlugin.m
```diff
@@ -93,7 +94,17 @@ - (void)httpsFunctionCall:(id)arguments withMethodCallResult:(FLTFirebaseMethodC
...
 
-  FIRHTTPSCallable *function = [functions HTTPSCallableWithName:functionName];
+  FIRHTTPSCallable *function;
+
+  if (![functionName isEqual:[NSNull null]]) {
+    function = [functions HTTPSCallableWithName:functionName];
+  } else if (![functionUri isEqual:[NSNull null]]) {
+    function = [functions HTTPSCallableWithURL:[NSURL URLWithString:functionUri]];
+  } else {
+    result.error(@"IllegalArgumentException", @"Either functionName or functionUri must be set",
```
russellwheatley, line 104: Still not 100% sure on this. Shouldn't we capture on android side the illegal argument exception and have an error code in the same format as other exception codes? (e.g. something like: `illegal-argument-exception`) 🤔 . Then we can match it here with the same exception code.

russellwheatley (changes requested):

packages/cloud_functions/cloud_functions/ios/Classes/FLTFirebaseFunctionsPlugin.m
Lyokone: Huum, for an internal error like this one which does not depend on input from the user (since the user can only provide either name or uri) I'm not sure if it's worth it (already caught in assert on the dart side)
It was more to prevent having a pending result. This error was more for future developers to see the case properly.

russellwheatley: That's true actually, LGTM then 👍

russellwheatley (approved):

packages/cloud_functions/cloud_functions/ios/Classes/FLTFirebaseFunctionsPlugin.m
Salakar: So my understanding is this should never happen since dart code prevents it?

packages/cloud_functions/cloud_functions/ios/Classes/FLTFirebaseFunctionsPlugin.m
Salakar: Can this code actually happen though, shouldn't Dart side block it?

Lyokone: This code shouldn't happened, it was done to handle a case if we do a mistake while programming (this could have been a else instead of a else if)

packages/cloud_functions/cloud_functions/ios/Classes/FLTFirebaseFunctionsPlugin.m
Lyokone: Yes exactly 👍

lesnitsky (approved):

packages/cloud_functions/cloud_functions/lib/src/firebase_functions.dart
```diff
@@ -60,12 +60,43 @@ class FirebaseFunctions extends FirebasePluginPlatform {
   String? _origin;
 
   /// A reference to the Callable HTTPS trigger with the given name.
-  HttpsCallable httpsCallable(String name, {HttpsCallableOptions? options}) {
+  ///
+  /// Should be the name of the Callable function in Firebase
+  /// or the URL of the 2nd gen Callable function in Firebase.
```
Salakar, line 65: This comment is incorrect @Lyokone

Lyokone: Fixed in https://github.com/firebase/flutterfire/pull/10642

PR #10680, opened 2023-03-28
chore: add an OSSF audit scorecard workflow

jhuleatt:
## Description

Add an OSSF audit to check security compliance to show a scorecard for the security of the repo.


jhuleatt:
The `all_plugins / pub_dry_run` check failed here, but it seems to be failing on `master` too, so we're going to merge with that failure.

FYI @Salakar @Lyokone

kroikie (approved):

puf (approved):

Issue #10167, opened 2022-12-22
🐛 [firebase_app_check] `getToken` may return an invalid App Check token on Android


rosalyntan:
## Bug report

**Describe the bug**
In the Android wrapper for the Flutter `getToken` method, I noticed it [calls through to Android's `getToken` method](https://github.com/firebase/flutterfire/blob/5abac2bae5da88d488daefb68d22a930f3d51eb5/packages/firebase_app_check/firebase_app_check/android/src/main/java/io/flutter/plugins/firebase/appcheck/FlutterFirebaseAppCheckPlugin.java#L129). This method was intended for internal interop use with other Firebase SDKs (i.e. Firestore) and as such, returns a [placeholder token](https://github.com/firebase/firebase-android-sdk/blob/350aef6f39695479a5cd40528f24780d31c3e339/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultAppCheckTokenResult.java#L42-L45) if the App Check token exchange fails. However, the expected behavior for developers is for the exception to be propagated if the App Check token exchange fails.

Instead of calling Android's `getToken` method, the [`getAppCheckToken` method](https://github.com/firebase/firebase-android-sdk/blob/350aef6f39695479a5cd40528f24780d31c3e339/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/FirebaseAppCheck.java#L80), which is the public API method for obtaining an App Check token and propagates any exceptions, should be called instead.

darshankawar:
Thanks for the report @rosalyntan

kinex:
Is this probably related to https://github.com/firebase/flutterfire/issues/10084? In my production application, I also see a strange amount of invalid requests (3-5% consistently).

blinghartig:
this is base64 encoded and means `{"error":"UNKNOWN_ERROR"}` i have the same issue with ^0.1.1+8

yamauchieduardo:
Same issue here. I think this have relation with sms authentication. Are you using this method?

PR #10819, opened 2023-04-19
fix(app-check): use correct `getAppCheckToken()` method. Print out debug token for iOS.

russellwheatley:
## Description

A few changes in this PR

1. We use correct method for calling app check token as mentioned [here](https://github.com/firebase/flutterfire/issues/10167).
2. Updated method channel `getToken()` method to return the token as a string. We previously returned a map with the token unnecessarily. 
3. I realised the plugin was no longer printing out the debug token because we override the `createProviderWithApp()` method so we can use a facade app check instance. I have created a log to print out the token and slightly updated the docs to reflect the change.

## Related Issues

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/docs/app-check/debug-provider.md
+++ b/docs/app-check/debug-provider.md
@@ -54,7 +54,7 @@ Future<void> main() async {
     Your app will print a local debug token to the debug output when Firebase
     tries to send a request to the backend. For example:
 
-    <pre>[Firebase/AppCheck][I-FAA001001] Firebase App Check Debug Token:
+    <pre>Firebase App Check Debug Token:
     123a4567-b89c-12d3-e456-789012345678</pre>
 
 {# Google-internal common file: #}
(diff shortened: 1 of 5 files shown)


russellwheatley:
Hey @kevinthecheung, I've slightly updated the documentation for app check as the log in the console is slightly different. I think we also need to think about updating the website at some point. It is still displaying the old way of implementing the debug provider which requires changing code in Xcode when that is no longer the case. Thanks!

Lyokone (approved):

kevinthecheung (approved):

Lyokone:
@russellwheatley CI fail seems to be related to this change

russellwheatley:
@Lyokone using the different android API has an affect on the android `getToken()` e2e test. The iOS failure seems to be unrelated to the PR. Will investigate.

russellwheatley:
Now that we're using the correct API, the e2e test for `getToken()` won't work. I've skipped it like we skip for iOS, macOS & web. I tested locally, once I added the debug token printed into the terminal and updated the Firebase console, I was able to retrieve the App Check token on android.

PR #10957, opened 2023-05-12
fix(firebase_ui_auth): Error added in ProfileScreen in firebase_ui_auth 1.4.0

zachary-russell:
## Description

firebase_ui_auth throws an error when building for version 1.4.0.:
`../../.pub-cache/hosted/pub.dev/firebase_ui_auth-1.4.0/lib/src/screens/profile_screen.dart:393:19: Error: Cannot invoke a non-'const' constructor where a const expression is expected. Try using a constructor or factory that is 'const'. const Row( ^^^`

## Related Issues

[Issue](https://github.com/firebase/flutterfire/issues/10954)

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process. Updating the `pubspec.yaml` and changelogs is not required.

- [ x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ x] All existing and new tests are passing.
- [x ] I updated/added relevant documentation (doc comments with `///`).
- [x ] The analyzer (`melos run analyze`) does not report any problems on my PR.
- [ x] I read and followed the [Flutter Style Guide].
- [ x] I signed the [CLA].
- [ x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change.
- [ x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/firebase/flutterfire/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #10963, opened 2023-05-14
fix(ui): removed const usage for Row for compatibility before 3.10

Heaust-ops:
## Description

The current code triggers an error due to using a non-constant constructor in a const context. Removing the const keyword from the Row widget resolves this error, allowing the code to compile successfully.


packages/firebase_ui_auth/lib/src/screens/profile_screen.dart
```diff
@@ -390,7 +390,7 @@ class _EmailVerificationBadgeState extends State<_EmailVerificationBadge> {
           ),
           const SizedBox(height: 16),
           if (state == EmailVerificationState.pending)
-            const Row(
+            Row(
               mainAxisAlignment: MainAxisAlignment.center,
               children: [
```
Lyokone, line 395:
```suggestion
              children: const [
```

Lyokone (approved):

lesnitsky:
#10957

PR #11016, opened 2023-05-21
docs: Update example main.dart

erlangparasu:
because previous example causing error, the declaration of the function is positional but the example use named parameter.


russellwheatley:
It's a named argument.

Issue #10966, opened 2023-05-15
🐛 [firebase_auth] signInWithEmailAndPassword returns error code "unknown"


jayjay911:
## Bug report

**Describe the bug**
I am using Flutter Web and running in Chrome.

After upgrading to Flutter 3.10.0 and all my firebase dependencies I started getting unusual behaviour when attempting to sign into firebase with email and password. 

Instead of throwing a FirebaseAuthMultiFactorException as expected, it throws a FirebaseAuthException with the following message:

An unknown error occurred: FirebaseError: Firebase: Proof of ownership of a second factor is required to complete sign-in. (auth/multi-factor-auth-required).

Error code is:

"unknown"

Error code is always "unknown" even in cases where email/password are incorrect.

Issue persists in a Flutter Demo Project with only firebase dependencies

### Steps to reproduce

1. Create a Flutter Demo Project

2. Add necessary code to initialise FirebaseApp

3. Add the following to  _incrementCounter() :

```
try {
      await FirebaseAuth.instance
          .signInWithEmailAndPassword(email: "email@domain.com", password: "password");
    } on FirebaseAuthMultiFactorException catch (e) {
      print(e.message);
      print(e.code);
      rethrow;
    } on FirebaseAuthException catch (e) {
      print(e.message);
      print(e.code);
      if (e.code == 'user-not-found') {
        print('The account does not exist');
      }
      if (e.code == 'second-factor-required') {
        print('Two factor challenge must be completed first');
      }
      if (e.code == 'multi-factor-auth-required') {
        print('Two factor challenge must be completed first');
      }
    }
```

4. Add necessary dependencies to pubspec.yaml and call pub.get

### Expected behavior

I am expecting a FirebaseAuthMultiFactorException where email and password are correct, otherwise a FirebaseAuthException with a code as before.

I need the FirebaseAuthMultiFactorException obviously so I can access the resolver, and I need the code in order to handle incorrect password, user does not exist etc cases.

## Additional context

Running Flutter Web version in Chrome on M1 MacBook

---

### Flutter doctor

Run `flutter doctor` and paste the output below:

<details><summary>Click To Expand</summary>

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.10.0, on macOS 13.3.1 22E772610a darwin-arm64, locale en-ZA)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[...]

Issue #11099, opened 2023-06-07
Refactor FlutterFire web implementation by removing `is` checks on web instances.


russellwheatley:
As per the recommendation by the Flutter SDK team, we need to remove all `is` checks on web instances and use `@staticInterop` in the interop layer.

## Example 

update this:
```dart
  if (exception is core_interop.FirebaseError) {
    return _firebaseExceptionFromCoreFirebaseError(
      exception,
      plugin: plugin,
      codeParser: codeParser,
      messageParser: messageParser,
    );
  }
```

To this:
```dart
  if (js_util.instanceof(exception, core_interop.FirebaseError)) {
    return _firebaseExceptionFromCoreFirebaseError(
      exception,
      plugin: plugin,
      codeParser: codeParser,
      messageParser: messageParser,
    );
  }
```
This would also require changing `core_interop.FirebaseError` to a `@staticInterop`.

### To Update
- [ ] Await this issue to be resolved first: https://github.com/dart-lang/sdk/issues/52572
- [ ] Refactor the above noted.
- [ ] Remove the test skips on web for Firestore here: https://github.com/firebase/flutterfire/pull/11092/files

sigmundch:
Note that the `instanceof` call takes as an argument a constructor, but Dart doesn't give you that when you refer to the dart class. So likely it is not:
```
 if (js_util.instanceof(exception, core_interop.FirebaseError)) {
    return _firebaseExceptionFromCoreFirebaseError(
      exception,
      plugin: plugin,
      codeParser: codeParser,
      messageParser: messageParser,
    );
  }
```

but more something like:
```
 if (js_util.instanceof(exception, core_interop.firebaseErrorJSConstructor)) {
    return _firebaseExceptionFromCoreFirebaseError(
      exception,
      plugin: plugin,
      codeParser: codeParser,
      messageParser: messageParser,
    );
  }
```

where `firebaseErrorJSConstructor` is a getter that returns the JavaScript `FirebaseError` type.

russellwheatley:
Closing out as we have now implemented the latest `js_interop`

Issue #10993, opened 2023-05-18
🐛 [firebase_auth] WEB: Bad state: Cannot fire new event. Controller is already firing an event


just7mile:
## Bug report

Dart: **3.0.0**
Flutter: **3.10.0**
Firebase Core: **2.13.0**
Firebase Auth: **4.6.1**

**Describe the bug**
After upgrading to Flutter 3.10 the Firebase Auth on web is giving `Bad state: Cannot fire new event. Controller is already firing an event` error during hot reload/restart. I tried the fix mentioned in this [comment](https://github.com/firebase/flutterfire/issues/3756#issuecomment-757500560), but it did not work.

<img width="703" alt="image" src="https://github.com/firebase/flutterfire/assets/29413456/13f2433f-21ff-47a8-9d49-a8bd265af1b2">

And the VSCode points to the file `firebase_auth_web-5.5.1/lib/src/interop/auth.dart`:

<img width="1059" alt="image" src="https://github.com/firebase/flutterfire/assets/29413456/e1051564-086c-49f8-a340-e3a11dfdb6e8">

darshankawar:
Thanks for the report @just7mile 
I tried to replicate it using [plugin example](https://github.com/firebase/flutterfire/tree/master/packages/firebase_auth/firebase_auth/example) running on latest stable with dart 3 environment and latest plugin versions, but I was able to login properly with a test user without throwing the reported error.

<img width="1193" alt="Screenshot 2023-05-18 at 5 32 48 PM" src="https://github.com/firebase/flutterfire/assets/67046386/7cf40e88-ebcc-4ac8-81dd-2627db0c02d1">

You can try the same and see if it still gives you same error or not. If not, a repo or minimal code sample would be helpful to debug this further.

just7mile:
Thank you for you response! I couldn't reproduce it with plugin's example app, but I was able to find the source of the problem. **It turns out the problem was related to Cloud Firestore, with the catching `FirebaseException` specifically**. 

When using Firestore emulators and enabling persistence it gives `Firestore has already been started and persistence can no longer be enabled. You can only enable persistence before calling any other methods on a Firestore object.` error on hot restart/reload, so I was catching this error and ignoring it if it's a `failed-precondition` error:
```dart
try {
    if (kIsWeb && persistenceEnabled) {
      await FirebaseFirestore.instance
          .enablePersistence(const PersistenceSettings(synchronizeTabs: true));
    }
  } on FirebaseException catch (e) {
[...]

PR #11258, opened 2023-07-11
fix(auth, web): convert `NativeError` to `FirebaseAuthError`

russellwheatley:
## Description

I've changed a few things which was inspired by @sigmundch's PR:https://github.com/sigmundch/flutterfire/compare/master...additional_fixes 

### Changes

1. We differentiate the AuthError from FirebaseError using the `customData` property that is on the former and not on the latter. See https://firebase.google.com/docs/reference/js/auth.autherror
2. Created an interop class for `customData`.
3. I tested on latest stable version of Flutter and Flutter `3.7.3` and received the same result. It is backwards compatible.

I used this code sample to check against the Auth example app:

```dart
try {
                    await FirebaseAuth.instance.signInWithEmailAndPassword(
                        email: 'russell@invertase.io', password: '------');
                  } on FirebaseAuthException catch (e) {
                    print('EEEEE: $e');
                    if (e.code == 'user-not-found') {
                      print('The account does not exist');
                    }
                    if (e.code == 'second-factor-required') {
                      print('Two factor challenge must be completed first');
                    }
                    if (e.code == 'multi-factor-auth-required') {
                      print('Two factor challenge must be completed first');
                    }

                    if (e.code == 'invalid-login-credentials') {
                      print('invalid-login-credentials');
                    }
                  }
```

When I used the wrong password, I got the correct output in the console:
Screenshot 2023-07-18 at 12 45 33

I also tried to log in with correct password whilst expecting multi-factor auth. I got the correct output:

Screenshot 2023-07-18 at 12 44 35

Please note the [additional print](https://github.com/firebase/flutterfire/pull/11258/commits/f2dc5a7acd6c04c795ca844e7d5559ac139ffe3d#diff-a522fde29aeb319e6149a657c49c90dcce5c910e0e1166a1b37a0f00b4360885L92) for `MultiFactorResolverWeb `.  I added an additional print to check changes to `auth_interop.MultiFactorError` worked on web.

# FOR CONTEXT - The below noted was the initial PR that is no longer relevant but keeping for understanding the issue

Created this PR to demonstrate that there is no way of capturing `FirebaseError` anymore.

I have amended [this function](https://github.com/firebase/flutterfire/compare/fix-web-exceptions-11099?expand=1#diff-a522fde29aeb319e6149a657c49c90dcce5c910e0e1166a1b37a0f00b4360885R26) to capture the `FirebaseError` as mentioned [in this comment](https://github.com/firebase/flutterfire/issues/11099#issuecomment-1581146404). As you can see below, it is skipped:
Screenshot 2023-07-11 at 11 59 07

Inspecting the `exception` object, you can see it is an instance of `NativeError`:
Screenshot 2023-07-11 at 11 59 35

I have also inspected the `core_interop.firebaseErrorJSConstructor` which shows that is an instance of `FirebaseError`:
Screenshot 2023-07-11 at 12 01 58

This stops us from ever capturing the `FirebaseError` exceptions when using Firebase in web applications. It is not just Auth but Firestore and presumably every plugin.

I have also amended the Auth example app to make it easy to receive an exception if you wish to test for yourself.

## Related Issues

related to: 
https://github.com/firebase/flutterfire/issues/11099
https://github.com/dart-lang/sdk/issues/52572
closes https://github.com/firebase/flutterfire/issues/10966
https://github.com/firebase/flutterfire/issues/10993

## Checklist

- [ ] I signed the [CLA].

## Breaking Change


packages/firebase_auth/firebase_auth_web/lib/src/interop/auth_interop.dart
```diff
@@ -553,6 +557,7 @@ abstract class AuthError {
   external set tenantId(String s);
   external String get phoneNumber;
   external set phoneNumber(String s);
+  external dynamic get customData;
```
sigmundch, line 560: minor nit: similarly here, consider using `Object` or `Object?` if you expect it to be nullable.

It is not ready yet, but in an upcoming release you'll be able to use a `JSObject` type instead, which will make it clear that this is not an arbitrary Dart object.

packages/firebase_auth/firebase_auth_web/lib/src/utils/web_utils.dart
```diff
@@ -22,20 +37,36 @@ FirebaseAuthException getFirebaseAuthException(
...
       message: 'An unknown error occurred: $exception',
     );
   }
 
-  auth_interop.AuthError firebaseError = exception as auth_interop.AuthError;
-
-  String code = firebaseError.code.replaceFirst('auth/', '');
-  String message = firebaseError.message
-      .replaceFirst(' (${firebaseError.code}).', '')
+  String code = getProperty(exception, 'code').replaceFirst('auth/', '');
```
sigmundch, line 47: Consider keeping the old pattern that casts to AuthError and then uses regular properties in Dart code. Casts to static interop types like AuthError will not break due to changes in DDC/dart2js (unlike the cast to the non-static interop types).

The reason behind this recommendation is to leverage the declared API and avoid subtle issues with `js_util`. Technically there are no issues here, but we generally recommend limiting the use of the js_util methods, like `getProperty`, because the extension declarations are safer and less error prone.

Above we could technically avoid the `getProperty(globalThis, 'Error'))` by writing instead:
```dart
@JS('Error')
external Object get errorConstructor;

bool _isirebaseAuthError(Object e) {
  instanceof(e, errorConstructor) &&
  hasProperty(e, 'customData')
}
```

packages/firebase_auth/firebase_auth_web/lib/src/utils/web_utils.dart
```diff
@@ -14,6 +15,20 @@ import 'package:firebase_core_web/firebase_core_web_interop.dart'
...
+/// Workaround test to check whether `e` is practically a FirebaseError.
+///
+/// Ideally we'd check whether `e instanceof FirebaseError` however
+/// there are two definitions of `FirebaseError` in deployed apps, one from
+/// `firebase-auth.js` (which is usually minified) and one from
+/// `firebase-app.js`.
+///
+/// Because they are not the same, the instanceof check fails. Instead, we test
+/// that it is a window.Error (since that's the base class of FirebaseError) and
+/// we check that it defines "customData" property which is on the AuthError but not the FirebaseError it extends
```
sigmundch, line 27: suggestion: I know these are all approximations, but given that the code below access `code` and `message` before it accesses `customData`, I wonder if it is safer to check for the code first, or check for both?

Another idea is to try to extract the code here, check that it starts with `auth/` as part of the validation, and return the code if it is available. For example:
```dart
String? _readFirebaseAuthErrorCode(Object o) {
  if (!instanceof(e, errorConstructor)) return null;
  String code = getProperty(e, 'code');
  if (code == null || !code.startsWith('auth/')) return null;
  return code.substring(5);
}

...

final code = _readFirebaseAuthErrorCode(exception);
if (code == null) {
  return FirebaseAuthException(code: 'unknown', ...);
}
...
```

sigmundch (commented):
Glad to see it all working!

I took the opportunity to share a few thoughts and comments below.

packages/firebase_auth/firebase_auth_web/lib/src/utils/web_utils.dart
russellwheatley: Thanks for the suggestion, I have reverted the cast back to as it was, I didn't expect it to work but work it did.

I've kept the `_isFirebaseAuthError(exception)` function as it is, your suggestion caused the following exception:
```
TypeError: Right-hand side of 'instanceof' is not an object
dart-sdk/lib/_internal/js_shared/lib/js_util_patch.dart 231:35       $instanceof
packages/firebase_auth_web/src/utils/web_utils.dart 29:50            _isFirebaseAuthError
```

sigmundch: Thanks!

> I have reverted the cast back to as it was, I didn't expect it to work but work it did.

Glad to hear :) - yes, with static-interop, it all works better. This is about to improve further with new efforts in improving how we write interop code.

> I've kept the _isFirebaseAuthError(exception) function as it is, your suggestion caused the following exception

Sounds good to keep as you did. I'm curious about the error, though. I didn't see it when I tried locally, so there may be a small difference in what we both did.

One possible reason is where we both placed the definition of `errorConstructor`. If by chance you added it in a library like `auth_interop.dart`, then because the library has a `@JS` annotation itself, it changes the meaning of what the getter does. In a library with no namespace, the external member reads `window.Error`, but in a library like auth_interop, it would read `window.firebase_auth.Error`.

Could that be why?

packages/firebase_auth/firebase_auth_web/lib/src/utils/web_utils.dart
```diff
@@ -22,20 +49,23 @@ FirebaseAuthException getFirebaseAuthException(
...
   }
 
   auth_interop.AuthError firebaseError = exception as auth_interop.AuthError;
-
   String code = firebaseError.code.replaceFirst('auth/', '');
   String message = firebaseError.message
       .replaceFirst(' (${firebaseError.code}).', '')
       .replaceFirst('Firebase: ', '');
 
+  // customData is a Map<String, String>, see Firebase docs: https://firebase.google.com/docs/reference/js/auth.autherror
```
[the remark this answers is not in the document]
russellwheatley: updated code comment to point to docs 👍

russellwheatley:
Thanks for suggestion, @sigmundch! I've created a separate file for window to grab the Error constructor and now the `_isFirebaseAuthError` function uses the `window_interop.errorConstructor` to check the error. Works as intended 😄

Lyokone (approved):

exaby73 (approved):

anjogithub:
if i may ask, how can i apply this fix? will this fix be included in the next updates of firebase_auth?

Issue #11364, opened 2023-07-24
[cloud_firestore_odm_generator] Support querying enum fields


Rexios80:
There doesn't seem to be a way to create an issue for a new feature sorry

PR #11361, opened 2023-07-24
feat(cloud_firestore_odm_generator): Support Sets as well as Lists for query method generation

Rexios80:
## Description

Support generating query methods for Sets.

## Related Issues

#11344

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process. Updating the `pubspec.yaml` and changelogs is not required.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`melos run analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change.
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/firebase/flutterfire/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #11365, opened 2023-07-24
feat(cloud_firestore_odm_generator): Add support for querying enum fields

Rexios80:
## Description

Support querying enum fields

## Related Issues

#11364

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
+++ b/packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
@@ -10,6 +10,7 @@ import 'package:collection/collection.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:recase/recase.dart';
 import 'package:source_gen/source_gen.dart';
+import 'package:source_helper/source_helper.dart';
 
 import 'collection_generator.dart';
 import 'names.dart';
@@ -348,6 +349,7 @@ represents the content of the collection must be in the same file.
         type.isDartCoreBool ||
         type.isPrimitiveList ||
         type.isJsonDocumentReference ||
+        type.isEnum ||
         dateTimeChecker.isAssignableFromType(type) ||
         timestampChecker.isAssignableFromType(type) ||
         geoPointChecker.isAssignableFromType(type);
--- a/packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/templates/query_reference.dart
+++ b/packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/templates/query_reference.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/element/nullability_suffix.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/dart/element/type_provider.dart';
+import 'package:source_helper/source_helper.dart';
 
 import '../collection_data.dart';
 
@@ -380,7 +381,19 @@ class ${data.queryReferenceImplName}
               ? '${field.type}'
               : '${field.type}?';
 
-      final operators = {
+      String parameterMapping(String type, String parameter) {
+        if (field.type.isEnum && type.contains(field.type.toString())) {
+          if (type.contains('List')) {
+            return '$parameter?.map((e) => _\$${field.type}EnumMap[e]!).toList()';
+          } else {
+            return '_\$${field.type}EnumMap[$parameter]!';
+          }
+        } else {
+          return parameter;
+        }
+      }
+
+      final operators = <String, String>{
         'isEqualTo': nullableType,
         'isNotEqualTo': nullableType,
         'isLessThan': nullableType,
@@ -390,7 +403,8 @@ class ${data.queryReferenceImplName}
         'isNull': 'bool?',
         if (field.type.isDartCoreList) ...{
           'arrayContains': data.libraryElement.typeProvider
-              .asNullable((field.type as InterfaceType).typeArguments.first),
+              .asNullable((field.type as InterfaceType).typeArguments.first)
+              .toString(),
           'arrayContainsAny': nullableType,
         } else ...{
           'whereIn': 'List<${field.type}>?',
@@ -401,7 +415,9 @@ class ${data.queryReferenceImplName}
       final prototype =
           operators.entries.map((e) => '${e.value} ${e.key},').join();
 
-      final parameters = operators.keys.map((e) => '$e: $e,').join();
+      final parameters = operators.entries
+          .map((e) => '${e.key}: ${parameterMapping(e.value, e.key)},')
+          .join();
 
       // TODO support whereX(isEqual: null);
       // TODO handle JsonSerializable case change and JsonKey(name: ...)
--- a/packages/cloud_firestore_odm/cloud_firestore_odm_generator/pubspec.yaml
+++ b/packages/cloud_firestore_odm/cloud_firestore_odm_generator/pubspec.yaml
@@ -20,6 +20,7 @@ dependencies:
   meta: ^1.8.0
   recase: ^4.0.0
   source_gen: ^1.3.2
+  source_helper: ^1.3.4
 
 dev_dependencies:
   build_runner: ^2.4.2


Rexios80:
We should merge #11361 in first

rrousselGit:
:wave:

This needs testing.

packages/cloud_firestore_odm/cloud_firestore_odm_generator/cloud_firestore_odm_generator_integration_test/lib/simple.dart
```diff
@@ -35,6 +35,12 @@ class Model {
   final String value;
 }
 
+enum TestEnum {
+  one,
```
rrousselGit, line 39: We'd need a test which executes the a where/orderBy query and see if they work.
We'll also need an update operation.
And we need to check reads too

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
```diff
@@ -348,12 +350,37 @@ represents the content of the collection must be in the same file.
...
+          final mapped =
+              '_\$${type.getDisplayString(withNullability: false)}EnumMap[$name]!';
+          if (nullable) {
+            return '$name == null ? null : $mapped';
+          } else {
+            return mapped;
+          }
+        };
+
+    if (type.isDartCoreList &&
```
rrousselGit, line 372: Why do we have list-specific logic?

Rexios80: Because lists of enums need mapped to lists of strings

Rexios80: How do you suggest we generate this without relying on list-specific logic?

Screenshot 2023-07-31 at 9 22 33 AM

Rexios80: Maybe we could call the toJson method first and then pipe the resulting parameter in? That seems kind of jank though.

rrousselGit: Folks using enums should use json_serializable on their enums. Meaning they use things like `@JsonValue` & stuff, or maybe have a from/toJson on their enum class.


In fact, this _should_ work today. Let me double check

rrousselGit: Right, I got confused for a bit.
What needs updating is different. We need to rely on _$TestEnumPerFieldToJson, by asking users to enable `@JsonSeriaizable(createPerFieldToJson: true)`

That's something I added in json-serializable for this purpose https://github.com/google/json_serializable.dart/issues/972

rrousselGit: TL;DR the `update` should serialize fields as followed:

```dart
        _$EnumCollectionFieldMap['myField']!: _$EnumCollectionPerFieldToJson.myField(myField as MyFieldType),
```

This should add support for enums, but also `List<Enum>` or `Map<Enum, Whatever>` or whatever other funky combination you could think of.

Rexios80:
@rrousselGit Let me know what you think of the new way of doing this

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
```diff
@@ -336,6 +344,22 @@ represents the content of the collection must be in the same file.
...
+  static void _assertValidJsonAnnotation(
+    DartType type, {
+    required bool hasPerFieldToJson,
+  }) {
+    final isEnumField = type.isEnum;
+    final hasEnumTypeArg = type.isDartCoreIterable &&
+        (type as InterfaceType).typeArguments.any((e) => e.isEnum);
+
+    if ((isEnumField || hasEnumTypeArg) && !hasPerFieldToJson) {
+      throw InvalidGenerationSourceError(
```
rrousselGit, line 356: I like the idea! But this is unlikely to work when the flag is set globally in a `build.yaml`

Because in that case, the annotation has `null` for value

packages/cloud_firestore_odm/cloud_firestore_odm_generator/cloud_firestore_odm_generator_integration_test/lib/simple.dart
```diff
@@ -35,7 +35,13 @@ class Model {
...
 }
 
-@JsonSerializable()
+enum TestEnum {
+  one,
+  two,
+  three;
+}
+
+@JsonSerializable(createPerFieldToJson: true)
```
rrousselGit, line 44: Could you remove the flag and update the `build.yaml` instead?

packages/cloud_firestore_odm/doc/defining-models.md
```diff
@@ -37,6 +37,7 @@ const firestoreSerializable = JsonSerializable(
   // The following values could alternatively be set inside your `build.yaml`
   explicitToJson: true,
   createFieldMap: true,
+  createPerFieldToJson: true,
```
rrousselGit, line 40: Great, thanks!

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
```diff
@@ -10,6 +10,7 @@ import 'package:collection/collection.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:recase/recase.dart';
 import 'package:source_gen/source_gen.dart';
+import 'package:source_helper/source_helper.dart';
```
rrousselGit, line 13: I'm not quite sure: What is this new dependency used for?

rrousselGit:
This is looking great! Some small test changes and we should be good to go :)

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
rrousselGit: To make such an error, we'd likely need https://github.com/dart-lang/build/issues/3557

Rexios80: This field is also used to actually generate the code, so if it's an issue here it's also an issue elsewhere

packages/cloud_firestore_odm/cloud_firestore_odm_generator/cloud_firestore_odm_generator_integration_test/lib/simple.dart
Rexios80: Not until we fix the issue with getting the perFieldToJson value

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
Rexios80: This gives us `DartType.isEnum`

packages/cloud_firestore_odm/cloud_firestore_odm/example/integration_test/common.dart
```diff
@@ -44,5 +45,21 @@ Movie createMovie({
   );
 }
 
+Enums createEnums({
```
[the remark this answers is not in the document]
Rexios80: I'm not sure what you mean. I just emulated the `createMovie` function above for the tests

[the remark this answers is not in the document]
rrousselGit: createMovie exists because some tests want the fields to be required (such as to test if they are missing)

We don't care about this for enum tests.

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
Rexios80: Assuming this is set causes issues with the `ManualJson` and `PublicRedirected` test classes

Rexios80:
So I actually tried generating this in my project and there's an issue with inherited classes.

```
class A {}

@firestoreSerializable
class B extends A {}

@firestoreSerializable
class C extends A {}

@Collection<A>('path')
final aRef = ACollectionReference();
```

This generates code that tries to use `_$APerFieldToJson` which obviously doesn't exist. I want to just fix this in my code by creating my own `_$APerFieldToJson`, but I was wondering if you had any better ideas.

Rexios80:
Maybe creating my own perFieldToJson isn't a great idea since all the methods are static 🙃

Rexios80:
Unless we could change the way the PerFieldToJson code is generated so that you can call do something like:

```dart
abstract class A {
  final TypeEnum type;

  Function perFieldToJson(String field) => switch (type) {
        TypeEnum.a => _$APerFieldToJsonMap[field],
        TypeEnum.b => _$BPerFieldToJsonMap[field],
      };
}
```

rrousselGit:
> This generates code that tries to use _$APerFieldToJson which obviously doesn't exist.

We can check for the presence of JsonSerializable/Freezed. That should be enough

But honestly I'm fine requiring that all Collection models are using JsonSerializable.

Rexios80:
Alright. What do you want me to do with the `ManualJson` tests then? Just remove them?

Rexios80:
Also now that I think about it... do we even need to check for type support anymore? If `json_serializable` successfully generates for a class then we should have a `PerFieldToJson` call for every serialized field in the class right? This would allow us to automatically support any type `json_serializable` inherently supports (right now I'm thinking the `Duration` type specifically, but I'm sure there are others). This would also give us support for anything using a `JsonConverter`.

rrousselGit:
Yes that's the goal :)
I didn't want to manually handle types on a case-by-case basis, hence why I made that PR to json_serializable

rrousselGit:
Do you plan on fixing the CI?

Rexios80:
Yeah I'll fix it. Do you want me to turn this PR into the one where we support all types or do that after? It's a simple change.

rrousselGit:
Let's support all types :)

packages/cloud_firestore_odm/cloud_firestore_odm_generator/test/document_reference_test.dart
```diff
@@ -13,21 +13,18 @@ Future<void> main() async {
   );
 
   group('where(arrayContains)', () {
```
rrousselGit, line 15: Why is this file changed?

packages/cloud_firestore_odm/cloud_firestore_odm/example/lib/integration.dart
```diff
@@ -22,6 +22,7 @@ class EmptyModel {
 final emptyModelRef = EmptyModelCollectionReference();
 
 @Collection<ManualJson>('root')
+@JsonSerializable()
```
rrousselGit, line 25: That defeats the point of this test. This test is about _not_ using json_serializable

packages/cloud_firestore_odm/cloud_firestore_odm_generator/cloud_firestore_odm_generator_integration_test/lib/model.dart
```diff
@@ -5,6 +5,11 @@
 // When separated accross multiple files, it is necessary to specify fromJson/toJson
 // We voluntarily don't use JsonSerializable here, as it is not supported due to
 // the generated FieldMap being private.
+import 'package:json_annotation/json_annotation.dart';
+
+part 'model.g.dart';
+
+@JsonSerializable()
```
rrousselGit, line 12: This is inconsistent with the comments a few lines above.

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
```diff
@@ -146,6 +138,16 @@ class CollectionData with Names {
       },
     ).toList();
 
+    final hasJsonSerializable =
+        jsonSerializableChecker.hasAnnotationOf(collectionTargetElement);
+    // Freezed classes are also JsonSerializable
+    if (!hasJsonSerializable && !hasFreezed) {
```
rrousselGit, line 144: Using Freezed or json-serializable wasn't required before. This is a breaking change

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
```diff
@@ -157,41 +159,11 @@ class CollectionData with Names {
...
-    // Whether the model class and the reference variable are defined in the same file
-    // This is important because json_serializable generates private code for
-    // decoding a Model class.
-    final modelAndReferenceInTheSameLibrary =
-        collectionTargetElement.librarySource == annotatedElementSource;
-
-    final fromJson = collectionTargetElement.constructors.firstWhereOrNull(
-      (ctor) => ctor.name == 'fromJson',
-    );
-    if (hasJsonSerializable && !modelAndReferenceInTheSameLibrary) {
```
rrousselGit, line 172: Why delete all of this?

Rexios80: Since we require every class to have JsonSerializable on it I was trying to clean this up. At least some of this should probably stay though.

Rexios80:
CI is failing but I want to talk about the changes required before fixing it

packages/cloud_firestore_odm/cloud_firestore_odm_generator/lib/src/collection_data.dart
rrousselGit: No we shouldn't require it.

Rexios80: Did you say here you were fine requiring it or am I misreading?

https://github.com/firebase/flutterfire/pull/11365#issuecomment-1681427822

rrousselGit: We require the flag if json_serializable is used. But if it isn't used, we don't.

Rexios80: What are the merits for using custom json serialization? Is there anything that json_serializable can't handle with converters? The only other option I see as viable is have people implement their own `perFieldToJson` methods if they don't want to use json_serializable. Otherwise we would have to maintain two different code-gen options. One that uses json_serializable with perFieldToJson that supports all types and one without that doesn't support all types. The second option does not sound great to me. What are your thoughts?

rrousselGit: I think that's fair. We can add support for it back if necessary later.

rrousselGit:
LGTM once CI is fixed

packages/cloud_firestore_odm/cloud_firestore_odm_generator/cloud_firestore_odm_generator_integration_test/lib/model.dart
rrousselGit: The test is useless if we don't support not using json_serializable

packages/cloud_firestore_odm/cloud_firestore_odm/example/lib/integration.dart
rrousselGit: Yes

packages/cloud_firestore_odm/cloud_firestore_odm_generator/cloud_firestore_odm_generator_integration_test/lib/model.dart
Rexios80: @rrousselGit Do you want me to remove this file?

Rexios80:
@rrousselGit how is this looking?

rrousselGit:
Sorry I missed your notification. Looking good!

rrousselGit (approved):
LGTM

exaby73 (approved):
LGTM

Rexios80:
What's the plan for releasing this?

gmarizy:
First, I would like to thanks @Rexios80 and @rrousselGit for the work. While non-blocking, having a bunch `.whereFieldPath(FieldPath.fromString('fieldName'), isEqualTo: MyEnum.myValue.name)` partly defeated the type safe promise from ODM IMHO. It was my main pain point regarding ODM, and you solved it. Thank you you two.

But I think you missed something, since models won't generate until you set `create_per_field_to_json: true` with version 1.0.0-dev.70. If the breaking change is necessary, changelog and documentation should be updated. If it's the case, I started a PR: https://github.com/firebase/flutterfire/pull/11728/

Issue #10969, opened 2023-05-15
🐛 [firebase_database] convertFirebaseDatabaseException


ublanke:
## Bug report

**Describe the bug**
In function *convertFirebaseDatabaseException* a null exception is thrown, as the field message does not exist. 
`
FirebaseException convertFirebaseDatabaseException(Object exception,
      [StackTrace? stackTrace]) {
        String code = 'unknown';
       String message = util.getProperty(exception, 'message');
       [..]
`

Object exception contains:
0: "code" -> "PERMISSION_DENIED"
key: "code"
value: "PERMISSION_DENIED"

It does **not** contain "message". 

### Steps to reproduce

Access to a Realtime Database resource for which permissions arent set.

### Expected behavior
Output with "Permission Denied"

---

## Additional context

---

### Flutter doctor

Run `flutter doctor` and paste the output below:

<details><summary>Click To Expand</summary>

Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.7.9, on macOS 13.0.1 22A400 darwin-x64, locale en-CH)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0-rc4)
[✓] Xcode - develop for iOS and macOS (Xcode 14.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] VS Code (version 1.78.0)
[✓] Connected device (3 available)
[✓] HTTP Host Availability
</details>

---

### Flutter dependencies

Run `flutter pub deps -- --style=compact` and paste the output below:

<details><summary>Click To Expand</summary>
art SDK 2.19.6
Flutter SDK 3.7.9
dashboard 1.0.0+1

dependencies:
- awesome_dialog 3.0.2 [flutter rive]
- cloud_firestore 4.7.0 [cloud_firestore_platform_interface cloud_firestore_web collection firebase_core firebase_core_platform_interface flutter meta]
- cupertino_icons 1.0.5
- dropdown_search 4.0.1 [flutter]
- firebase_app_check 0.1.4 [firebase_app_check_platform_interface firebase_app_check_web firebase_core firebase_core_platform_interface flutter]
- firebase_auth 4.6.0 [firebase_auth_platform_interface firebase_auth_web firebase_core firebase_core_platform_interface flutter meta]
- firebase_core 2.12.0 [firebase_core_platform_interface firebase_core_web flutter meta]
[...]

PR #11522, opened 2023-08-28
fix(database): Fix convertFirebaseDatabaseException failing if no message is provided

smith897:
## Description

convertFirebaseDatabaseException throws a null exception if the exception does not contain a message. This fixes the issue by allowing the retrieved message to be null.

This null exception is caught before it reaches the caller, so the caller is never notified their subscription is no longer active. If there's a .onError callback attached, it is never called.

## Related Issues

https://github.com/firebase/flutterfire/issues/10969
This issue describes the problem, but was closed due to inactivity.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,70 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2023-08-23
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
+Packages with other changes:
+
+ - [`cloud_firestore` - `v4.9.0`](#cloud_firestore---v490)
+ - [`cloud_firestore_platform_interface` - `v5.16.0`](#cloud_firestore_platform_interface---v5160)
+ - [`cloud_firestore_web` - `v3.7.0`](#cloud_firestore_web---v370)
+ - [`firebase_auth` - `v4.8.0`](#firebase_auth---v480)
+ - [`firebase_auth_platform_interface` - `v6.17.0`](#firebase_auth_platform_interface---v6170)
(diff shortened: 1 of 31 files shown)


smith897:
@russellwheatley  Sure, I was able to recreate it by just calling this from the root of my application
`FirebaseDatabase.instance.ref("test").onValue.listen((event) {
  print("New event: ${event.toString()}");
});`

I created the "test" entry through the firebase console. Once I was listening, I went into the console and changed the "reads" rule in the rules to false.

Basically, it can be recreated by listening to something, and then revoking access to it.

smith897:
I guess I should also mention, as this happens in convertFirebaseDatabaseException, it must be done in Flutter web.

russellwheatley:
I followed up on this and I was able to recreate. It seems odd that no message was provided so I checked on a pure web implementation and it seems correct that a message is not provided:
Screenshot 2023-08-31 at 16 13 04

smith897:
@russellwheatley Thank you! I'm surprised that Android end to end tests are preventing a merge, since the only change was in the web implementation. Do you know if that's something I need to look into?

russellwheatley:
@smith897 - the android CI is somewhat flaky, don't worry about that. I'll ensure it goes green 👍

russellwheatley (approved):

exaby73:
@smith897 Can you add a test for this? I believe this is only needed for the web platform. You can update a `ref` in a query to one that would through a permission denied error and check if the exception is generated correctly. Other than this, it LGTM

russellwheatley:
> @smith897 Can you add a test for this? I believe this is only needed for the web platform. You can update a `ref` in a query to one that would through a permission denied error and check if the exception is generated correctly. Other than this, it LGTM

Hey @smith897 - are you able to write an e2e test as mentioned above? Thanks

PR #11795, opened 2023-10-30
feat(firebase_analytics, ios): add onDeviceConversionMeasurement

lucasuracosta:
## Description

This PR adds the onDeviceConversionMeasurement function for ios [which is needed for an increased volume and precision of conversion tracking](https://support.google.com/google-ads/answer/12119136?hl=en). 

I followed as example @mikehardy's [implementation of the feature in React Native](https://github.com/invertase/react-native-firebase/commit/f41666f8b126a87bbbd91a1a5a30dd7ca8e0a4f2) and @Lyokone's [implementation of Pigeon](https://github.com/firebase/flutterfire/pull/8921).

## Checklist

- [x] I signed the [CLA].

## Breaking Change


mikehardy:
Just dropping by for a moment since I saw my tag here 👋  - wanted to note that there is a phone API as well with firebase-ios-sdk 10.13.0+

https://github.com/invertase/react-native-firebase/pull/7358

Same basic pattern, might make sense to get them both in at once

packages/firebase_analytics/firebase_analytics/example/lib/main.dart
```diff
@@ -87,6 +88,23 @@ class _MyHomePageState extends State<MyHomePage> {
     }
   }
 
+  Future<void> initiateOnDeviceConversionMeasurementWithEmailAddress() async {
+    if (!Platform.isIOS) {
+      setMessage(
+        '"setDefaultEventParameters()" is only supported on iOs',
```
Lyokone, line 94:
```suggestion
        '"setDefaultEventParameters()" is only supported on iOS',
```

packages/firebase_analytics/firebase_analytics/ios/Classes/FLTFirebaseAnalyticsPlugin.m
```diff
@@ -156,6 +160,27 @@ - (void)setDefaultEventParameters:(id)arguments
...
+                                         enabled:(nonnull NSNumber *)enabled
+                                      completion:
+                                          (nonnull void (^)(FlutterError *_Nullable))completion {
+  FIRApp *firebaseApp = [FLTFirebasePlugin firebaseAppNamed:appName];
+  if (firebaseApp) {
+    [firebaseApp setDataCollectionDefaultEnabled:[enabled boolValue]];
+  }
+
+  completion(nil);
+}
```
Lyokone, line 182: I'm not seeing where this method is being called. You probably need to add something in the `handleMethodCall`

packages/firebase_analytics/firebase_analytics_platform_interface/pubspec.yaml
```diff
@@ -20,3 +20,4 @@ dev_dependencies:
   firebase_core_platform_interface: ^5.0.0
   flutter_test:
     sdk: flutter
+  pigeon: ^11.0.1
```
Lyokone, line 23:
```suggestion
  pigeon: 11.0.1
```
We lock the version to prevent having breaking change in other packages since we are generating the pigeons messages in different packages

Lyokone (commented):
Some comments, but looks really good, thanks for this :)

packages/firebase_analytics/firebase_analytics/example/lib/main.dart
lucasuracosta: Fixed ✅

packages/firebase_analytics/firebase_analytics_platform_interface/pubspec.yaml
lucasuracosta: Fixed ✅

packages/firebase_analytics/firebase_analytics/ios/Classes/FLTFirebaseAnalyticsPlugin.m
lucasuracosta: @Lyokone I thought the same at first glance, but actually it's being called inside the `FirebaseAnalyticsHostApiSetup`

firebase_analytics_messages.g.m
```
void FirebaseAnalyticsHostApiSetup(id<FlutterBinaryMessenger> binaryMessenger,
                                   NSObject<FirebaseAnalyticsHostApi> *api) {
  {
    FlutterBasicMessageChannel *channel = [[FlutterBasicMessageChannel alloc]
           initWithName:
               @"dev.flutter.pigeon.firebase_analytics_platform_interface.FirebaseAnalyticsHostApi."
               @"initiateOnDeviceConversionMeasurementWithEmailAddress"
        binaryMessenger:binaryMessenger
                  codec:FirebaseAnalyticsHostApiGetCodec()];
    if (api) {
      NSCAssert([api respondsToSelector:@selector
                     (initiateOnDeviceConversionMeasurementWithEmailAddressEmailAddress:
                                                                             completion:)],
                @"FirebaseAnalyticsHostApi api (%@) doesn't respond to "
                @"@selector(initiateOnDeviceConversionMeasurementWithEmailAddressEmailAddress:"
                @"completion:)",
                api);
      [channel setMessageHandler:^(id _Nullable message, FlutterReply callback) {
        printf("Inside newMessageHandler");
        NSArray *args = message;
        NSString *arg_emailAddress = GetNullableObjectAtIndex(args, 0);
        [api initiateOnDeviceConversionMeasurementWithEmailAddressEmailAddress:arg_emailAddress
                                                                    completion:^(
                                                                        FlutterError
                                                                            *_Nullable error) {
                                                                      callback(
                                                                          wrapResult(nil, error));
                                                                    }];
      }];
    } else {
      [channel setMessageHandler:nil];
    }
  }
}
```

It receives the binary messenger and sets a parallel channel, which then calls to the appropriate function. If you run the example, you'll see that the function is properly executed and (if you add the print I wrote in this example) the print too.

In short, there's no need to call it in `handleMethodCall`, Pigeon already does it.

The ideal would be to have all the functions of firebase_analytics migrated into pigeon, then all of them would be called in the same place. But I think this would be overkill for this scenario and it would transform this non-breaking PR into a breaking one.

lucasuracosta:
@Lyokone All your recommendations were implemented and I [left a comment on your concern about its functioning](https://github.com/firebase/flutterfire/pull/11795#discussion_r1402070023).

All things considered we should be good to go, right?

P.S: I'll eventually add the equivalent of this implementation for phone numbers in a separate PR but I can't promise it will be soon, so I would rather go only with this one for now.

mikehardy:
> P.S: I'll eventually add the equivalent of this implementation for phone numbers in a separate PR but I can't promise it will be soon, so I would rather go only with this one for now.

Sure - did not mean to imply any sort of pressure, just wanted to mention it since whenever this PR makes it through review, using whatever code looks good, the other API should be doing *exactly* the same thing, so it's potentially a really really quick hit that re-uses all the kind-of-ephemeral wisdom built up on how to do the PR and review it etc.

But no one has endless time of course, I certainly feel that! Cheers

lucasuracosta:
Ohhh I get what you meant now, thanks!!!

lucasuracosta:
@Lyokone Just checking back on this one, is there anything missing for it to be ready? 😄

Lyokone:
@lucasuracosta Thanks for your contribution, we ended going another way with the API (`initiateOnDeviceConversionMeasurementWithEmailAddress`) and it's available in the Analytics plugin now :)

PR #12331, opened 2024-02-15
test(analytics): fix broken `getSessionId()` test that requires async API

russellwheatley:
## Description

## Related Issues

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_analytics/firebase_analytics_web/lib/firebase_analytics_web.dart
+++ b/packages/firebase_analytics/firebase_analytics_web/lib/firebase_analytics_web.dart
@@ -44,8 +44,7 @@ class FirebaseAnalyticsWeb extends FirebaseAnalyticsPlatform {
   }
 
   @override
-  Future<int?> getSessionId() {
-    // TODO: change UnimplementedError to UnsupportedError
+  Future<int?> getSessionId() async {
     throw UnimplementedError('getSessionId() is not supported on Web.');
   }
 


packages/firebase_analytics/firebase_analytics_web/lib/firebase_analytics_web.dart
```diff
@@ -44,8 +44,7 @@ class FirebaseAnalyticsWeb extends FirebaseAnalyticsPlatform {
   }
 
   @override
-  Future<int?> getSessionId() {
-    // TODO: change UnimplementedError to UnsupportedError
```
rrousselGit, line 48: Sounds like this would warrant a regression test

russellwheatley: @rrousselGit - I don't think it is necessary. The API is a Future as far as the user is concerned: https://github.com/firebase/flutterfire/blob/master/packages/firebase_analytics/firebase_analytics/lib/src/firebase_analytics.dart#L80

This change fixes that bug. It also resolves to an UnimplementedError on web so it will not be used by any users. The fact that no user has pointed out this incorrect API infers that it isn't being used which makes sense when it does nothing for their web app.

rrousselGit: This is a behaviour change though (throwing synchronously vs asynchronously). If the behaviour changed, a test helps prevent inadvertently reverting back to the old behaviour. So I think it makes sense

russellwheatley: "a test helps prevent inadvertently reverting back to the old behaviour" - The test now works (It has never worked but the web e2e tests were always "successful" even when they failed). Now web e2e tests have been fixed, it will fail if behaviour reverts back to previous.

rrousselGit: Oh so it fixes an existing test. Nevermind then

rrousselGit (approved):

packages/firebase_analytics/firebase_analytics_web/lib/firebase_analytics_web.dart
russellwheatley: Sorry, I didn't make that clear, it fixes this test: https://github.com/firebase/flutterfire/blob/master/tests/integration_test/firebase_analytics/firebase_analytics_e2e_test.dart#L28 😄

PR #12593, opened 2024-04-05
fix(firestore): deprecate `databaseURL` in favor of `databaseId`

russellwheatley:
## Description

- Updated user facing code to allow both `databaseURL` &  `databaseId`. `databaseId` takes precedence over `databaseURL`  if both are input. Otherwise, they have the same value (if one or the other are selected).
- Refactored method channel, platform and web implementations to use `databaseId` naming.
- Wrote unit test.
- We can update pigeon and native code to use `databaseId` naming in a future PR, perhaps when `databaseURL` is removed.

## Related Issues

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
+++ b/packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
@@ -16,9 +16,17 @@ part of cloud_firestore;
 /// FirebaseFirestore firestore = FirebaseFirestore.instanceFor(app: secondaryApp);
 /// ```
 class FirebaseFirestore extends FirebasePluginPlatform {
-  FirebaseFirestore._({required this.app, required this.databaseURL})
-      : super(app.name, 'plugins.flutter.io/firebase_firestore') {
+  FirebaseFirestore._({
+    required this.app,
+    @Deprecated(
+      '`databaseURL` has been deprecated. Please use `databaseId` instead.',
+    )
+    required this.databaseURL,
+    required this.databaseId,
+  }) : super(app.name, 'plugins.flutter.io/firebase_firestore') {
+    // ignore: deprecated_member_use_from_same_package
     if (databaseURL.endsWith('/')) {
+      // ignore: deprecated_member_use_from_same_package
       databaseURL = databaseURL.substring(0, databaseURL.length - 1);
     }
   }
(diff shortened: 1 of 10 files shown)


Lyokone (approved):

packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
```diff
@@ -35,16 +37,25 @@ class FirebaseFirestore extends FirebasePluginPlatform {
...
     if (_cachedInstances.containsKey(cacheKey)) {
       return _cachedInstances[cacheKey]!;
     }
 
     FirebaseFirestore newInstance =
-        FirebaseFirestore._(app: app, databaseURL: url);
+        // Both databaseURL and databaseId are required so we have to pass both for now. We can remove databaseURL in a future release.
+        FirebaseFirestore._(
+      app: app,
+      databaseURL: firestoreDatabaseId,
```
exaby73, line 56: Couldn't we remove it here since it's a private constructor, and won't be a breaking change for the user?

packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
```diff
@@ -16,12 +16,14 @@ part of cloud_firestore;
...
-    }
-  }
+  FirebaseFirestore._({
+    required this.app,
+    @Deprecated(
+      '`databaseURL` has been deprecated. Please use `databaseId` instead.',
+    )
+    required this.databaseURL,
+    required this.databaseId,
+  }) : super(app.name, 'plugins.flutter.io/firebase_firestore');
```
exaby73, line 26: What happened to the following. Is it not needed anymore?

```dart
if (databaseURL.endsWith('/')) {
  databaseURL = databaseURL.substring(0, databaseURL.length - 1);
}
```

exaby73 (approved):
LGTM; just a couple of questions

packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
russellwheatley: no, it isn't a URL, it is a databaseId so it will be like this:
https://github.com/firebase/flutterfire/pull/12593/files#diff-69186946971119c262bbc33818abecdc9ae077adb87cdfd1c0a489787d6a7d44R40

The default databaseId is "(default)".

packages/cloud_firestore/cloud_firestore/lib/src/firestore.dart
russellwheatley: We could but we would then have to initialise it in a constructor function. It is one or the other so it doesn't make a difference.

PR #12662, opened 2024-04-17
chore: add a BoM compatibility table to repository

Lyokone:
## Description

This PR allows binding a set of Firebase plugin versions into a single parent version named a BoM (Bill of Materials) version - an overarching version that internally specifies all versions of the FlutterFire packages that work together for that BoM release (much like [Android](https://developer.android.com/develop/ui/compose/bom)).

This PR adds a Melos script `bom` which updates 2 files:
- `versions.json` which is a JSON file that can be consumed externally (e.g. the FlutterFire CLI) to get a match between a BoM version and all the other Flutter plugin versions and Firebase SDK versions
- `VERSIONS.md` which is a readable display version of the same information

The contributing guide has been updated with instructions on how to run a BoM release. A BoM release can be run at any time. We use Melos (`melos bom`) to automatically suggest the next BoM version number (based on the types of package version changes between the last BoM version and now) but it can also be overridden manually (`melos bom [customVersion]`).

A typical BoM would look like:

```shell
guillaume@macbook $ melos bom
New BoM version number (0.2.0): 0.2.0
JSON version data has been successfully written to scripts/versions.json
Version 0.2.0 has been generated successfully!
```

## Why do we need this?

To solve customer pain points in regards to using compatible versions of the FlutterFire set of plugins together, e.g. this [tweet](https://twitter.com/DamianKBast/status/1780568163324731839).

Some alternatives have been considered in the past:

- Pub has no concept of peer dependencies, so we can't specify optional dependencies and their versions and have pub enforce that you use the right versions together, see https://github.com/dart-lang/pub/issues/1837
- We could make all packages have the same version numbering but you'd get versions of packages getting bumped even with no changes just to keep them the same version number, getting us quickly to the max 1000 versions of a package limit pub has and also getting us to the pub rate limiting of publishing too many packages at once.
- We could make a new package that depends on all the FlutterFire packages with the correct exact versions specified and you just use that package instead, and, while that would work there's the issue that your Dart code would get tree shaken to remove code for FlutterFire packages you don't actually use but the native parts of the FlutterFire plugins won't since they all get built and included in your app by flutter regardless of whether you use them in Dart or not because of how the plugin system in Flutter works, so you end up with every single Firebase service included in your native build.

Recently however we've been in a good position where the FlutterFire CLI has become the defacto way to do a bunch of things with FlutterFire and so we're able to explore solving problems like this with the CLI instead.


version.md
```diff
@@ -0,0 +1,111 @@
...
+
+## [Flutter BoM 1.1.0 (2024-04-17)](https://github.com/firebase/flutterfire/blob/master/CHANGELOG.md#2024-04-16)
+
+Install this version using FlutterFire CLI
+
+```bash
+flutterfire install 1.1.0
+```
+
+### Included Native SDK Versions
```
Salakar, line 50:
```suggestion
### Included Native Firebase SDK Versions
```

version.md
```diff
@@ -0,0 +1,111 @@
...
+
+| Native SDK | Version | Link |
```
Salakar, line 52:
```suggestion
| Firebase SDK | Version | Link |
```

version.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Install this version using FlutterFire CLI
+
+```bash
+flutterfire install 1.1.0
+```
+
+### Included Native Firebase SDK Versions
+
+| Firebase SDK | Version | Link |
```
puf, line 52: Should we include this information in the JSON blurb at the top of the page too?

Lyokone: I can add this to the JSON as well, in case it needs to be parsed by something else. Since we were not going to use it directly for the install command, didn't feel it was necessary, but can always be useful 🚀

exaby73 (approved):

TarekkMA (approved):

versions.mdx
```diff
@@ -0,0 +1,67 @@
...
+        "firebase_app_check": "0.2.2+2",
+        "firebase_ml_model_downloader": "0.2.5+2",
+        "native_sdk": {
+            "android": "32.8.0",
+            "ios": "10.24.0",
+            "web": "10.10.0",
+            "windows": "11.6.0"
+        },
+    }
+}
```
russellwheatley, line 24: We ought to put this at the bottom of the file. Also - this is still displayed on preview mode when we should probably hide it:
Screenshot 2024-04-18 at 10 13 28

also - a trailing comma here which would break parsing I think:
```json
        "native_sdk": {
            "android": "32.8.0",
            "ios": "10.24.0",
            "web": "10.10.0",
            "windows": "11.6.0"
        },
```

russellwheatley (changes requested):
Nice, just a few things from me 👍

versions.mdx
Lyokone: I'm quite keen on keeping it at the top since it's part of MDX spec as [Frontmatter](https://mdxjs.com/guides/frontmatter/), at the bottom it's just random things, as you can see if I add it to the bottom:
Screenshot 2024-04-18 at 11 38 54

Lyokone: I've fixed the trailing comma

Salakar:
Based on chats internally my feedback would be:

- can we write the script in Dart, we already have a root pubspec, no extra files or setup needed other than the script file, others not familiar with bash can pick up making changes safely
- rename `versions.mdx` to `VERSIONS.md` to match naming convention of other root MD files
- move versions out to separate json file, perhaps next to the script, e.g. `./scripts/versions.json`
- detach BoM versioning away from main release process, allow us to run normal releases regularly and group a bunch of releases into a single BoM at a different schedule to regular releases
- BoM versions auto determined based on diff of versions of packages in last BoM vs versions of packages now to be used in new BoM, e.g. if any packages had a breaking change then BoM would be major, same with minor & patch

scripts/generate_bom.dart
```diff
@@ -0,0 +1,201 @@
...
+    iosVersionFile,
+  );
+  String webSdkVersion = await getSdkVersion(
+    "awk -F \"'\" '/const String supportedFirebaseJsSdkVersion =/ {print \$2}' \$VERSION_FILE",
+    webVersionFile,
+  );
+  String windowsSdkVersion = await getSdkVersion(
+    "awk -F '\"' '/set\\(FIREBASE_SDK_VERSION/ {print \$2}' \$VERSION_FILE",
+    windowsVersionFile,
+  );
```
Salakar, line 61: we can just use file read and simple regexes to do this, 

Something like this should work (untested):

```dart
String androidPattern = r'FirebaseSDKVersion\s*=\s*\'([^\']+)\'';
String iosPattern = r'def firebase_sdk_version!()\s*\'([^\']+)\'';
String webPattern = r'const String supportedFirebaseJsSdkVersion = \'([^\']+)\'';
String windowsPattern = r'FIREBASE_SDK_VERSION\s*\"([^\"]+)\"';

Future<String> getSdkVersion(String pattern, String versionFile) async {
  RegExp regex = RegExp(pattern);
  String fileContents = await File(versionFile).readAsString();
  Match? match = regex.firstMatch(fileContents);
  return match?.group(1)?.trim() ?? "Version not found";
}

// ...
// Fetch native versions
String androidSdkVersion = await getSdkVersion(
  androidPattern,
  androidVersionFile,
);
String iosSdkVersion = await getSdkVersion(
  iosPattern,
  iosVersionFile,
);
String webSdkVersion = await getSdkVersion(
  webPattern,
  webVersionFile,
);
String windowsSdkVersion = await getSdkVersion(
  windowsPattern,
  windowsVersionFile,
);
```

scripts/versions.json
```diff
@@ -0,0 +1,24 @@
...
+        "firebase_storage": "11.7.2",
+        "firebase_database": "10.5.2",
+        "firebase_messaging": "14.8.2",
+        "firebase_crashlytics": "3.5.2",
+        "firebase_performance": "0.9.4+2",
+        "firebase_remote_config": "4.4.2",
+        "firebase_analytics": "10.10.2",
+        "firebase_in_app_messaging": "0.7.5+2",
+        "firebase_app_check": "0.2.2+2",
+        "firebase_ml_model_downloader": "0.2.5+2"
```
Salakar, line 44: nit: can we nest this under another key like `packages`?

scripts/versions.json
```diff
@@ -0,0 +1,24 @@
+{
+    "0.1.0": {
+        "date": "2024-04-16",
+        "native_sdk": {
```
Salakar, line 26:
```suggestion
        "firebase_sdk": {
```

nit: can we call this firebase_sdk?

Lyokone: Updated :)

scripts/generate_bom.dart
```diff
@@ -0,0 +1,227 @@
...
+  'firebase_remote_config',
+  'firebase_analytics',
+  'firebase_dynamic_links',
+  'firebase_in_app_messaging',
+  'firebase_app_check',
+  'firebase_app_installations',
+  'firebase_ml_model_downloader',
+];
+
+const jsonEncoder = JsonEncoder.withIndent('    ');
```
Salakar, line 43:
```suggestion
const jsonEncoder = JsonEncoder.withIndent('  ');
```

nit: can we do 2 spaces for indentation like other source files in the repo

version.mdx
Lyokone: It has been added

CONTRIBUTING.md
```diff
@@ -304,11 +304,24 @@ Some things to keep in mind before publishing the release:
 5. Push your branch to git running `git push origin [RELEASE BRANCH NAME]`.
 6. Run `melos version` to automatically version packages and update Changelogs.
 7. Run `melos publish` to dry run and confirm all packages are publishable.
-8. Run `git push origin [RELEASE BRANCH NAME]` & open pull request for review on GitHub.
-9. After successful review and merge of the pull request, switch to `master` branch locally, & run `git pull origin master`.
-10. Run `melos publish --no-dry-run` to now publish to Pub.dev.
-11. Run `git push --tags` to push tags to repository.
-12. Ping @kevinthecheung to get the changelog in Firebase releases.
+8. Run `melos bom [optionnal-version]` to update the `VERSIONS.md` and `scripts/versions.json` files.
```
Salakar, line 307:
```suggestion
8. Run `melos bom [optional-version]` to update the `VERSIONS.md` and `scripts/versions.json` files.
```

CONTRIBUTING.md
```diff
@@ -304,11 +304,24 @@ Some things to keep in mind before publishing the release:
...
+12. Run `git push --tags` to push tags to repository.
+13. Ping @kevinthecheung to get the changelog in Firebase releases.
+
+### Run a BoM release only...
+
+1. Switch to `master` branch locally.
+2. Run `git pull origin master`.
+3. Run `git pull --tags` to make sure all tags are fetched.
+4. Create new branch with the signature "release/[year]-[month]-[day]-BoM".
+5. Run `melos bom [optionnal-version]` to update the `VERSIONS.md` and `scripts/versions.json` files.
```
Salakar, line 320:
```suggestion
5. Run `melos bom [optional-version]` to update the `VERSIONS.md` and `scripts/versions.json` files.
```

scripts/versions.json
```diff
@@ -0,0 +1,29 @@
+{
```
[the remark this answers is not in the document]
cynthiajoan: nvm found the script writing the file. Would like to confirm for the script, when new release published, is the json content being replaced, or append the new version block?

TarekkMA (approved):

russellwheatley (approved):

PR #12693, opened 2024-04-23
Fix more unsafe casts to `int`.

eyebrowsoffire:
Taking values from JSON and directly casting them to `int` or `double` makes them incompatible with dart2wasm, since in wasm they are two distinct types (as opposed to JavaScript where they are actually the same under the hood).


eyebrowsoffire:
I think these were addressed in another PR

PR #12719, opened 2024-04-29
Concept for instanceFor function

Attempt3035:
## Description

I've had issues with occasionally on particular emulators (usually android) where trying to instance crashlytics results in no default firebase being found, as I don't use the default, I've created a named firebase instance as:

```
final mainFirebase = await Firebase.initializeApp(
    name: 'master',
    options: DefaultFirebaseOptions.currentPlatform,
);
```

Shouldn't there be an instanceFor method in crashlytics? My changes in this PR are incomplete and likely not functional, but I don't understand why every other firebase system has instanceFor functionality but crashlytics does not...

Note: I have no experience in the backend of these firebase services, hence this is conceptual and will need advice and clarification from someone with experience :)

### DIFF
--- a/packages/firebase_crashlytics/firebase_crashlytics/lib/src/firebase_crashlytics.dart
+++ b/packages/firebase_crashlytics/firebase_crashlytics/lib/src/firebase_crashlytics.dart
@@ -34,6 +34,12 @@ class FirebaseCrashlytics extends FirebasePluginPlatform {
     return _instance!;
   }
 
+  /// Returns an instance using a specified [FirebaseApp].
+  static FirebaseCrashlytics instanceFor({required FirebaseApp app}) {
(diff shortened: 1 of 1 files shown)


russellwheatley:
@Attempt3035 - Firebase Crashlytics can only have a default instance for android and apple. Example, see [iOS docs](https://firebase.google.com/docs/reference/ios/firebasecrashlytics/api/reference/Classes/FIRCrashlytics?_gl=1*1w8fpt3*_up*MQ..*_ga*MjA4MzQ3OTk1OC4xNzE0NDc1MDk5*_ga_CW55HF8NVT*MTcxNDQ3NTA5OC4xLjAuMTcxNDQ3NTA5OC4wLjAuMA..#+crashlytics). Hence why FlutterFire only allows creating a default instance.

Issue #12711, opened 2024-04-26
[FIREBASE_STORAGE]:  Wrong json key for customMetadata


ciceromsjr:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Which plugins are affected?

Storage

### Which platforms are affected?

Windows

### Description

I'm having an issue retrieving metadata from a file on Windows. The file has metadata, but when I fetch this data, the 'customMetadata' attribute comes back null. Inspecting the returned object, it's possible to see that it exists but under a different key in the mapping: 'metadata'.

I also noticed that the custom metadata I send when uploading a file isn't being sent. It might be related to this same issue.

I'm using firebase_storage version 11.7.4."

![image](https://github.com/firebase/flutterfire/assets/3678693/5989eee0-6e1a-48d2-bb7b-b2e48d051946)

### Reproducing the issue

Using the firebase_storage plugin version 11.7.4 on Windows:

Retrieve the metadata of a file that has custom metadata.

The customMetadata object comes back null, but inspecting the FullMetadata object, it exists under the key "metadata".

### Firebase Core version

2.30.1

### Flutter Version

3.19.6

### Flutter dependencies

<details>
<summary>Expand <code>Flutter dependencies</code> snippet</summary>
<br>

```yaml
- app_tracking_transparency 2.0.4 [flutter]
- archive 3.4.10 [crypto path pointycastle]
- background_downloader 8.4.3 [flutter logging http path_provider path async mime collection]
- cached_network_image 3.3.1 [cached_network_image_platform_interface cached_network_image_web flutter flutter_cache_manager octo_image]
- cloud_firestore 4.17.2 [cloud_firestore_platform_interface cloud_firestore_web collection firebase_core firebase_core_platform_interface flutter meta]
- collection 1.18.0
- community_material_icon 5.9.55 [flutter]
- connectivity_checker 1.1.0 [flutter provider]
- crypto 3.0.3 [typed_data]
- csv 5.1.1
- cupertino_icons 1.0.6
- currency_picker 2.0.21 [flutter collection]
- d_chart 2.6.9 [flutter community_charts_flutter community_charts_common]
- data_table_2 2.5.12 [flutter async]
- desktop_disk_space 0.0.1 [flutter plugin_platform_interface]
- device_info_plus 9.1.2 [device_info_plus_platform_interface ffi file flutter flutter_web_plugins meta win32 win32_registry]
[...]

PR #12723, opened 2024-04-29
fix(storage, windows): putFile(), putString(), putData() & Task streaming event fixes

russellwheatley:
## Fixes
1. `putData()` is now writing the data as intended.
2. `putString()` is now decoding from base64 and base64Url before writing string (previously writing string as base64).
3. `putFile()` metadata was not working correctly.
4.  `putFile()` we have also had to do a [hack to pass metadata](https://github.com/firebase/flutterfire/pull/12723/files#diff-5be57535e40bc3dc68ed0a4706153ddc03710b570c24066b9178441e7ea591ccR260-R264) as `null` on windows, it was throwing exception in pigeon even though it is [allowable as a nullable value](https://github.com/firebase/flutterfire/blob/master/packages/firebase_storage/firebase_storage/windows/messages.g.cpp#L1256-L1259). Not [ideal, but it works](https://github.com/firebase/flutterfire/pull/12723/files#diff-cd14e729c2a123769d24deb04c095c7e5c520143b9fa4508a63797ebe890fd85R360-R364).
5. `customMetadata` is now returned with the snapshot which fixes: https://github.com/firebase/flutterfire/issues/12711
6. Some responses were coming back completely incorrect, they have been fixed by using `EncodableValue()` to pass values back to Flutter.
7. task!.snapshotEvents events were never firing. The storage event listener (onProgress, and onPause) were incorrectly passed to the API handler (e.g. `putBytes()`, `putFile()`, etc). This is the [correct way to instantiate](https://github.com/firebase/flutterfire/pull/12723/files#diff-cd14e729c2a123769d24deb04c095c7e5c520143b9fa4508a63797ebe890fd85R650).
8. `task.pause()`, `task.cancel()`, `task.resume()` were also broken. The incorrect encoding of values passed back to Flutter stopped the flutter side from having proper values to parse.
9. The `error` map was never passed back in the stream event handling so exception were not being thrown correctly when listening to events.
10. Fixed an issue in Firebase core C++ were `AppOptions` (equivalent to `FirebaseOptions` in Flutter side) was creating every property as an empty string. For example, storage bucket option was being passed back to Flutter as an empty string.
11. `snapshot` was missing `path` property causing exception on Flutter side when handling File streaming events.
12. `metadata` as a property on `snapshot` was also missing `bucket` property.

## Tests
1. Storage windows now has integration tests albeit using a live Firebase project. There are some tests that are skipped which are largely stream handling events, I spent quite a bit of time on this. It works when running the app but fails in a test environment. The good news is most tests are being used.
2. I updated the `putData()`, `putFile()` and `putString()` tests to also check the content is correct and not gibberish or base64 encoding.
3. I've found a bug in the C++ SDK, if you try to update metadata on an object that does not exist, it will throw an unauthorized exception rather than an object-not-found exception. I checked native, it is a firebase SDK bug.

## Related Issues

fixes https://github.com/firebase/flutterfire/issues/12743
fixes https://github.com/firebase/flutterfire/issues/12711
## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_storage/firebase_storage_platform_interface/lib/src/method_channel/method_channel_task.dart
+++ b/packages/firebase_storage/firebase_storage_platform_interface/lib/src/method_channel/method_channel_task.dart
@@ -7,6 +7,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:firebase_core/firebase_core.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 
 import '../../firebase_storage_platform_interface.dart';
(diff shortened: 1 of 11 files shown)


packages/firebase_storage/firebase_storage_platform_interface/lib/src/method_channel/method_channel_task.dart
```diff
@@ -257,14 +257,22 @@ class MethodChannelPutFileTask extends MethodChannelTask {
 
   static Future<String> _getTask(int handle, FirebaseStoragePlatform storage,
       String path, File file, SettableMetadata? metadata) {
+    PigeonSettableMetadata? pigeonSettableMetadata;
+    if (defaultTargetPlatform == TargetPlatform.windows) {
+      // TODO(russellwheatley): sending null to windows throws exception so we pass empty metadata
+      pigeonSettableMetadata =
+          MethodChannelFirebaseStorage.getPigeonSettableMetaData(metadata);
```
Lyokone, line 264: Nice, probably an issue in Pigeon, but the workaround looks OK

TarekkMA (approved):

PR #12794, opened 2024-05-17
refactor(vertexai): Split into separate libraries

natebosch:
The current approach uses `part` files which cannot be imported
separately. One drawback of this approach is that any members which need
to be referenced by test code must be public to the library. Using
`src/` libraries, as opposed to `src/` part files, allows making APIs
that are "package private" by convention - they are public to the
library, but only exported through the `src/` import.

One potential downside (alternatively seen as a benefit) of using
separate libraries is that private members are truly private to their
library, and we cannot make a member of a class private to the package.
The convention for package imports only allows top level declarations to
be considered non-public implementation details.

- Remove imports from the top level library. Replace `part` lines with
  `export` of the same files.
- Add explicit `show` clauses to the exports. This is what allows an API
  to be public when imported from the `src/` library, but not exposed
  through the package public library.
- Change some private `_toGoogleAI<TypeName>` methods to extension
  methods. Use the name `toGoogleAI` since the specific type names are
  clear from context.
- Change some private `_fromGoogleAI<TypeName>` factory constructors to
  extension methods on the google AI SDk types. Name `toVertex()`.
- Remove some unused private methods that were filled in for
  consistency. More extension members can be added as needed.
- Add an extension to expose the private field `_googleAiModel` from
  `GenerativeModel`.
- Add a top level method to forward to the private `GenerativeModel`
  constructor.
- Add TODO comments to stop exporting some methods that don't need to be
  public. These could have been private in the `part` pattern. For now
  they are exported for backwards compatibility, and we can remove them
  when we are ready for a major version bump.

### DIFF
--- a/packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
+++ b/packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
@@ -228,10 +228,13 @@ final class GenerativeModel {
   }
 }
 
+/// Conversion utilities for [GenerativeModel].
 extension GoogleAIGenerativeModelConversion on GenerativeModel {
+  /// Return this model as a [google_ai.GenerativeModel].
   google_ai.GenerativeModel get googleAIModel => _googleAIModel;
 }
 
+/// Returns a [GenerativeModel] using it's private constructor.
 GenerativeModel createGenerativeModel({
   required FirebaseApp app,
   required String location,
--- a/packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
+++ b/packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
@@ -14,7 +14,18 @@
 
 // ignore_for_file: use_late_for_private_fields_and_variables
 
-part of firebase_vertexai;
+import 'dart:async';
+
+import 'package:firebase_app_check/firebase_app_check.dart';
+import 'package:firebase_core/firebase_core.dart';
+import 'package:google_generative_ai/google_generative_ai.dart' as google_ai;
+// ignore: implementation_imports, tightly coupled packages
+import 'package:google_generative_ai/src/vertex_hooks.dart';
+
+import 'vertex_api.dart';
+import 'vertex_content.dart';
+import 'vertex_function_calling.dart';
+import 'vertex_version.dart';
 
 const _baseUrl = 'firebaseml.googleapis.com';
 const _apiVersion = 'v2beta';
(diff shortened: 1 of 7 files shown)


natebosch:
If you have any branches that don't merge cleanly with this I can hold off and land it whenever it's convenient.

This is a refactor I would recommend, but it's not urgent. Let me know if you have any questions about the reasoning or implications of this change.

If you want to go forward with this refactoring I can work on jumping through the CI hurdles.

packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
```diff
@@ -225,9 +223,34 @@ final class GenerativeModel {
...
-        .batchEmbedContents(
-            requests.map((e) => e._toGoogleAIEmbedContentRequest()))
-        .then(
-            BatchEmbedContentsResponse._fromGoogleAIBatchEmbedContentsResponse);
+        .batchEmbedContents(requests.map((e) => e.toGoogleAI()))
+        .then((r) => r.toVertex());
   }
 }
+
+extension GoogleAIGenerativeModelModel on GenerativeModel {
```
cynthiajoan, line 231: Is ModelModel a typo?

cynthiajoan:
> If you have any branches that don't merge cleanly with this I can hold off and land it whenever it's convenient.
> 
> This is a refactor I would recommend, but it's not urgent. Let me know if you have any questions about the reasoning or implications of this change.
> 
> If you want to go forward with this refactoring I can work on jumping through the CI hurdles.

This is great, thanks for doing this! I'd prefer you to go ahead and get this in first. It's much cleaner.

packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
natebosch: Yes it was.

natebosch:
I've added some API docs to these members for now. I'd recommend dropping the lint which requires this - it currently doesn't understand "package private" APIs (which these all are), and we have found it to be the case that some APIs don't have any useful information to add - and in those cases we prefer omitting the docs to adding redundant docs.

natebosch:
> [cloud_firestore/permission-denied] The caller does not have permission to execute the specified
  operation.
  
  Are these flakes?
  It looks like there might also be some failures on `master` branch - is it OK to merge on red CI in this repo?

russellwheatley:
@natebosch - you can ignore the firestore e2e as they relate to using second database which doesn't have emulator support and thus uses a live firebase project, I have app check enforcement on the firestore API as I am testing something out. When this PR gets to review state, I'll remove enforcement 😄, just ping me here (I think enforcement will be removed sooner to be honest).

cynthiajoan (changes requested):

packages/firebase_vertexai/firebase_vertexai/lib/src/vertex_model.dart
```diff
@@ -232,9 +230,37 @@ final class GenerativeModel {
...
+  google_ai.GenerativeModel get googleAIModel => _googleAIModel;
+}
+
+/// Returns a [GenerativeModel] using it's private constructor.
+GenerativeModel createGenerativeModel({
+  required FirebaseApp app,
+  required String location,
+  required String model,
+  Content? systemInstruction,
+  FirebaseAppCheck? appCheck,
```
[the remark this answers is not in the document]
natebosch: I had confused myself during the merge and added the argument in the wrong place, but I think it's cleaned up now. This should forward directly because it's taking place of the private constructor in uasge.

cynthiajoan (approved):

russellwheatley (approved):
Small nit for spelling, but LGTM

Issue #13003, opened 2024-06-25
[firebase_database]: web no result when listening to multiple onValue streams on the same path


robindijkhof:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Which plugins are affected?

Database

### Which platforms are affected?

Web

### Description

Listening to the same path multiple times will only return a result to the latest stream. Please take a look at the example below for more info.  This worked fine in `firebase_database 10.5.0` with `firebase_core: 2.28.0` but is now broken in `firebase_database: 11.0.2` with `firebase_core 3.1.1`. 

### Reproducing the issue

Consider the following block
    
    var ref = FirebaseDatabase.instance.ref().child('/somepath');

    ref.onValue.listen((data) {print('got them 1');}, onError: (e) => print('error'), onDone: () => print('done'));
    ref.onValue.listen((data) {print('got them 2');}, onError: (e) => print('error'), onDone: () => print('done'));
    ref.onValue.listen((data) {print('got them 3');}, onError: (e) => print('error'), onDone: () => print('done'));

On web this will only print `got them 3`. On Android it will print `got them 1`, `got them 2`, `got them 3`

Listening to a subpath works fine though. So the following block works as expected:

    FirebaseDatabase.instance.ref().child('/somepath').listen(...)
    FirebaseDatabase.instance.ref().child('/somepath/sub').listen(...)

### Firebase Core version

3.1.1

### Flutter Version

3.22.2

### Flutter dependencies

<details>
<summary>Expand <code>Flutter dependencies</code> snippet</summary>
<br>

```yaml

Dart SDK 3.4.3
Flutter SDK 3.22.2
urenapp 5.20.0+215

dependencies:
- cached_network_image 3.3.1 [cached_network_image_platform_interface cached_network_image_web flutter flutter_cache_manager octo_image]
- collection 1.18.0
- cupertino_icons 1.0.8
- device_info_plus 10.1.0 [device_info_plus_platform_interface ffi file flutter flutter_web_plugins meta web win32 win32_registry]
- dropdown_search 5.0.6 [flutter]
- file 7.0.0 [meta path]
- firebase_analytics 11.1.0 [firebase_analytics_platform_interface firebase_analytics_web firebase_core firebase_core_platform_interface flutter]
- firebase_auth 5.1.1 [firebase_auth_platform_interface firebase_auth_web firebase_core firebase_core_platform_interface flutter meta]
[...]

PR #13029, opened 2024-07-02
fix(database, web): ensure exact same streams are not unsubscribed in debug mode

russellwheatley:
## Description

If exact same stream is created (i.e same hashCode generated from parameters), it will unsubscribe in debug mode.

I've created a map used in debug mode to track the count of the exact same stream and appended the count to the hashCode to ensure they are subscribed/unsubscribed separately.

Here is the user's code I used in example app:
```dart
ref.onValue.listen((data) {
                print('got them 1');
              }, onError: (e) => print('error: ${e}'), onDone: () => print('done'));
              ref.onValue.listen((data) {
                print('got them 2');
              }, onError: (e) => print('error ${e}'), onDone: () => print('done'));
              ref.onValue.listen((data) {
                print('got them 3');
              }, onError: (e) => print('error ${e}'), onDone: () => print('done'));
```

Here are console logs showing they all get the event as it was previously.
Screenshot 2024-07-02 at 13 18 02

Here is a screenshot to show 3 hot restarts, after the third hot restart I updated the reference data in the Firebase console which is why there is an additional set of logs. This proves that they are still being cleaned up as expected on hot restarts otherwise we would expect an additional 9 set of logs instead of 3:
Screenshot 2024-07-02 at 13 39 50

## Related Issues

fixes https://github.com/firebase/flutterfire/issues/13003

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_database/firebase_database_web/lib/src/interop/database.dart
+++ b/packages/firebase_database/firebase_database_web/lib/src/interop/database.dart
@@ -250,7 +250,7 @@ class Query<T extends database_interop.QueryJsImpl> extends JsObjectWrapper<T> {
   /// DatabaseReference to the Query's location.
   DatabaseReference get ref => DatabaseReference.getInstance(jsObject.ref);
 
-  Stream<QueryEvent> _onValue(String appName, int hashCode) => _createStream(
+  Stream<QueryEvent> _onValue(String appName, String hashCode) => _createStream(
         'value',
         appName,
         hashCode,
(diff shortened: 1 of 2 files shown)


Lyokone (approved):

TarekkMA (approved):

alextekartik:
I think the debug map that holds subscribed key only works when listening to the same DocumentReference (i.e. not when listening to two different DocumentReferences pointing to the same path)

A simple code like this should help reproduce this issue.

```
var path = ...
var ref1 = firestore.doc(path);
var ref2 = firestore.doc(path);
ref1.snapshots().listen((_) => ...);
// This cancels the first listen.
ref2.snapshots().listen((_) => ...);
```

Basically I don't think _docListeners (and others similar chear for debug) should be store at the DocumentReference level but rather be at the firestore level as the key (ending with 0) will be the same for 2 different refs pointing to the same path.

I'm not how existing registrations are really cancelled upon hot restart. Every stored callback in a js global should be called but this is not the case. I understand registration are saved in js global in the form xxx_0, xxx_1, xxx_2... But xxx_0 is only cancelled when a first registration occurs. I think (maybe in firebase_core_web) that all existing callbacks should be called upon start.

PR #13191, opened 2024-08-16
chore(deps): updating web dependency in all packages. bump web from 0.5.1 to 1.0.0.

vinodh013:
updating all packages web dependency due to dependency constraints


behzodfaiziev (approved):

JanBrin (approved):

Lyokone:
As you can see in https://pub.dev/packages/web/changelog (and it was explained in already 2 issues)
```
Require Dart ^3.4.0.
```

We are not going to drop support for Dart < 3.4.0 right away. If you have an issue with this, you can use 
```
dependency_overrides:
  web: ^1.0.0
```

PR #13358, opened 2024-09-18
Add GMP header

maneesht:


cynthiajoan (approved):

cynthiajoan:
@maneesht seems the e2e test needs to be updated as well for this change.

packages/firebase_data_connect/firebase_data_connect/example/test/widget_test.dart
```diff
@@ -0,0 +1,34 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// This is a basic Flutter widget test.
+//
+// To perform an interaction with a widget in your test, use the WidgetTester
+// utility in the flutter_test package. For example, you can send tap and scroll
+// gestures. You can also use WidgetTester to find child widgets in the widget
+// tree, read text, and verify that the values of widget properties are correct.
+
+import 'package:flutter/material.dart';
```
russellwheatley, line 12: Isn't this test boilerplate test of a Flutter widget? Not sure if this is supposed to be here.

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_options.dart
```diff
@@ -21,9 +21,9 @@ class ConnectorConfig {
   /// String representation of connectorConfig
   String toJson() {
     return jsonEncode({
-      'location': location,
-      'connector': connector,
-      'serviceId': serviceId,
+      location: location,
+      connector: connector,
+      serviceId: serviceId,
```
russellwheatley, line 26: If the keys are strings, then you should revert this change I believe.

packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
```diff
@@ -54,24 +54,19 @@ class FirebaseDataConnect extends FirebasePluginPlatform {
...
+    transport = getTransport(
+        transportOptions!, options, app.options.appId, auth, appCheck);
   }
 
   /// Returns a [QueryRef] object.
   QueryRef<Data, Variables> query<Data, Variables>(
       String operationName,
       Deserializer<Data> dataDeserializer,
-      Serializer<Variables>? varsSerializer,
+      Serializer<Variables> varsSerializer,
```
russellwheatley, line 65: Considering the varSerializer` has been made a required arg here as well, we should probably aim to merge this one first: https://github.com/firebase/flutterfire/pull/13386 so it is in the CHANGELOG.

russellwheatley (changes requested):

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_options.dart
maneesht: Is it a best practice to have the keys as quotes?

russellwheatley: The problem is, the key would be the same value as the value. 

So if you didn't change, the result would be:
```dart
{
"usa":"usa"
}
```

Rather than what I think would be more appropriate:
```dart
{
"location": "usa"
}
```

maneesht: Ah, thanks for catching that!

russellwheatley (approved):

PR #13373, opened 2024-09-20
feat(fdc): Upgrade to v1beta endpoint

maneesht:


packages/firebase_data_connect/firebase_data_connect/lib/src/generated/connector_service.pbenum.dart
```diff
@@ -1,7 +1,3 @@
-// Copyright 2024, the Chromium project authors.  Please see the AUTHORS file
```
russellwheatley, line 1: I appreciate this is quite annoying, but all the generated src code has lost their license headers.

russellwheatley (changes requested):
The license headers + running `melos run format` for this PR 🙏

packages/firebase_data_connect/firebase_data_connect/lib/src/generated/connector_service.pbenum.dart
maneesht: It looks like the melos scripts aren't actually adding the headers. I'll do it manually

maneesht: Updated

russellwheatley: just noticed this in the melos script: https://github.com/firebase/flutterfire/blob/main/melos.yaml#L285. Not sure if it needs to be there but this is why you're probably not seeing it added.

maneesht: Ah, makes sense

russellwheatley (approved):

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/seed_movies.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  FirebaseDataConnect dataConnect;
+}
+
+class SeedMoviesTheMatrix {
+  late String id;
+
+  SeedMoviesTheMatrix.fromJson(Map<String, dynamic> json) : id = json['id'] {}
+
+  // TODO(mtewani): Fix up to create a map on the fly
```
cynthiajoan, line 23: I thought I see this removed in the generation code.

cynthiajoan (approved):

aashishpatil-g (approved):

PR #13376, opened 2024-09-20
feat(fdc): Implement any scalar support

maneesht:


packages/firebase_data_connect/firebase_data_connect/lib/src/optional.dart
```diff
@@ -50,16 +50,13 @@ class Optional<T> {
   }
 
   /// Converts the value to String.
```
cynthiajoan, line 52: Converts to dynamic?

cynthiajoan (approved):

russellwheatley (approved):

PR #13387, opened 2024-09-23
Implement Timestamp and DateTime types

maneesht:


packages/firebase_data_connect/firebase_data_connect/example/lib/main.dart
```diff
@@ -109,8 +109,14 @@ class _DataConnectWidgetState extends State<DataConnectWidget> {
...
     super.initState();
 
-    QueryRef<ListMoviesResponse, void> ref =
+    QueryRef<ListMoviesData, void> ref =
         MoviesConnector.instance.listMovies.ref();
+    // MoviesConnector.instance.addDateAndTimestamp
+    //     .ref(date: DateTime.now(), timestamp: Timestamp(123456789, 864000))
+    //     .execute();
+    MoviesConnector.instance.listTimestamps.ref().execute().then((res) {
+      print(res.data.timestampHolders.first.toJson());
```
cynthiajoan, line 118: let's not use print

packages/firebase_data_connect/firebase_data_connect/example/lib/main.dart
```diff
@@ -109,8 +109,14 @@ class _DataConnectWidgetState extends State<DataConnectWidget> {
   void initState() {
     super.initState();
 
-    QueryRef<ListMoviesResponse, void> ref =
+    QueryRef<ListMoviesData, void> ref =
         MoviesConnector.instance.listMovies.ref();
+    // MoviesConnector.instance.addDateAndTimestamp
```
cynthiajoan, line 114: Still need to keep those commented out code?

maneesht: Removed

packages/firebase_data_connect/firebase_data_connect/example/lib/main.dart
maneesht: Removed

packages/firebase_data_connect/firebase_data_connect/lib/src/optional.dart
```diff
@@ -65,11 +65,16 @@ class Optional<T> {
...
-  if (type is bool || type is int || type is double || type is num) {
-    return type.toString();
-  } else if (type is String) {
+dynamic nativeToJson<T>(T type) {
+  if (type is bool ||
+      type is int ||
+      type is double ||
+      type is num ||
+      type is String) {
     return type;
```
russellwheatley, line 74: Are you sure this is correct? If we're making a type a string, wouldn't we need to keep:

```dart
if (type is bool || type is int || type is double || type is num) {
    return type.toString();
  }
```

russellwheatley (changes requested):

packages/firebase_data_connect/firebase_data_connect/lib/src/optional.dart
maneesht: I modified the serializers to accept dynamic return types because of the `AnyScalar`, which allows you to pass in any json-serializable type. If we allow serialization to Strings in this function, over the wire, we'll be sending `"true"` instead of `true`

PR #13433, opened 2024-09-30
fix(fdc): Handled errors

maneesht:
Fixed issue where if the server responds with an error, then the Data Connect library wouldn't throw an error.


packages/firebase_data_connect/firebase_data_connect/example/dataconnect/connector/mutations.gql
```diff
@@ -70,6 +71,7 @@ mutation thing($title: Any! = "ABC") @auth(level: PUBLIC) {
 mutation seedData @auth(level: PUBLIC) {
   the_matrix: movie_insert(
     data: {
+      id: "09d5f835656c467787347668bbb44522"
```
maneesht, line 74: Changes to reproduce an error. If a user calls this more than once, the second time would throw an error.

Lyokone (approved):

PR #13450, opened 2024-10-01
fix(fdc): Fix serializing errors

maneesht:
* Fix issue where on GRPC, since protobuf doesn't distinguish between double and int, all numbers were being parsed as doubles.
* Fix issue where `execute.ignore` was called, causing errors not being thrown.

### DIFF
--- a/packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
+++ b/packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
@@ -111,7 +111,6 @@ class QueryRef<Data, Variables> extends OperationRef<Data, Variables> {
           serializer(variables as Variables), this, res.data, null);
       return res;
     } on Exception catch (e) {
-      print(e);
       await _queryManager.triggerCallback<Data, Variables>(
           operationName, serializer(variables as Variables), this, null, e);
       rethrow;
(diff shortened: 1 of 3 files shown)


packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -124,7 +123,11 @@ class QueryRef<Data, Variables> extends OperationRef<Data, Variables> {
         .addQuery(operationName, variables, varsSerialized)
         .cast<QueryResult<Data, Variables>>();
     if (_queryManager.containsQuery(operationName, variables, varsSerialized)) {
-      this.execute().ignore();
+      try {
+        this.execute();
+      } catch (_) {
+        // Call to `execute` should properly pass the error to the Stream.
```
cynthiajoan, line 129: Should we give some message here?

maneesht: As in throw something? Or log?

Lyokone (approved):

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
maneesht: Added

PR #16672, opened 2024-11-15
fix wasm analysis

kevmoo:


Lyokone (approved):

russellwheatley (approved):

arktronic:
Why was this `BREAKING CHANGE` published as a _patch_ version of 5.3.1?

kevmoo:
> Why was this `BREAKING CHANGE` published as a _patch_ version of 5.3.1?

I wasn't involved in the versioning discussion of this change.

What broken on your side? How are you using the test features!

And also: sorry for the pain. Are you able to work-around the issues?

arktronic:
Multiple CI pipelines broke because of tests being unable to execute. We worked around the issue by pinning version 5.3.0. I just wanted to inform whomever is involved with the release/versioning process that this should have been a major version bump, rather than a patch bump. Thanks!

PR #16827, opened 2024-12-05
fix: Remove dart:io imports for analytics, auth and app check

kevmoo:
Allows these packages to be detected as wasm-compatible

Packages:
- firebase_analytics
- firebase_app_check
- firebase_auth


kevmoo:
@russellwheatley @Lyokone – how should I handle versioning and changelogs here?

russellwheatley (approved):

russellwheatley:
@kevmoo - you don't have to worry about versioning, we use melos that handles it for us based on conventional commits.

kevmoo:
@russellwheatley – okay. Unlike the other ones, this shouldn't be breaking. I also don't think the e2e failure is related to this PR, but I'll let you land it.

Lyokone (approved):

PR #17078, opened 2025-02-11
fix(vertexai): make token-based metrics non nullable

rlazo:
If missing, they'll default to empty.


packages/firebase_vertexai/firebase_vertexai/lib/src/api.dart
```diff
@@ -33,7 +33,7 @@ final class CountTokensResponse {
   final int? totalBillableCharacters;
```
cynthiajoan, line 33: If totalBillableCharacters is also required, then we should remove the ? here

rlazo:
After discussing the proposal, the API will remain as-is

PR #17148, opened 2025-02-28
feat(fdc): Implemented partial errors

maneesht:


packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -28,6 +28,36 @@ class DataConnectError extends FirebaseException {
...
+/// Nested class containing errors and decoded data.
+class DataConnectOperationFailureResponse<T> {
+  DataConnectOperationFailureResponse(this.errors, this.data, this.decodedData);
+  final Map<String, dynamic>? data;
+  final List<DataConnectOperationFailureREsponseErrorInfo> errors;
+  final T? decodedData;
+}
+
+/// Error information per error.
+class DataConnectOperationFailureREsponseErrorInfo {
```
cynthiajoan, line 48: Capital E is typo?

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -18,8 +18,8 @@ part of 'common_library.dart';
 enum DataConnectErrorCode { unavailable, unauthorized, other }
 
 /// Error thrown when DataConnect encounters an error.
-class DataConnectError extends FirebaseException {
-  DataConnectError(this.dataConnectErrorCode, message)
+class DataConnectError<T> extends FirebaseException {
```
dconeybe, line 21: Is this intended that `DataConnectError` has a type parameter? It doesn't look like it's used anywhere. Should it just be the subclass `DataConnectOperationError` that has a type parameter?

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -28,6 +28,36 @@ class DataConnectError extends FirebaseException {
   final DataConnectErrorCode dataConnectErrorCode;
 }
 
+/// Error thrown when an operation is partially successful.
+class DataConnectOperationError extends DataConnectError {
+  DataConnectOperationError(
+      DataConnectErrorCode code, String message, this.response)
+      : super(code, message);
+  final DataConnectOperationFailureResponse response;
```
dconeybe, line 36: Shouldn't `DataConnectOperationFailureResponse` have a type parameter here?

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -28,6 +28,36 @@ class DataConnectError extends FirebaseException {
...
+
+/// Error information per error.
+class DataConnectOperationFailureResponseErrorInfo {
+  DataConnectOperationFailureResponseErrorInfo(this.path, this.message);
+  String message;
+  List<DataConnectOperationFailureErrorInfoPathSegment> path;
+}
+
+/// Path where error occurred.
+class DataConnectOperationFailureErrorInfoPathSegment {
```
dconeybe, line 55: Could PathSegment be defined as a `sealed class` type? I got an example like this from Gemini, and this is what Kotlin does:

```dart
sealed class StringOrInt {}

class StringValue extends StringOrInt {
  final String value;
  StringValue(this.value);
}

class IntValue extends StringOrInt {
  final int value;
  IntValue(this.value);
}
```

maneesht: It feels like it's more work for the user to check whether it's an IntValue or a StringValue in this situation. Why can't the user just do `?.intValue` or `stringValue`?

dconeybe: It is, indeed, a bit more code for customers to write if we go with my suggested "sealed classes" approach. However, IMO it's generally preferable to enforce invariants via the type system, if possible.  The way that it is currently written, with the `DataConnectOperationFailureErrorInfoPathSegment` class with `field` and `listIndex` properties, it is technically possible for both `field` and `listIndex` to be null, and also technically possible for both to be non-null, both of which are invalid states because _exactly one_ of them must be non-null. Using the "sealed class" approach enforces this invariant.

Feel free to do whatever you think will be most idiomatic in Dart/Flutter code. I had assumed that Dart lacked the expressiveness of sealed classes so didn't bring up this suggestion earlier. Then I decided to ask my favorite chat bot, Gemini :)

dconeybe (approved):

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -28,6 +28,44 @@ class DataConnectError extends FirebaseException {
...
+
+/// Error information per error.
+class DataConnectOperationFailureResponseErrorInfo {
+  DataConnectOperationFailureResponseErrorInfo(this.path, this.message);
+  String message;
+  List<DataConnectOperationFailureErrorInfoPathSegment> path;
+}
+
+/// Path where error occurred.
+sealed class DataConnectOperationFailureErrorInfoPathSegment {}
```
aashishpatil-g, line 55: Can we name this as just DataConnectPathSegment? It will have reusabality outside of an error context too.

maneesht: Done.

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -28,6 +28,42 @@ class DataConnectError extends FirebaseException {
...
+  DataConnectOperationError(
+      DataConnectErrorCode code, String message, this.response)
+      : super(code, message);
+  final DataConnectOperationFailureResponse<T> response;
+}
+
+/// Nested class containing errors and decoded data.
+class DataConnectOperationFailureResponse<T> {
+  DataConnectOperationFailureResponse(this.errors, this.data, this.decodedData);
+  final Map<String, dynamic>? data;
```
aashishpatil-g, line 42: I think we have to call this `rawData` per API review feedback. 

cc: @dconeybe

dconeybe: Yes, that is true: `data`/`decodedData` has been renamed to `rawData`/`data`, as per api council feedback.

maneesht: Updated

aashishpatil-g (approved):

cynthiajoan (approved):

packages/firebase_data_connect/firebase_data_connect/example/android/app/build.gradle
```diff
@@ -26,7 +26,7 @@ if (flutterVersionName == null) {
 android {
     namespace = "io.flutter.plugins.firebase.dataconnect.example"
     compileSdk = flutter.compileSdkVersion
-    ndkVersion = flutter.ndkVersion
+            ndkVersion = "27.0.12077973"
```
russellwheatley, line 29: is this needed change needed? either way, please format.

russellwheatley (changes requested):
small comment, otherwise, LGTM

russellwheatley (approved):

PR #17332, opened 2025-05-05
feat(vertexai): Add a new package firebase_ai

natebosch:
Copy all the code from firebase_vertexai and rename with "Vertex"
removed or replaced with "Firebase".

Make firebase_vertexai an export package, retaining only the
FirebaseVertexAI implementation and adding `typedef` aliases for 2
renamed exception types. Replace the rest with exports from
`package:firebase_ai`.

Revert the examples changes to use googleAI - the example in the
`firebase_ai` package will show both while the old example shows only
the usage of vertex..

Import from src across the packages to keep existing tests as close to
identical as possible despite doubling up test cases with the upstream
package.


natebosch:
If you'd prefer to review and land any parts of this as separate commits I'm happy to split it up. Most of the LOC here are copy/paste (and lots of boilerplate for the flutter stuff) but there is no way to indicate a file rename in git so the review will show everything as new.

For the most part this reflects a global rename from `FirebaseVertexAI*` or `VertexAI*` to `FirebaseAI*`. Some URLs to help docs are not changed here and may need to be updated in coordination with the doc hosting.

natebosch:
We may need some CI config changes.

packages/firebase_ai/firebase_ai/lib/src/firebase_ai.dart
```diff
@@ -0,0 +1,187 @@
...
+  /// If [app] is not provided, the default Firebase app will be used.
+  /// If pass in [appCheck], request session will get protected from abusing.
+  static FirebaseAI vertexAI({
+    FirebaseApp? app,
+    FirebaseAppCheck? appCheck,
+    FirebaseAuth? auth,
+    String? location,
+  }) {
+    app ??= Firebase.app();
+
```
cynthiajoan, line 66: would need to apply the fix of var instanceKey = '${app.name}::googleai';

See here: https://github.com/firebase/flutterfire/pull/17319/files#diff-edacf0f74d51415959b1f96ed817e5a7cc222da1b27d3fd5904776f21a62815b

packages/firebase_vertexai/firebase_vertexai/example/lib/main.dart
cynthiajoan, line 1: the example app in firebase_vertexai seems a few versions behind. Any reason they cannot be same as the example app in firebase_ai?

packages/firebase_vertexai/firebase_vertexai/example/lib/pages/bidi_page.dart
```diff
@@ -59,7 +59,7 @@ class _BidiPageState extends State<BidiPage> {
     super.initState();
 
     final config = LiveGenerationConfig(
-      speechConfig: SpeechConfig(voiceName: 'Fenrir'),
+      speechConfig: SpeechConfig(voice: Voice.fenrir),
```
cynthiajoan, line 62: This needs to keep as the left one. The api has changed for bidi feature.

packages/firebase_vertexai/firebase_vertexai/lib/firebase_vertexai.dart
```diff
@@ -29,54 +32,53 @@ export 'src/api.dart'
...
         SafetySetting,
-        // TODO(cynthiajiang) remove in next breaking change.
-        TaskType,
-        UsageMetadata;
-export 'src/base_model.dart'
-    show GenerativeModel, ImagenModel, LiveGenerativeModel;
-export 'src/chat.dart' show ChatSession, StartChatExtension;
-export 'src/content.dart'
-    show
+        UsageMetadata,
```
cynthiajoan, line 35: I was planning to remove this field, since there's no use case for developer to instantiate the UsageMetadata. Do you think it's a good time for this change?

cynthiajoan:
> We may need some CI config changes.

Yeah would need to add the new package to melos config.

packages/firebase_ai/firebase_ai/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.5.0
```
cynthiajoan, line 1: let's start with 2.0.0

packages/firebase_ai/firebase_ai/pubspec.yaml
```diff
@@ -0,0 +1,39 @@
+name: firebase_ai
+description: Firebase AI SDK.
+version: 0.1.0
```
cynthiajoan, line 3: let's have the version start 2.0.0

packages/firebase_vertexai/firebase_vertexai/example/lib/main.dart
natebosch: I reverted the change that made the app support googleai because I removed googleai support from the class with the old name. I can add it back if we want both demos identical for now.

packages/firebase_ai/firebase_ai/lib/src/firebase_ai.dart
natebosch: Done.

packages/firebase_vertexai/firebase_vertexai/example/lib/pages/bidi_page.dart
natebosch: Ah, I had analysis errors because I wasn't overriding to the local package copy. Fixed.

packages/firebase_vertexai/firebase_vertexai/lib/firebase_vertexai.dart
natebosch: We can drop the export of TaskType since I think that's unused.

Do we really want to remove this type? It's still user visible right? We can consider removing the public constructor, but if we expect user code to read from this class I think we do want to export it.

packages/firebase_vertexai/firebase_vertexai/test/base_model_test.dart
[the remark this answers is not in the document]
natebosch: Sure, I think it's safe to drop stuff that isn't testing the public API.

packages/firebase_ai/firebase_ai/CHANGELOG.md
natebosch: Done.

packages/firebase_ai/firebase_ai/pubspec.yaml
natebosch: Done.

packages/firebase_vertexai/firebase_vertexai/lib/firebase_vertexai.dart
cynthiajoan: Sounds good, let's remove the public constructor in a follow up PR.

cynthiajoan (approved):

PR #17468, opened 2025-06-30
WIP: Grounding with Google Search

dlarocque:
API proposal: [go/fal-grounding-api](http://go/fal-grounding-api) (internal)

### DIFF
--- a/packages/firebase_vertexai/firebase_vertexai/lib/src/api.dart
+++ b/packages/firebase_vertexai/firebase_vertexai/lib/src/api.dart
@@ -169,7 +169,8 @@ final class Candidate {
   // TODO: token count?
   // ignore: public_member_api_docs
   Candidate(this.content, this.safetyRatings, this.citationMetadata,
-      this.finishReason, this.finishMessage);
+      this.finishReason, this.finishMessage,
+      {this.groundingMetadata});
 
   /// Generated content returned from the model.
   final Content content;
@@ -193,6 +194,8 @@ final class Candidate {
 
   /// Message for finish reason.
   final String? finishMessage;
+
+  final GroundingMetadata? groundingMetadata;
 
   /// The concatenation of the text parts of [content], if any.
   ///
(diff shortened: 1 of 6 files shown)


packages/firebase_vertexai/firebase_vertexai/lib/src/api.dart
```diff
@@ -938,29 +994,33 @@ Candidate _parseCandidate(Object? jsonObject) {
...
-      {'finishReason': final Object finishReason} =>
-        FinishReason._parseValue(finishReason),
-      _ => null
-    },
-    switch (jsonObject) {
-      {'finishMessage': final String finishMessage} => finishMessage,
-      _ => null
-    },
-  );
+      jsonObject.containsKey('content')
```
cynthiajoan: the indent looks wrong here.

cynthiajoan (approved):

packages/firebase_ai/firebase_ai/lib/firebase_ai.dart
```diff
@@ -51,7 +51,7 @@ export 'src/error.dart'
         ServerException,
         UnsupportedUserLocation;
 export 'src/firebase_ai.dart' show FirebaseAI;
-export 'src/function_calling.dart'
```
cynthiajoan, line 54: Can you also check and update the export part under firebase_vertexai/firebase_vertexai. Even though we would want to deprecate that package, still need to make sure it's not broken because of this change.

cynthiajoan:
Pelease fix the all_plugins/analyze and all_plugins/test
https://github.com/firebase/flutterfire/actions/runs/16124843776/job/45499206956?pr=17468

packages/firebase_ai/firebase_ai/lib/firebase_ai.dart
dlarocque: It looks like [`lib/firebase_vertexai.dart`](https://github.com/firebase/flutterfire/blob/dl/grounding/packages/firebase_vertexai/firebase_vertexai/lib/firebase_vertexai.dart) does not export directly from the name of the file `src/function_calling.dart` anymore. Does this mean this should be fine?

cynthiajoan: Yeah that would be fine

russellwheatley (approved):

PR #17563, opened 2025-07-24
feat(messaging)!: remove deprecated functions

Lyokone:
# Remove deprecated sendMessage method from firebase_messaging

This PR removes the deprecated `sendMessage` method from the firebase_messaging package as part of preparing for a breaking change release. The method was deprecated due to Firebase decommissioning this functionality in June 2024.

## Changes

- Removed `sendMessage` method from main FirebaseMessaging class
- Removed method from platform interface
- Removed method from method channel implementation  
- Updated tests to remove sendMessage test cases

## Migration Guide

The `sendMessage` method was Android-only and allowed sending FCM messages from the client. To migrate, use Firebase Cloud Functions or your backend server to send messages instead of the client-side method.

## Breaking Change

This removes the deprecated `sendMessage` method entirely. Apps using this method will need to migrate to server-side message sending before upgrading.


roberto-gigli:
Old installation of firebase messaging stopped working after this update

/firebase_messaging-14.9.1/lib/src/messaging.dart:220:22: Error: The method 'sendMessage' isn't defined for the class 'FirebaseMessagingPlatform'.

PR #17761, opened 2025-10-02
feat(firebase_ai): add ImageConfig and new FinishReason enums

thatfiredev:


paulb777:
Changes should now be incorporated into #18180

PR #17930, opened 2026-01-01
test(ci): implement best practices for caching in workflows

mikehardy:
## Description

👋  this is started with the work on #17871 from @SelaseKay 

I worked with Jude on that change in 17871 and while I was in there I noticed that the caching setup for CI in this repo was likely to be subject to severe LRU thrashing and very very low hit rates, with low useful caches even with hits.

In fact when I went to verify that, it was true: https://github.com/firebase/flutterfire/actions/caches?query=sort%3Acreated-asc

The oldest cache is only 35 **minutes** old - and all the caches are from a branch which means main builds or builds from other branches can't even see them to use them. Thus the caching is for most intents and purposes a lot of computational work for zero benefit at the moment.

**How to review this PR** -  Each commit in this series is a fully separate idea to improve things, and each should be reviewed separately by *reading the commit message* which explains that specific change then checking the diff for that commit

The basic idea is this though:
- caches are not visible from branch to main, or between OSs by rule (so you want to cache main but read-only on branch)
- caches are immutable if the primary key was hit, so key uniqueness is vital, but it must actually change when the cache is known to be stale
- cache failures shouldn't fail a workflow (it happens, it's a useless flake to fail on)
- firestore emulators are truly cross-platform so may be shared

In addition, I pinned the Android Emulator runner to a concrete SHA as is practice in this repo

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process. Updating the `pubspec.yaml` and changelogs is not required.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] The analyzer (`melos run analyze`) does not report any problems on my PR.
- [ ] I read and followed the [Flutter Style Guide].
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change.
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/firebase/flutterfire/blob/main/CONTRIBUTING.md
[...]

PR #17943, opened 2026-01-12
test(ci): cache key names must include enough information to expire correctly

mikehardy:
## Description

After the initial CI caching best practices PR merged I noticed a problem that was hidden by the LRU thrashing before

https://github.com/firebase/flutterfire/actions/caches?query=sort%3Acreated-asc

See the flutter- and pub- caches? They are not unique keys, they will never update.

The same problem was present for the Pods caches - but for a different reason, in that case the Podfile.lock files don't exist so they cannot be used to generate a cache key name part. I deferred to the pubspec.lock file as it drives the Pods dependencies anyway

----

the flutter action caches things by default with a cache key name that is not specific enough that it will expire if a new version comes out

alter the key naming to include all pertinent details about the flutter version (or pub version) being cached so that new versions will get a fresh cache

see https://github.com/subosito/flutter-action?tab=readme-ov-file#caching in combination with knowledge that github actions/cache produces *immutable* caches so they will never update meaning you must make a new cache if a new version comes out

## Related Issues

- #17930 

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/.github/workflows/all_plugins.yaml
+++ b/.github/workflows/all_plugins.yaml
@@ -28,6 +28,8 @@ jobs:
         with:
           channel: 'stable'
           cache: true
+          cache-key: "flutter-:os:-:channel:-:version:-:arch:-:hash:"
+          pub-cache-key: "flutter-pub-:os:-:channel:-:version:-:arch:-:hash:"
       - uses: bluefireteam/melos-action@c7dcb921b23cc520cace360b95d02b37bf09cdaa
         with:
           melos-version: '5.3.0'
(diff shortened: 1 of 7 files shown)


russellwheatley (approved):

mikehardy:
Here's how I saw this on the caches page, and the assumption is that after merge and a CI run the keys will be fixed.

Square boxes are the flutter/pub names that are overly general. Those caches will never be updated or expire because they will hit but all caches are immutable so the hit means it will remain unchanged for eternity. Adding all the version / arch etc info makes for an unambiguous key name which will change as flutter changes.

Circle is similar problem but for Pods, and it was because the file that *should* have made a unique key when it was hashed was `Podfile.lock` which is not available during cache restore. `pubspec.lock` will be though.

image

SelaseKay (approved):

PR #17961, opened 2026-01-20
fix(firebase_ai): Rename `groundingSupport` to `groundingSupports`

dlarocque:
We currently parse for a `groundingSupport` field, when it should be `groundingSupports`. This fixes an issue where `groundingSupports` are not parsed in responses or present in the API, and cannot be retrieved by users.

Fixes [b/477107542](https://b.corp.google.com/issues/477107542)

### DIFF
--- a/packages/firebase_ai/firebase_ai/lib/src/api.dart
+++ b/packages/firebase_ai/firebase_ai/lib/src/api.dart
@@ -418,6 +418,13 @@ final class GroundingMetadata {
 
   /// A list of [GroundingSupport]s.
   ///
+  /// This list will always be empty. See b/477107542.
+  /// Keeping for backwards compatibility.
+  @Deprecated('Use groundingSupports instead')
+  final List<GroundingSupport> groundingSupport = [];
+
+  /// A list of [GroundingSupport]s.
+  ///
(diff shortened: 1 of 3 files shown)


cynthiajoan (approved):

packages/firebase_ai/firebase_ai/lib/src/api.dart
```diff
@@ -420,7 +420,7 @@ final class GroundingMetadata {
   ///
   /// Each object details how specific segments of the
   /// model's response are supported by the `groundingChunks`.
-  final List<GroundingSupport> groundingSupport;
+  final List<GroundingSupport> groundingSupports;
```
cynthiajoan, line 429:
```suggestion
  final List<GroundingSupport> groundingSupports;
  
  /// A list of [GroundingSupport]s.
  @Deprecated('Use groundingSupports instead')
  List<GroundingSupport> get groundingSupport => groundingSupports;
```

cynthiajoan (changes requested):
keep the current groundingSupport and mark it as deprecated

cynthiajoan (approved):

russellwheatley (approved):

PR #18131, opened 2026-03-20
feat(firebaseai): Add ability for Schema class to export to json schema

cynthiajoan:
## Description

## Related Issues

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_ai/firebase_ai/example/ios/Podfile
+++ b/packages/firebase_ai/firebase_ai/example/ios/Podfile
@@ -29,6 +29,7 @@ flutter_ios_podfile_setup
 
 target 'Runner' do
   use_frameworks!
+  use_modular_headers!
 
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
   target 'RunnerTests' do
--- a/packages/firebase_ai/firebase_ai/example/ios/Podfile
+++ b/packages/firebase_ai/firebase_ai/example/ios/Podfile
@@ -29,7 +29,6 @@ flutter_ios_podfile_setup
 
 target 'Runner' do
   use_frameworks!
-  use_modular_headers!
 
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
   target 'RunnerTests' do
--- a/packages/firebase_ai/firebase_ai/lib/src/schema.dart
+++ b/packages/firebase_ai/firebase_ai/lib/src/schema.dart
@@ -371,12 +371,11 @@ final class JSONSchema extends Schema {
 
   /// Construct a schema for an integer number.
   ///
-  /// The [format] may be "int32" or "int64".
+  /// Json schema integer doesn't support format.
   JSONSchema.integer({
     String? description,
     String? title,
     bool? nullable,
-    String? format,
     int? minimum,
     int? maximum,
   }) : this(
@@ -384,19 +383,17 @@ final class JSONSchema extends Schema {
           description: description,
           title: title,
           nullable: nullable,
-          format: format,
           minimum: minimum?.toDouble(),
           maximum: maximum?.toDouble(),
         );
 
   /// Construct a schema for a non-integer number.
   ///
-  /// The [format] may be "float" or "double".
+  /// Json schema number doesn't support format.
   JSONSchema.number({
     String? description,
     String? title,
     bool? nullable,
-    String? format,
     double? minimum,
     double? maximum,
   }) : this(
@@ -404,7 +401,6 @@ final class JSONSchema extends Schema {
           description: description,
           title: title,
           nullable: nullable,
-          format: format,
           minimum: minimum,
           maximum: maximum,
         );
--- a/packages/firebase_ai/firebase_ai/lib/src/schema.dart
+++ b/packages/firebase_ai/firebase_ai/lib/src/schema.dart
@@ -35,6 +35,14 @@ final class Schema {
     this.optionalProperties,
     this.propertyOrdering,
     this.anyOf,
+    this.ref,
+    this.minProperties,
+    this.maxProperties,
+    this.minLength,
+    this.maxLength,
+    this.pattern,
+    this.example,
+    this.defaultValue,
   });
 
   /// Construct a schema for an object with one or more properties.
@@ -45,6 +53,10 @@ final class Schema {
     String? description,
     String? title,
     bool? nullable,
+    int? minProperties,
+    int? maxProperties,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.object,
           properties: properties,
@@ -53,6 +65,10 @@ final class Schema {
           description: description,
           title: title,
           nullable: nullable,
+          minProperties: minProperties,
+          maxProperties: maxProperties,
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema for an array of values with a specified type.
@@ -63,6 +79,8 @@ final class Schema {
     bool? nullable,
     int? minItems,
     int? maxItems,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.array,
           description: description,
@@ -71,6 +89,8 @@ final class Schema {
           items: items,
           minItems: minItems,
           maxItems: maxItems,
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema for bool value.
@@ -78,11 +98,15 @@ final class Schema {
     String? description,
     String? title,
     bool? nullable,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.boolean,
           description: description,
           title: title,
           nullable: nullable,
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema for an integer number.
@@ -95,6 +119,8 @@ final class Schema {
     String? format,
     int? minimum,
     int? maximum,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.integer,
           description: description,
@@ -103,6 +129,8 @@ final class Schema {
           format: format,
           minimum: minimum?.toDouble(),
           maximum: maximum?.toDouble(),
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema for a non-integer number.
@@ -115,6 +143,8 @@ final class Schema {
     String? format,
     double? minimum,
     double? maximum,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.number,
           description: description,
@@ -123,6 +153,8 @@ final class Schema {
           format: format,
           minimum: minimum,
           maximum: maximum,
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema for String value with enumerated possible values.
@@ -131,6 +163,8 @@ final class Schema {
     String? description,
     String? title,
     bool? nullable,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.string,
           enumValues: enumValues,
@@ -138,6 +172,8 @@ final class Schema {
           title: title,
           nullable: nullable,
           format: 'enum',
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema for a String value.
@@ -146,12 +182,22 @@ final class Schema {
     String? title,
     bool? nullable,
     String? format,
+    int? minLength,
+    int? maxLength,
+    String? pattern,
+    Object? example,
+    Object? defaultValue,
   }) : this(
           SchemaType.string,
           description: description,
           title: title,
           nullable: nullable,
           format: format,
+          minLength: minLength,
+          maxLength: maxLength,
+          pattern: pattern,
+          example: example,
+          defaultValue: defaultValue,
         );
 
   /// Construct a schema representing a value that must conform to
@@ -178,6 +224,12 @@ final class Schema {
   }) : this(
           SchemaType.anyOf, // The type will be ignored in toJson
           anyOf: schemas,
+        );
+
+  /// Construct a schema referencing another schema.
+  Schema.ref(String ref) : this(
+          SchemaType.ref,
+          ref: ref,
         );
 
   /// The type of this value.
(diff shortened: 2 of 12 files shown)


packages/firebase_ai/firebase_ai/example/ios/Podfile
```diff
@@ -29,7 +29,6 @@ flutter_ios_podfile_setup
 
 target 'Runner' do
   use_frameworks!
```
a-maurice, line 31: Just to double check, are all of these project changes meant to be included?

a-maurice (approved):

packages/firebase_ai/firebase_ai/example/ios/Podfile
cynthiajoan: I meant to have those changes but they don't have to be included in this PR, I will remove them

packages/firebase_ai/firebase_ai/lib/src/schema.dart
```diff
@@ -35,6 +35,14 @@ final class Schema {
...
     this.propertyOrdering,
     this.anyOf,
+    this.ref,
+    this.minProperties,
+    this.maxProperties,
+    this.minLength,
+    this.maxLength,
+    this.pattern,
+    this.example,
+    this.defaultValue,
```
andrewheard: Have we tested that the model supports these fields? I can't find Vertex AI's documentation for JSON Schema and the Gemini Developer API does not [list them](https://ai.google.dev/gemini-api/docs/structured-output?example=recipe#json_schema_support).

Notes:
- These fields have been added to `Schema` (OpenAPI) but, without testing, I'm unsure if the Gemini actually uses them.
  - I'm unsure if the Gemini supports these in JSON Schema format since they aren't in the supported list and the API documentation doesn't [list them](https://ai.google.dev/api/generate-content#:~:text=Specifically%2C%20only%20the%20following%20properties%20are%20supported%3A).
  - It should be safe to add them since [`While the full JSON Schema may be sent, not all features are supported.`](https://ai.google.dev/api/generate-content#:~:text=While%20the%20full%20JSON%20Schema%20may%20be%20sent%2C%20not%20all%20features%20are%20supported.)
    - It may be somewhat misleading to developers if Gemini ignores them though.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
```diff
@@ -35,6 +35,14 @@ final class Schema {
     this.optionalProperties,
     this.propertyOrdering,
     this.anyOf,
+    this.ref,
```
andrewheard: We should add [`defs`](https://docs.cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.cachedContents#Schema:~:text=json%2Dschema/structuring-,defs,in%20the%20schema.%20Only%20allowed%20at%20root%20level%20of%20the%20schema.,-JSON%20representation) since `ref` could only reference self (`#`) without it.

Notes:
- In JSON Schema format it would need to use the key `"$defs"`, rather than `"defs"`.
- The Gemini API docs don't list `$defs` or `$ref` as being supported but they contradict themselves by showing it in the [Recursive Structures](https://ai.google.dev/gemini-api/docs/structured-output?example=recursive#rest_1) example.
  - We're using `$defs` or `$ref` heavily in the Swift SDK and they're working well. It seems unlikely that they're only working by chance (i.e., I don't think Gemini is ignoring them).

packages/firebase_ai/firebase_ai/lib/src/schema.dart
```diff
@@ -287,6 +367,59 @@ final class Schema {
...
+        if (example case final example?) 'example': example,
+        if (defaultValue case final defaultValue?) 'default': defaultValue,
+      };
+
+  /// Convert to standard JSON Schema object.
+  ///
+  /// Reference: https://ai.google.dev/api/caching#FunctionDeclaration
+  Map<String, Object> toJSONSchemaJson() => {
+        if (type != SchemaType.anyOf && type != SchemaType.ref)
+          'type': nullable == true ? [type.name, 'null'] : type.name,
```
andrewheard, line 505: `type.name` would be lowercase, right? Seems to be, based on the unit tests.

packages/firebase_ai/firebase_ai/test/schema_test.dart
```diff
@@ -253,10 +253,96 @@ void main() {
...
           'null'); // As per implementation, 'null' string for anyOf
     });
 
+    // Test Schema.ref
+    test('Schema.ref', () {
+      final schema = Schema.ref('#/components/schemas/User');
+      expect(schema.type, SchemaType.ref);
+      expect(schema.ref, '#/components/schemas/User');
+      expect(schema.toJson(), {
+        r'$ref': '#/components/schemas/User',
```
andrewheard, line 301: IIUC, `toJson()` returns the OpenAPI Schema format. In that case, it needs to be [`ref`](https://docs.cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.cachedContents#Schema:~:text=to%20the%20schema.-,ref,string,-Optional.%20Allows%20referencing).

Note: In OpenAPI format, `defs` and `ref` only exist in Vertex AI.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
cynthiajoan: I did a rough verification and seems even when those fields exist in the proto definition of Schema, model is not respecting them. Good catch and I will remove them.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
cynthiajoan: yes it would be lower case in JSON schema, but upper case in OAI schema

packages/firebase_ai/firebase_ai/lib/src/schema.dart
cynthiajoan: Added

packages/firebase_ai/firebase_ai/test/schema_test.dart
cynthiajoan: updated with new JSONSchema class

packages/firebase_ai/firebase_ai/test/schema_test.dart
```diff
@@ -147,6 +147,40 @@ void main() {
...
+        properties: properties,
+        defs: defs,
+      );
+      expect(schema.type, SchemaType.object);
+      expect(schema.properties, properties);
+      expect(schema.defs, defs);
+      expect(schema.toJson(), {
+        'type': 'object',
+        'properties': {
+          'metadata': {r'$ref': '#/metadata_schema'},
```
andrewheard, line 169: I'm not certain that `#/metadata_schema` would resolve to `metadata_schema` since it's nested in `$defs`. Since we can't reference any external schemas, I wonder if it would be safer to standardize on `#/$defs/SomeName` (so `#/$defs/metadata_schema` in this case).

It's definitely possible that there's a shorthand notation that I don't know about but I know that `#` works for self-referencing and `#/$defs/schema_name` works. In [this example](https://json-schema.org/understanding-json-schema/structuring#defs):
> Sometimes we have small subschemas that are only intended for use in the current schema and it doesn't make sense to define them as separate schemas. Although we can identify any subschema using JSON Pointers or named anchors, the $defs keyword gives us a standardized place to keep subschemas intended for reuse in the current schema document.

`name` is a subschema within the schema and it's referenced with, e.g., `"first_name": { "$ref": "#/$defs/name" }`.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
```diff
@@ -290,6 +290,255 @@ final class Schema {
...
+  Map<String, JSONSchema>? defs;
+
+  /// Convert to standard JSON JSONSchema object.
+  ///
+  /// Reference: https://ai.google.dev/api/caching#FunctionDeclaration
+  @override
+  Map<String, Object> toJson() => {
+        if (type != SchemaType.anyOf && type != SchemaType.ref)
+          'type': nullable == true ? [type.name, 'null'] : type.name,
+        if (ref case final ref?) r'$ref': ref,
```
andrewheard, line 506: We may want to consider prefixing `ref` with `#/$defs/`. See the [comment below](https://github.com/firebase/flutterfire/pull/18131/changes#r2984708040) for more details.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
```diff
@@ -290,6 +290,255 @@ final class Schema {
       };
 }
 
+/// The definition of a JSON Schema data type.
+///
+/// This class supports `$ref` and `$defs` for reusable sub-schemas.
+final class JSONSchema extends Schema {
+  // ignore: public_member_api_docs
+  JSONSchema(
```
andrewheard, line 298: Note: I'm not certain what Gemini's behaviour is if [`additionalProperties`](https://ai.google.dev/gemini-api/docs/structured-output?example=recipe#type-specific_properties:~:text=additionalProperties%3A%20Controls%20whether%20properties%20not%20listed%20in%20properties%20are%20allowed.%20Can%20be%20a%20boolean%20or%20a%20schema.) is omitted. The `@Generable` macro aways produces a schema with `"additionalProperties" : false` so that's all I've tested.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
```diff
@@ -290,6 +290,255 @@ final class Schema {
...
+        );
+
+  /// Construct a schema for a non-integer number.
+  ///
+  /// The [format] may be "float" or "double".
+  JSONSchema.number({
+    String? description,
+    String? title,
+    bool? nullable,
+    String? format,
```
andrewheard: As far as I'm aware `format` only applies to [string values](https://www.learnjsonschema.com/2020-12/format-annotation/format/). This aligns with the Gemini [documentation](https://ai.google.dev/gemini-api/docs/structured-output?example=recipe#type-specific_properties:~:text=format%3A%20Specifies%20a,maximum%20inclusive%20value.) as well (only listed for `string`).

cynthiajoan: good catch. But for clarification, the oai integer and number supports format? They stay under the Schema class from the beginning.

packages/firebase_ai/firebase_ai/test/schema_test.dart
cynthiajoan: in the example you linked, what does the address it resolves to? 
`"shipping_address": { "$ref": "/schemas/address" }`

packages/firebase_ai/firebase_ai/lib/src/schema.dart
cynthiajoan: it says "Can be a boolean or a schema.", confused of what data structure I should use if I add it.

packages/firebase_ai/firebase_ai/test/schema_test.dart
andrewheard: It resolves to `https://example.com/schemas/address` (due to `"$id": "https://example.com/schemas/customer"` as the Base URI for the schema). The tutorial starts from [Base URI](https://json-schema.org/understanding-json-schema/structuring#base-uri) on that page.

Note: None of the external schema stuff is supported by Gemini, as far as I'm aware though.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
andrewheard: I'd recommend just using a boolean if it can be changed later, or omitting it if we find there are no problems. `"additionalProperties" : false` is essentially saying "don't generate any properties that I didn't ask for."

packages/firebase_ai/firebase_ai/test/schema_test.dart
cynthiajoan: That explains, in that case I should add the #/$defs/, and add some extra documentation in the api usage.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
andrewheard: Yep, in OAPI format it does support it for both integer and number.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
cynthiajoan: During my testing I don't have it and the model seems fine with that. So I would omit for now

packages/firebase_ai/firebase_ai/lib/src/schema.dart
cynthiajoan: Removed

packages/firebase_ai/firebase_ai/test/schema_test.dart
cynthiajoan: but after making that change, I find it doesn't work with this[ test case](https://github.com/firebase/flutterfire/pull/18131/changes#diff-0c50d71fcabefdb2ccfed8e92039e8c4cd2e34d85176408ec268cf8a655bf25eR173). If i make it points to `#/$defs/transactionDef`, the model will complain `reference to undefined schema at properties.transactionsBlock.properties.transactionsList.items`

packages/firebase_ai/firebase_ai/lib/src/schema.dart
andrewheard: Sounds good to me. I don't see it in the Gemini REST examples either.

packages/firebase_ai/firebase_ai/test/schema_test.dart
andrewheard: In the test `defs` is nested inside a `transactionsBlock` object, rather than the top-level object (if I'm reading Dart correctly) that `#` refers to, so it makes sense that it can't find it at `#/$defs/transactionDef`. Unfortunately I've only tested examples where `$defs` are defined in the top-level object.

It's good to know that the errors that Gemini returns explicitly mention that it can't find the references.

packages/firebase_ai/firebase_ai/lib/src/schema.dart
andrewheard: Feel free to ignore this. Seems reasonable to allow devs to enter any value if your test case below is working as-is. I assumed Gemini only supported top-level defs but seems like they can be nested wherever.

andrewheard (approved):

packages/firebase_ai/firebase_ai/test/schema_test.dart
cynthiajoan: Yup that's the case. It's actually nice that gemini supports the defs in any nested level

Lyokone (approved):

PR #18151, opened 2026-03-27
feat(fdc): Happy Path Implementation

aashishpatil-g:
Inital implementation of happy path.


packages/firebase_data_connect/firebase_data_connect/example/test_stream.dart
```diff
@@ -0,0 +1,68 @@
...
+  );
+
+  final queryName = 'ListBlogs';
+  
+  // Execute via REST for comparison
+  print('Trying to perform a standard unary executeQuery...');
+  try {
+    final response = await transport.invokeQuery<dynamic, dynamic>(
+      queryName,
+      (String s) => s, // dummy deserializer
```
dconeybe: nit: "dummy" is considered a "non-inclusive" term. Consider using another term, like "identity" or "placeholder", instead.

https://www.egalresources.com/inclusive-terms-glossary#Dummy

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/result_tree_processor.dart
```diff
@@ -29,7 +29,7 @@ class DehydrationResult {
 class ResultTreeProcessor {
   /// Takes a server response, traverses the data, creates or updates `EntityDataObject`s,
   /// and builds a dehydrated `EntityNode` tree.
-  Future<DehydrationResult> dehydrateResults(
+  Future<DehydrationResult>  dehydrateResults(
```
dconeybe: This appears to be an unintentional whitespace change. Do you need to run a code formatter on the PR?

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -60,8 +60,16 @@ abstract class OperationRef<Data, Variables> {
 
   FirebaseDataConnect dataConnect;
 
-  Future<OperationResult<Data, Variables>> execute(
-      {QueryFetchPolicy fetchPolicy = QueryFetchPolicy.preferCache});
+  static String createOperationId<Variables>(String operationName,
+      Variables? vars, Serializer<Variables>? serializer) {
+    if (vars != null && serializer != null) {
+      return '$operationName::${serializer(vars)}';
```
dconeybe: Is the serializer guaranteed to traverse the keys of the variables in a consistent ordering? I ask because if this "operation ID" is persisted into the cache database then you need to guarantee that the ordering of the output of the serializer is always the same. For example, in Kotlin I wrote a custom Struct traversal algorithm that always sorts the keys before traversing them. Do you need to do something like that here as well?

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -60,8 +60,16 @@ abstract class OperationRef<Data, Variables> {
 
   FirebaseDataConnect dataConnect;
 
-  Future<OperationResult<Data, Variables>> execute(
-      {QueryFetchPolicy fetchPolicy = QueryFetchPolicy.preferCache});
+  static String createOperationId<Variables>(String operationName,
+      Variables? vars, Serializer<Variables>? serializer) {
```
dconeybe, line 78: nit: There appears to be a non-sensical case here where `vars` can be non-null but `serializer` is null, leading to the `vars` being ignored. Is there an idiomatic way in Dart to eliminate this possibility all together? Or, is it indeed a valid case?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
```diff
@@ -0,0 +1,174 @@
+// Copyright 2024 Google LLC
```
dconeybe: nit: Looks like the wrong copyright year. Perhaps scan all of the new files you've added and ensure they're using `2026` as the copyright year.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,391 @@
...
+  /// Initializes necessary protocol and port.
+  WebSocketTransport(
+    this.transportOptions,
+    this.options,
+    this.appId,
+    this.sdkType,
+    this.appCheck, [
+    this.auth,
+  ]) {
+    String protocol = 'ws';
```
dconeybe: ultranit: In google3 there is general guidance to avoid assign-then-mutate pattern for variables initialization (go/python-tips/026). Consider changing this to initialize protocol exactly once, something like this:

```dart
String protocol = (transportOptions.isSecure ?? true) ? 'wss' : 'ws';
```

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,391 @@
...
+    }
+    String host = transportOptions.host;
+    int port = transportOptions.port ?? 443;
+    String location = options.location;
+    _url = '$protocol://$host:$port/v1/Connect/locations/$location';
+    
+    _currentUid = auth?.currentUser?.uid;
+    _authSubscription = auth?.idTokenChanges().listen((user) async {
+      final newUid = user?.uid;
+      // Don't disconnect if auth state changes from not logged in to logged in.
```
dconeybe, line 44: Note: The updated design does not allow transitioning for/from unauthenticated. This code will need to be updated accordingly.

dconeybe (approved):
LGTM with some comments.

packages/firebase_data_connect/firebase_data_connect/example/test_stream.dart
aashishpatil-g: Great feedback. This was a test file thats been removed.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/result_tree_processor.dart
aashishpatil-g: Done. Ran both analyzer and formatter.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
aashishpatil-g: That case is not valid - I think both were kept as optional because serializer can be null when vars is null.

The internal check only uses the serializer if both vars and serializer are not-null. So we don't hit that case.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
aashishpatil-g: done

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
aashishpatil-g: fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
aashishpatil-g: I will fix this in the proceeding PR that handles disconnects.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
aashishpatil-g: This is good feedback. I do that in Swift too with a .sortedKeys option. 

Dart doesn't seem to have a built in option. JSON retains the insertion order so thats what we have been relying on. Anyway, I've added a sortedKeys function now.

aashishpatil-g:
Failing tasks are not related to Data Connect package

PR #18173, opened 2026-04-06
feat(fdc): Hardening 

aashishpatil-g:
- Handle QueryRef cancellations
- Additional debug statements
- Fix: only create one multicast controller per QueryRef


stephenarosaj (approved):
LGTM.

PR #18174, opened 2026-04-06
feat(fdc): Streaming implementation for data connect

aashishpatil-g:
Implements streaming support in Firebase Data Connect for subscribe().


packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
```diff
@@ -36,9 +36,7 @@ class DataConnectError extends FirebaseException {
 
 /// Error thrown when an operation is partially successful.
 class DataConnectOperationError<T> extends DataConnectError {
-  DataConnectOperationError(
-      DataConnectErrorCode code, String message, this.response)
-      : super(code, message);
+  DataConnectOperationError(super.code, String super.message, this.response);
```
cynthiajoan: We don't need the extra String declaration for super.message

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2026 Google LLC
```
cynthiajoan: we usually keep the original year for existing files, don't need to update when there's modification.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -53,15 +53,46 @@ abstract class OperationRef<Data, Variables> {
   );
   Variables? variables;
   String operationName;
-  DataConnectTransport _transport;
+  final DataConnectTransport _transport;
```
cynthiajoan, line 56: why only change this to final?

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -53,15 +53,46 @@ abstract class OperationRef<Data, Variables> {
   );
   Variables? variables;
```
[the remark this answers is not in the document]
aashishpatil-g: fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
aashishpatil-g: Fixed. Since it was new, the old vars were not touched. But all should be final. I've made the change and verified there is no impact.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
aashishpatil-g: reverted

packages/firebase_data_connect/firebase_data_connect/lib/src/common/dataconnect_error.dart
aashishpatil-g: fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,653 @@
...
+          for (final controller in controllers) {
+            controller.add(serverResponse);
+          }
+        }
+      }
+    } catch (e) {
+      // JSON decoding error or unknown format
+      developer.log('error decoding server response $e');
+    }
+  }
```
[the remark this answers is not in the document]
aashishpatil-g: Taking this as a tech debt.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
```diff
@@ -0,0 +1,174 @@
...
+  final ResumeRequest? resume;
+
+  /// Signal that the client is no longer interested.
+  final bool? cancel;
+
+  /// Etag for caching.
+  final String? dataEtag;
+
+  Map<String, dynamic> toJson() {
+    final Map<String, dynamic> data = <String, dynamic>{};
```
cynthiajoan, line 91: since all the fields are optional, is it possible the json blob being empty? is it a valid case? is server going to error out?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
```diff
@@ -0,0 +1,174 @@
...
+    return data;
+  }
+}
+
+/// Request to resume a query.
+class ResumeRequest {
+  ResumeRequest();
+
+  Map<String, dynamic> toJson() {
+    return <String, dynamic>{};
```
cynthiajoan, line 45: this behave same as cancel, are we reserve this for future possible variables?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
```diff
@@ -0,0 +1,174 @@
...
+    this.cancelled,
+    this.extensions,
+  });
+
+  factory StreamResponse.fromJson(Map<String, dynamic> json) {
+    if (json.containsKey('result')) {
+      json = json['result'] as Map<String, dynamic>;
+    } else if (json.containsKey('error')) {
+      final errObj = json['error'] as Map<String, dynamic>;
+      json = {
```
cynthiajoan, line 135: is there going to be multiple errors in the original json? is this going to override if error and errors both exist?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_library.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2026 Google LLC
```
cynthiajoan: don't need to change the year

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_stub.dart
```diff
@@ -60,6 +60,19 @@ class TransportStub implements DataConnectTransport {
...
+  /// Stub for subscribing to a query.
+  @override
+  Stream<ServerResponse> invokeStreamQuery<Data, Variables>(
+    String queryName,
+    Deserializer<Data> deserializer,
+    Serializer<Variables>? serializer,
+    Variables? vars,
+    String? token,
+  ) {
+    // TODO: implement invokeStreamQuery
```
cynthiajoan, line 74: is this TODO valid?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_stub.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2026 Google LLC
```
cynthiajoan: no need to change the year

packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
```diff
@@ -20,10 +20,8 @@ import 'package:firebase_data_connect/src/common/common_library.dart';
 import 'package:firebase_data_connect/src/core/ref.dart';
 import 'package:flutter/foundation.dart';
 
-import './network/transport_library.dart'
-    if (dart.library.io) './network/grpc_library.dart'
```
cynthiajoan, line 24: are these platform specific import not valid anymore?

packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
```diff
@@ -93,13 +91,22 @@ class FirebaseDataConnect extends FirebasePluginPlatform {
...
+    final ws = WebSocketTransport(
+      transportOptions!,
+      options,
+      app.options.appId,
+      _sdkType,
+      appCheck,
+      auth,
+    );
+    transport = _RoutingTransport(rest, ws);
   }
```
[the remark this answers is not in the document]
aashishpatil-g: This seems to have reverted - had been fixed in an earlier commit. 

Now fixed.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,653 @@
...
+            requestId: _generateRequestId('auth'),
+            authToken: token,
+          );
+          _channel?.sink.add(jsonEncode(request.toJson()));
+        } catch (_) {
+          // Ignored
+        }
+      }
+      _currentUid = newUid;
+    });
```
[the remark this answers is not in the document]
aashishpatil-g: Taking this as tech debt. This is a bit tricky to implement.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -307,22 +334,107 @@ class QueryRef<Data, Variables> extends OperationRef<Data, Variables> {
...
+    _serverStream = null;
+    log("QueryRef $_queryId: All subscribers cancelled. Unsubscribed from server stream.");
+  }
 
   Stream<QueryResult<Data, Variables>> subscribe() {
     _streamController ??= _queryManager.addQuery(this);
 
-    execute();
+    final stream =
+        _streamController!.stream.cast<QueryResult<Data, Variables>>();
```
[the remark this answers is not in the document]
aashishpatil-g: the flow gurantees a streamcontroller. ignoring.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -53,15 +53,46 @@ abstract class OperationRef<Data, Variables> {
...
+    }
+    return value;
+  }
+
+  static String createOperationId<Variables>(String operationName,
+      Variables? vars, Serializer<Variables>? serializer) {
+    if (vars != null && serializer != null) {
+      try {
+        final decoded = jsonDecode(serializer(vars));
+        final sortedStr = jsonEncode(_sortKeys(decoded));
```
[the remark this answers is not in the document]
aashishpatil-g: Fixed this. 

operationId is now lazily computed in the base OperationRef class and used everywhere instead of computing each time. Also, removed the _queryId and instead use operationId, which is valid for both queries and mutations.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,653 @@
...
+
+      // Append errors if any exist on the stream payload
+      if (response.errors != null && response.errors!.isNotEmpty) {
+        // We simulate a DataConnectOperationError payload structure
+        // so that ref.dart can parse it correctly
+        serverResponse.data['errors'] = response.errors;
+      }
+
+      if (_unaryListeners.containsKey(requestId)) {
+        final pendings = _unaryListeners.remove(requestId)!;
```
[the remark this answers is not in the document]
aashishpatil-g: fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,653 @@
...
+        final headers = _buildHeaders(authToken, appCheckToken);
+
+        final request = StreamRequest(
+          authToken: authToken,
+          appCheckToken: appCheckToken,
+          requestId: requestId,
+          requestKind: RequestKind.subscribe,
+          subscribe: ExecuteRequest(queryName, variables),
+          headers: headers,
+        );
```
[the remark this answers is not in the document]
aashishpatil-g: fixed.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_stub.dart
aashishpatil-g: Yes, this is to satisfy protocol requirements for the existing stub class which also implements the DataConnectTransport protocol.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_stub.dart
aashishpatil-g: reverted

packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
aashishpatil-g: Yes, grpc transport has been deprecated internally. I'll remove the code in a later version but for now its not used for transport. 

All platforms either use RestTransport (non-streaming) or WebSocketTransport (streaming).

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
aashishpatil-g: Both this and cancel are used within a containing StreamRequest object. 

It depends on which key in the StreamRequest is set and the request id. 

If cancel key is set, then the requestId is considered cancelled
If resume key is set, then requestId is resumed => results are republished on the stream.

aashishpatil-g: Yes, the maps have been still kept in case of future changes to a resume or cancel request.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
aashishpatil-g: I think server ignores the message but I've updated the code to check before sending an empty message to server.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_stub.dart
aashishpatil-g: I am not sure of the origin of this stub class. It was preexisting.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
aashishpatil-g: Sorry - good catch. I think this is from some earlier versions of the protocol. The errors field is within the result / peer of data field. So this isn't necessary. Removed it.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/transport_library.dart
aashishpatil-g: reverted

packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
aashishpatil-g: I just removed the grpc transport files. It was not used or referenced anywhere anymore.

cynthiajoan (approved):

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -51,7 +51,7 @@ class Cache {
 
   String _constructCacheIdentifier() {
     final rawPrefix =
-        '${_settings.storage}-${dataConnect.app.options.projectId}-${dataConnect.app.name}-${dataConnect.connectorConfig.serviceId}-${dataConnect.connectorConfig.connector}-${dataConnect.connectorConfig.location}-${dataConnect.transport.transportOptions.host}';
+        '${_settings.storage}-${dataConnect.app.options.projectId}-${dataConnect.app.name}-${dataConnect.connectorConfig.serviceId}-${dataConnect.connectorConfig.connector}-${dataConnect.connectorConfig.location}-${dataConnect.transport?.transportOptions.host}';
```
stephenarosaj, line 54: is there any chance for transport to be null? because if so, i believe this string could end like `storage-project-name-service-connector-null`

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -175,24 +209,20 @@ class QueryManager {
...
-    final queryId = ref._queryId;
+    final queryId = ref.operationId;
     trackedQueries[queryId] = ref;
 
     final streamController =
-        StreamController<QueryResult<Data, Variables>>.broadcast();
+        StreamController<QueryResult<Data, Variables>>.broadcast(
+      onCancel: () {
+        trackedQueries.remove(queryId);
+        ref._onAllSubscribersCancelled();
```
stephenarosaj, line 219: just want to confirm my understanding - is this adding to the entire stream's cleanup, saying "make sure you clean up this tracked query too" ?

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
```diff
@@ -307,22 +335,109 @@ class QueryRef<Data, Variables> extends OperationRef<Data, Variables> {
...
+        operationId,
+        operationName,
+        deserializer,
+        serializer,
+        variables,
+        _lastToken,
+      );
 
-    return _streamController!.stream.cast<QueryResult<Data, Variables>>();
+      _serverStreamSubscription = _serverStream!.listen(
```
stephenarosaj, line 387: nit-suggestion: extract the anonymous functions here out to dedicated handler functions that you then pass to the `listen()` call

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  /// The resource name of the connector.
+  final String? name;
+
+  /// Optional headers.
+  final Map<String, String>? headers;
+
+  /// Optional Auth token.
+  final String? authToken;
```
stephenarosaj: is the `authToken` a top-level field? shouldn't it be inside of `headers`, and as `X-Firebase-Auth-Token` - same for app check token ?

aashishpatil-g:
The test failures are due to some auth sdk errors. I'll need to rebase on main which has some auth sdk fixes. But didn't want to disturb the ongoing reviews. 

EDIT: 
Fixed with rebase

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,646 @@
...
+      _isExpectedDisconnect = true;
+      _disconnect();
+      _clearState();
+    }
+  }
+
+  final Random _random = Random();
+  static const String _chars = 'abcdefghijklmnopqrstuvwxyz0123456789';
+
+  String _generateRequestId(String operationName) {
```
stephenarosaj, line 143: is there a reason for a random ID instead of one that's monotonically increasing?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
```diff
@@ -0,0 +1,160 @@
...
+    this.requestId,
+    this.data,
+    this.dataEtag,
+    this.errors,
+    this.cancelled,
+    this.extensions,
+  });
+
+  factory StreamResponse.fromJson(Map<String, dynamic> json) {
+    if (json.containsKey('result')) {
```
stephenarosaj, line 131: currently on the web, if we receive a malformed message (no result, no requestId, etc.) we terminate the connection. i figured that was the right choice since that indicates some kind of bigger issue in our communication with the server, and we should try reconnecting if we want to keep communicate via stream

we should probably align our SDKs one way or the other. thoughts on which way and why?

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
```diff
@@ -0,0 +1,646 @@
...
+
+  // Active listeners for stream subscriptions mapped by requestId.
+  final Map<String, List<StreamController<ServerResponse>>> _streamListeners =
+      {};
+
+  // Pending information for subscriptions mapped by requestId.
+  final Map<String, _PendingSubscription> _pendingSubscriptions = {};
+
+  // Active completers for unary operations mapped by requestId.
+  final Map<String, List<_PendingUnary>> _unaryListeners = {};
```
stephenarosaj, line 123: at some point in the future, we should align our method and field names across platforms

packages/firebase_data_connect/firebase_data_connect/.metadata
```diff
@@ -0,0 +1,30 @@
+# This file tracks properties of this Flutter project.
```
stephenarosaj: just checking - should this file be added? it does say to be version controlled - but it wasn't in the previous version

stephenarosaj (changes requested):
LGTM - requesting changes but really just looking for a few responses / ACK on some comments

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
aashishpatil-g: This is in the QueryRef -  the cancel is called when the QueryRef doesn't have any more subscribers i.e. no more Views subscribing to this QR. 

In Cancel, we tell the FDC instance to remove the streamcontroller for this query and also tell the websocket transport that it can send a cancel for this QueryRef.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
aashishpatil-g: There isn't a way for transport to be null. I am just following a best practice to not coerce the optional.  I had to keep transport an optional in the FDC class to allow for a nullability check to not recreate a transport instance each time.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
aashishpatil-g: Trying to reduce chances of guessing a sequence.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
aashishpatil-g: This is fixed. The field has been removed. Thanks for catching.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/stream_protocol.dart
aashishpatil-g: I think lets chat with the server team on the possibility of this. 

If there isn't an error. I felt, we should keep the stream open. Its possible there was a temporary glitch and maybe teardown and reconnect is more expensive. 

Either way lets align on the responses and error handling based on the type of error.

packages/firebase_data_connect/firebase_data_connect/lib/src/network/websocket_transport.dart
aashishpatil-g: Agreed - tech debt.

aashishpatil-g: FYI - I borrowed the grpc naming unary vs stream.

packages/firebase_data_connect/firebase_data_connect/.metadata
aashishpatil-g: Good point. Took it out. I think I might have mistakenly tried to run a Flutter app from this folder, which caused the issue.

packages/firebase_data_connect/firebase_data_connect/lib/src/core/ref.dart
aashishpatil-g: Good point. I'll track this for next release.

stephenarosaj (approved):

Lyokone (approved):

PR #18219, opened 2026-04-21
chore: update example dependencies across repo

kevmoo:
including fixing the latest lints


cynthiajoan (approved):

kevmoo:
@cynthiajoan – zero clue why CI is failing. My changes should affect those e2e tests.

Happy to try to do a merge from `main` and see what happens 🤷

cynthiajoan:
> @cynthiajoan – zero clue why CI is failing. My changes should affect those e2e tests.
> 
> Happy to try to do a merge from `main` and see what happens 🤷

I think those CI fails aren't related. We would need to do some work to improve the flakiness

packages/firebase_auth/firebase_auth/ios/firebase_auth/Package.swift
```diff
@@ -33,8 +33,8 @@ let package = Package(
         .process("Resources"),
       ],
       cSettings: [
-        .headerSearchPath("include/firebase_auth/Private"),
```
kevmoo, line 36: This was an unrelated fix, but it did improve test pass rate! @cynthiajoan 

See below, too!

cynthiajoan: I did notice this problem, and I was wondering if other packages have the same issue, haven't checked. @Lyokone mentioned he's going to check that.

russellwheatley: I've created a PR to ensure header search paths are correct: https://github.com/firebase/flutterfire/pull/18239

I'm not quite convinced this would fix anything broken as the source imports in FLTFirebaseAuthPlugin.m already use explicit quoted paths (See https://github.com/firebase/flutterfire/blob/update-search-paths/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m#L14-L17), which would allow imports to resolve directly without relying on broken `headerSearchPath `.

Lyokone (approved):

Lyokone:
Merging since failing CI tests have been fixed on main and are unrelated 👍 Thanks :)

PR #18234, opened 2026-04-28
chore: fix melos bs

Lyokone:
`melos bootstrap` was failing with `Cannot override workspace packages` because melos was generating per-package `pubspec_overrides.yaml` files


SelaseKay (approved):

cynthiajoan (approved):
Not intend to block, but curious about why we need to add the Podfile for app check package?

Lyokone:
@cynthiajoan It was regenerated automatically when running melos bs. I'll delete this it's not directly related to this PR :)

PR #18255, opened 2026-05-06
fix(storage,android): fix an issue that could happen when app would get detached from the engine

### LINKED ISSUE #18240: 🐛 [firebase_storage] ArrayIndexOutOfBoundsException in cancelInProgressTasks (self-modifying loop on engine detach)
## Bug report

`FlutterFirebaseStorageTask.cancelInProgressTasks()` (Android) crashes the host process with `ArrayIndexOutOfBoundsException` whenever a `FlutterEngine` is detached while three or more `FlutterFirebaseStorageTask`s are in `inProgressTasks`. Reproduces on `firebase_storage` **13.2.0** and **13.3.0** (latest).

This is a different code path from the closed issues #9830, #8823, #4334. Those addressed races in `<init>` / `SparseArray.put`. The bug here is in the cancellation loop and is purely single-threaded — the lock doesn't help because the iteration mutates what it's iterating.

### Steps to reproduce

1. App has any setup that produces ≥3 simultaneous entries in the static `FlutterFirebaseStorageTask.inProgressTasks` SparseArray. Easiest in practice: a foreground engine doing concurrent uploads while a `workmanager` (or any other) background isolate's engine completes and is detached.
2. The detach triggers `FlutterFirebaseStoragePlugin.onDetachedFromEngine` → `FlutterFirebaseStorageTask.cancelInProgressTasks()`.
3. Process crashes.

Because `inProgressTasks` is a static companion-object field, **every engine in the JVM contributes to the same list**. A background isolate detaching will try to cancel uploads owned by the foreground engine.

### Expected behavior

Detaching an engine should not crash the host app, regardless of how many uploads are in flight in any engine.

### Actual behavior / stack trace

\`\`\`
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: Array index out of range: 2
    at android.util.SparseArray.valueAt(SparseArray.java:347)
    at io.flutter.plugins.firebase.storage.FlutterFirebaseStorageTask\$Companion.cancelInProgressTasks(FlutterFirebaseStorageTask.java:99)
    at io.flutter.plugins.firebase.storage.FlutterFirebaseStoragePlugin.onDetachedFromEngine(FlutterFirebaseStoragePlugin.kt:40)
    at io.flutter.embedding.engine.FlutterEngineConnectionRegistry.remove(FlutterEngineConnectionRegistry.java:272)
    ...
    at io.flutter.embedding.engine.FlutterEngine.destroy(FlutterEngine.java:510)
    at dev.fluttercommunity.workmanager.BackgroundWorker.stopEngine\$lambda\$0(BackgroundWorker.java:187)
\`\`\`

The numeric "index out of range" matches `(initial size of inProgressTasks) - 1` in every crash report I have. Sample crashes in production: index 2 with 3 active uploads, consistently.

### Root cause

`packages/firebase_storage/firebase_storage/android/src/main/kotlin/io/flutter/plugins/firebase/storage/FlutterFirebaseStorageTask.kt`, lines 95–104:

\`\`\`kotlin
@JvmStatic
fun cancelInProgressTasks() {
  synchronized(inProgressTasks) {
    for (i in 0 until inProgressTasks.size()) {
      val task: FlutterFirebaseStorageTask? = inProgressTasks.valueAt(i)
      task?.destroy()
    }
    inProgressTasks.clear()
  }
}
\`\`\`

`task.destroy()` (same file, lines 73–85) calls `inProgressTasks.remove(handle)`. Kotlin's `synchronized` is reentrant, so reacquiring the monitor inside `destroy()` does not prevent the mutation. The Kotlin range `0 until inProgressTasks.size()` evaluates `size()` **once** when the `IntRange` is constructed, so:

- start: size = 3, range = 0..2
- i=0 → `valueAt(0)`, then `destroy()` → size becomes 2
- i=1 → `valueAt(1)`, then `destroy()` → size becomes 1
- i=2 → `valueAt(2)` → AIOOBE

With 1 or 2 entries the loop happens to exit cleanly, which is why this only fires when the array has ≥3 entries at detach time.

### Proposed fix

Snapshot first, mutate after:

\`\`\`kotlin
@JvmStatic
fun cancelInProgressTasks() {
  synchronized(inProgressTasks) {
    val snapshot = ArrayList<FlutterFirebaseStorageTask>(inProgressTasks.size())
    for (i in 0 until inProgressTasks.size()) {
      inProgressTasks.valueAt(i)?.let { snapshot.add(it) }
    }
    inProgressTasks.clear()
    snapshot.forEach { it.destroy() }
  }
}
\`\`\`

`destroy()`'s subsequent `remove(handle)` calls become no-ops on the cleared array, which is fine.

### Environment

- Affected versions: `firebase_storage` 13.2.0, 13.3.0 (verified by reading source in pub cache)
- Devices observed: Moto G54 5G on Android 14 and Android 15
- Triggering setup: foreground app uploads + `workmanager` 0.9.0 background isolate; both call `firebase_storage`
- Plugin code path is the same on every Android version, so any `FlutterEngine` detach with ≥3 in-flight tasks should reproduce

Lyokone:
## Description

Fixes an Android crash in `firebase_storage` when Firebase Core reinitialization or engine detach cancels three or more in-progress Storage tasks. The fix snapshots the shared in-progress task list before destroying tasks so cancellation no longer mutates the collection being iterated.

## Related Issues

- Fixes https://github.com/firebase/flutterfire/issues/18240

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_storage/firebase_storage/android/src/main/kotlin/io/flutter/plugins/firebase/storage/FlutterFirebaseStorageTask.kt
+++ b/packages/firebase_storage/firebase_storage/android/src/main/kotlin/io/flutter/plugins/firebase/storage/FlutterFirebaseStorageTask.kt
@@ -95,11 +95,13 @@ internal class FlutterFirebaseStorageTask private constructor(
     @JvmStatic
     fun cancelInProgressTasks() {
       synchronized(inProgressTasks) {
+        val tasks = ArrayList<FlutterFirebaseStorageTask>(inProgressTasks.size())
         for (i in 0 until inProgressTasks.size()) {
           val task: FlutterFirebaseStorageTask? = inProgressTasks.valueAt(i)
-          task?.destroy()
+          task?.let { tasks.add(it) }
         }
         inProgressTasks.clear()
+        tasks.forEach { it.destroy() }
       }
     }
 
--- a/tests/integration_test/firebase_storage/task_e2e.dart
+++ b/tests/integration_test/firebase_storage/task_e2e.dart
@@ -6,9 +6,12 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:firebase_core/firebase_core.dart';
+import 'package:firebase_core_platform_interface/firebase_core_platform_interface.dart';
 import 'package:firebase_storage/firebase_storage.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:tests/firebase_options.dart';
 
 import './test_utils.dart';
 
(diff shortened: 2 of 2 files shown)


cabljac (approved):
lgtm

russellwheatley (approved):

PR #18256, opened 2026-05-06
fix(messaging,android): fix call race that could happen when using requestPermission

### LINKED ISSUE #18246: DuplicateTaskCompletionException crash in requestPermissions() on Android due to unsynchronized requestInProgress flag
## Description

Calling `FirebaseMessaging.instance.requestPermission()` from two code paths in quick succession crashes the app on Android with `DuplicateTaskCompletionException`. The crash is reproducible in production at meaningful scale and has not been addressed in any version through 16.2.0.

## Stack trace

```
java.lang.IllegalStateException: DuplicateTaskCompletionException
  at com.google.android.gms.tasks.TaskCompletionSource.setResult(...)
  at io.flutter.plugins.firebase.messaging.FlutterFirebaseMessagingPlugin
       .lambda$requestPermissions$10(FlutterFirebaseMessagingPlugin.java:371)
```

## Root cause

`requestPermissions()` in `FlutterFirebaseMessagingPlugin.java` (line 358) creates a new `TaskCompletionSource` per call and submits work to `cachedThreadPool`. Two calls can execute concurrently on separate threads.

The race is in `FlutterFirebasePermissionManager`:

```java
// Neither field is volatile or synchronized
@Nullable private RequestPermissionsSuccessCallback successCallback;
private boolean requestInProgress = false;
```

The guard at the top of `requestPermissions()`:

```java
if (requestInProgress) {
    errorCallback.onError("A request for permissions is already running...");
    return;
}
```

...is a non-atomic check-then-act on a non-`volatile` field. Two threads from `cachedThreadPool` can both read `requestInProgress == false` before either writes `true`, both passing the guard. From there:

- Both write `this.successCallback`, the second overwriting the first
- `ActivityCompat.requestPermissions()` may be called twice, delivering `onRequestPermissionsResult` twice
- Both completions attempt `taskCompletionSource.setResult()` on the same instance → `DuplicateTaskCompletionException`

## Suggested fix

The minimal fix is to make `requestInProgress` an `AtomicBoolean` with a compare-and-set:

```java
private final AtomicBoolean requestInProgress = new AtomicBoolean(false);

// In requestPermissions():
if (!requestInProgress.compareAndSet(false, true)) {
    errorCallback.onError("A request for permissions is already running...");
    return;
}
```

Alternatively, defensively guard the `TaskCompletionSource` completions with `trySetResult` / `trySetException` (which return `false` instead of throwing on an already-completed task):

```java
// Line 371
taskCompletionSource.trySetResult(permissions);

// Line 374
taskCompletionSource.trySetException(new Exception(errorDescription));
```

The `AtomicBoolean` fix is more correct (prevents the double OS dialog call). The `trySet*` fix is a safety net that prevents the crash even if the race occurs.

## Versions affected

Verified identical code in `firebase_messaging` 15.1.6 through 16.2.0. No fix or acknowledgment found in any changelog entry.

## Workaround

Cache the in-flight permission `Future` on the Dart side so only one call ever reaches the native layer:

```dart
Future<NotificationSettings>? _permissionFuture;

void _requestPermissionOnce() {
  _permissionFuture ??= FirebaseMessaging.instance.requestPermission();
}
```

Lyokone:
## Description

Fixes an Android crash in `firebase_messaging` when concurrent `requestPermission()` calls race through the native permission manager. The fix synchronizes the in-flight permission state and stored callback so only one OS permission request can start at a time, and the callback cannot be overwritten while a request is pending.

No automated test was added because FlutterFire CI does not currently run Android plugin JVM tests, and the existing integration test setup cannot reliably drive the Android runtime permission dialog. This was verified with the Android example build plus existing Dart analysis/tests.

## Related Issues

- Fixes https://github.com/firebase/flutterfire/issues/18246

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_messaging/firebase_messaging/android/src/main/java/io/flutter/plugins/firebase/messaging/FlutterFirebasePermissionManager.java
+++ b/packages/firebase_messaging/firebase_messaging/android/src/main/java/io/flutter/plugins/firebase/messaging/FlutterFirebasePermissionManager.java
@@ -17,7 +17,11 @@ import io.flutter.plugin.common.PluginRegistry;
 import java.util.ArrayList;
 
 class FlutterFirebasePermissionManager implements PluginRegistry.RequestPermissionsResultListener {
+  private static final String REQUEST_IN_PROGRESS_ERROR =
+      "A request for permissions is already running, please wait for it to finish before doing "
+      + "another request.";
 
+  private final Object requestLock = new Object();
   private final int permissionCode = 240;
   @Nullable private RequestPermissionsSuccessCallback successCallback;
   private boolean requestInProgress = false;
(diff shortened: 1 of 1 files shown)


cabljac (approved):
lgtm

russellwheatley (approved):

PR #18263, opened 2026-05-07
chore(fdc): Use service classes to fetch auth and appcheck instances

aashishpatil-g:
Use service resolvers to get auth and app check parameters.

### DIFF
--- a/packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
+++ b/packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
@@ -215,8 +215,8 @@ class FirebaseDataConnect extends FirebasePluginPlatform {
       required ConnectorConfig connectorConfig,
       CacheSettings? cacheSettings}) {
     app ??= Firebase.app();
-    auth ??= app.getService<FirebaseAuth>();
-    appCheck ??= app.getService<FirebaseAppCheck>();
+    auth ??= FirebaseAuth.instanceFor(app: app);
+    appCheck ??= FirebaseAppCheck.instanceFor(app: app);
 
     if (cachedInstances[app.name] != null &&
         cachedInstances[app.name]![connectorConfig.toJson()] != null) {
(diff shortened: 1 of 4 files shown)


packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
```diff
@@ -94,20 +98,23 @@ class FirebaseDataConnect extends FirebasePluginPlatform {
...
     final ws = WebSocketTransport(
       transportOptions!,
       options,
       app.options.appId,
       _sdkType,
-      appCheck,
-      auth,
+      effectiveAppCheck,
+      effectiveAuth,
     );
```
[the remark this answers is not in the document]
aashishpatil-g: fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/firebase_data_connect.dart
```diff
@@ -198,14 +205,18 @@ class FirebaseDataConnect extends FirebasePluginPlatform {
...
+          'Passing an explicit instance is deprecated, internal handling is now automatic.')
       FirebaseAppCheck? appCheck,
       CallerSDKType? sdkType,
       required ConnectorConfig connectorConfig,
       CacheSettings? cacheSettings}) {
     app ??= Firebase.app();
-    auth ??= FirebaseAuth.instanceFor(app: app);
-    appCheck ??= FirebaseAppCheck.instanceFor(app: app);
+    auth ??= app.getService<FirebaseAuth>();
+    appCheck ??= app.getService<FirebaseAppCheck>();
```
cynthiajoan: This is a bit of changing the behavior. The previous is if auth and appcheck is not passed in, instantiate one and put that. The changed code will be, if auth and appCheck is not passed in, assign it with an already registered auth/appcheck instance. If they are not instantiated, it will still keep null.

But also the previous behavior doesn't config auth and appcheck after instantiated. What is the expected thing here?

aashishpatil-g: Reverted to the previous behavior.

cynthiajoan (approved):

SelaseKay (approved):

PR #18415, opened 2026-07-02
fix(auth, iOS): add detailed error handling for Apple authorization failures

### LINKED ISSUE #18413: [firebase_auth]: iOS Sign in with Apple discards the native NSError domain/code — unrecognised ASAuthorization failures all surface as "unknown / An unknown error has occurred."
### Is there an existing issue for this?

- [x] I have searched the existing issues.

Closest prior art, none of which covers this: #10055 (user cancel used to surface as `unknown` — fixed by typing it as `canceled`), #18172 (overlapping Apple requests — fixed in 6.4.0 by serialising and typing `operation-not-allowed`), #17813 (same surface symptom, closed as an app configuration error). This issue is about the remaining fall-through branch, which still destroys the error information.

### Which plugins are affected?

Auth

### Which platforms are affected?

iOS

### Description

On iOS, `signInWithProvider(AppleAuthProvider())` runs Sign in with Apple natively via the plugin's own `ASAuthorizationController`. When the flow fails, [`authorizationController:didCompleteWithError:`](https://github.com/firebase/flutterfire/blob/firebase_auth-v6.5.4/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m#L495-L538) maps the four typed `ASAuthorizationError` codes (1001 `canceled`, 1002 `invalid-response`, 1003 `not-handled`, 1004 `failed`) to useful Flutter error codes — but `ASAuthorizationErrorUnknown` (1000) and **every other NSError domain** fall through to [`convertToFlutterError:`](https://github.com/firebase/flutterfire/blob/firebase_auth-v6.5.4/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m#L172-L200), which only extracts a code when `FIRAuthErrorUserInfoNameKey` is present. Native AuthenticationServices/AuthKit errors never carry that key, so the Dart side receives the literal fallback:

```
[firebase_auth/unknown] An unknown error has occurred.
```

`NSError.domain`, `NSError.code`, and any `NSUnderlyingError` are discarded. The only place the real error survives is the plugin's own `NSLog(@"Sign in with Apple errored: %@", error)` — visible in the Xcode console, unreachable in production.

**Why this matters in production:** fleet monitoring on a shipping app shows real users failing Sign in with Apple in under 500 ms with this opaque error. From Dart it is impossible to distinguish, for example:

- `com.apple.AuthenticationServices.AuthorizationError Code=1000` — commonly a device with no usable Apple ID session (user signed out of iCloud, restricted/managed Apple ID), where the right response is a targeted "check you're signed in to your Apple ID in Settings" message;
- `AKAuthenticationError Code=-7026` ("Told not to present authorization sheet") — a presentation conflict, where a retry works;
- anything else Apple decides to throw.

Cancellation (#10055) and request overlap (#18172) were both fixed by surfacing what actually happened; this fall-through branch is the last place where the information is destroyed instead.

**Proposal** — keep `code: "unknown"` for backward compatibility, but stop discarding the native identity. Something like:

```objc
case ASAuthorizationErrorUnknown:
default: {
  NSError *underlying = error.userInfo[NSUnderlyingErrorKey];
  NSString *detail =
      [NSString stringWithFormat:@"%@ (Domain=%@ Code=%ld%@)",
                                 error.localizedDescription.length > 0
                                     ? error.localizedDescription
                                     : @"An unknown error has occurred.",
                                 error.domain, (long)error.code,
                                 underlying == nil
                                     ? @""
                                     : [NSString stringWithFormat:
                                           @", Underlying: Domain=%@ Code=%ld",
                                           underlying.domain, (long)underlying.code]];
  completion(nil, [FlutterError errorWithCode:@"unknown" message:detail details:nil]);
  break;
}
```

(Equally happy with the domain/code going into `details` instead of the message if that's preferred.) I'm willing to submit a PR if maintainers agree with the direction.

### Reproducing the issue

1. On a physical device, sign out of the Apple Account (Settings → sign out), or use any device state where AuthenticationServices refuses the request without presenting the sheet.
2. Call `FirebaseAuth.instance.signInWithProvider(AppleAuthProvider())`.
3. The call fails immediately. The Xcode console shows the full NSError via the plugin's NSLog; Dart receives only `[firebase_auth/unknown] An unknown error has occurred.` with no way to recover the domain/code.

Any unrecognised `ASAuthorizationController` delegate error takes the same path.

### Firebase Core version

4.11.0

### Flutter Version

3.44.4 (stable)

### Relevant Log Output

```shell
# Xcode console — the information that never reaches Dart:
Sign in with Apple errored: Error Domain=com.apple.AuthenticationServices.AuthorizationError Code=1000 "(null)"

# What Dart receives:
FirebaseAuthException([firebase_auth/unknown] An unknown error has occurred.)
```

### Flutter dependencies

<details>
<summary>Expand <code>Flutter dependencies</code> snippet</summary>
<br>

```yaml
firebase_core: 4.11.0
firebase_auth: 6.5.4
firebase_ui_auth: 3.0.1
firebase_ui_oauth_apple: 2.0.1
```

</details>

### Additional context and comments

The typed ladder in `didCompleteWithError` has improved case by case (`canceled` in #10055's fix, `operation-not-allowed` via #18172) precisely because each shape became distinguishable. This change would do the same for the remaining bucket: everything currently labelled `unknown` would still be `unknown`, but carry `Domain=… Code=…` so apps can classify (the message format is already parseable the same way as the `Code: N` pattern some FIRAuth errors embed).

SelaseKay:
## Description

## Related Issues
Fixes https://github.com/firebase/flutterfire/issues/18413

## Checklist

- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m
+++ b/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m
@@ -242,7 +242,7 @@ static NSMutableDictionary<NSNumber *, FIRAuthCredential *> *credentialsMap;
     additionalData[@"underlyingNativeErrorCode"] = underlyingErrorCode;
 
     underlyingMessage =
-        [NSString stringWithFormat:@", Underlying: Domain=%@ Code=%ld", underlyingErrorDomain,
+        [NSString stringWithFormat:@", Underlying Domain=%@ Code=%ld", underlyingErrorDomain,
                                    (long)underlyingError.code];
   }
 
--- a/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m
+++ b/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.m
@@ -219,6 +219,40 @@ static NSMutableDictionary<NSNumber *, FIRAuthCredential *> *credentialsMap;
   return [FlutterError errorWithCode:code message:message details:additionalData];
 }
 
++ (FlutterError *)convertAppleAuthorizationErrorToFlutterError:(NSError *)error {
+  NSString *message = @"An unknown error has occurred.";
+  if (error.localizedDescription.length > 0) {
+    message = error.localizedDescription;
+  }
+
+  NSMutableDictionary *additionalData = [NSMutableDictionary dictionary];
+  NSString *nativeErrorDomain = error.domain ?: @"unknown";
+  NSNumber *nativeErrorCode = @((long)error.code);
+
+  additionalData[@"nativeErrorDomain"] = nativeErrorDomain;
+  additionalData[@"nativeErrorCode"] = nativeErrorCode;
+
+  NSError *underlyingError = error.userInfo[NSUnderlyingErrorKey];
+  NSString *underlyingMessage = @"";
+  if (underlyingError != nil) {
+    NSString *underlyingErrorDomain = underlyingError.domain ?: @"unknown";
+    NSNumber *underlyingErrorCode = @((long)underlyingError.code);
+
+    additionalData[@"underlyingNativeErrorDomain"] = underlyingErrorDomain;
+    additionalData[@"underlyingNativeErrorCode"] = underlyingErrorCode;
+
+    underlyingMessage =
+        [NSString stringWithFormat:@", Underlying: Domain=%@ Code=%ld", underlyingErrorDomain,
+                                   (long)underlyingError.code];
+  }
+
+  NSString *detailMessage =
+      [NSString stringWithFormat:@"%@ (Domain=%@ Code=%ld%@)", message, nativeErrorDomain,
+                                 (long)error.code, underlyingMessage];
+
+  return [FlutterError errorWithCode:@"unknown" message:detailMessage details:additionalData];
+}
+
 + (id)getNSDictionaryFromAuthCredential:(FIRAuthCredential *)authCredential {
   if (authCredential == nil) {
     return [NSNull null];
@@ -532,7 +566,7 @@ static void handleSignInWithApple(FLTFirebaseAuthPlugin *object, FIRAuthDataResu
 
     case ASAuthorizationErrorUnknown:
     default:
-      completion(nil, [FLTFirebaseAuthPlugin convertToFlutterError:error]);
+      completion(nil, [FLTFirebaseAuthPlugin convertAppleAuthorizationErrorToFlutterError:error]);
       break;
   }
 }


bjrochem72:
Thanks for picking this up so quickly — the native side is exactly what I was hoping for.

Two things I noticed when tracing what happens to the new error on the Dart side, one of which I think is worth fixing before merge:

**1. The `Underlying: ` label gets the message truncated in Dart.**

In `firebase_auth_platform_interface/lib/src/method_channel/utils/exception.dart`, the Pigeon path does:

```dart
var parsedMessage = platformException.message?.split(': ').last;
```

The new format includes the literal `", Underlying: Domain=%@ Code=%ld"`, so whenever an underlying error is present, everything before that `": "` is discarded. For example:

Built natively:
`The operation couldn't be completed. (com.apple.AuthenticationServices.AuthorizationError error 1000.) (Domain=com.apple.AuthenticationServices.AuthorizationError Code=1000, Underlying: Domain=AKAuthenticationError Code=-7026)`

Received in Dart:
`Domain=AKAuthenticationError Code=-7026)`

— the localizedDescription and the primary Domain/Code are lost, which is the exact case (AKAuthenticationError −7026) that motivated the issue. The no-underlying case is fine, since Apple's synthesized descriptions don't contain `": "`.

Minimal fix: drop the colon from the label, e.g.

```objc
[NSString stringWithFormat:@", Underlying Domain=%@ Code=%ld", ...]
```

**2. The structured `details` dictionary doesn't currently reach app code.**

`platformExceptionToFirebaseAuthException` only reads `authCredential` and `email` from details, and `FirebaseAuthException` has no field to carry the rest, so `nativeErrorDomain` / `nativeErrorCode` / `underlyingNativeErrorDomain` / `underlyingNativeErrorCode` are dropped during conversion. Totally fine by me if the message stays the single source (that's parseable once point 1 is fixed) — just flagging it so the keys aren't assumed to be reachable from Dart.

Happy to test the branch against the device states from the original report once it's updated.

SelaseKay:
Good catch @bjrochem72. I've updated the PR. Will wait for your feedback on the test.

bjrochem72:
Tested the updated branch on device (iOS 26.4 simulator, no Apple ID signed in — the `AuthorizationError` 1000 case from the report). Confirming the truncation fix in `5a0c012d` works: Dart now receives

```
[firebase_auth/unknown] The operation couldn't be completed. (com.apple.AuthenticationServices.AuthorizationError error 1000.) (Domain=com.apple.AuthenticationServices.AuthorizationError Code=1000)
```

— the primary `Domain`/`Code` survive the platform-interface `message.split(': ').last` path intact, where before this PR it was just `An unknown error has occurred.` The colon-free `, Underlying Domain=… Code=…` label is the right fix for the underlying-error case; I wasn't able to force `AKAuthenticationError -7026` deterministically on device, but by inspection nothing before that label can be truncated now, so 👍.

Note on my harness: I couldn't pull the `auth_18413` branch directly because its Dart layer references `FirebasePlugin`/`pluginConstants` from unreleased `firebase_core` master. I tested by applying just this PR's `FLTFirebaseAuthPlugin.m` on top of the released `firebase_auth` 6.5.4 — so this exercises exactly the native change.

On "the test": I only see `FLTFirebaseAuthPlugin.m` in the diff — if you pushed a test it may not have synced. Since the `details` dictionary isn't surfaced to Dart, an ObjC-level unit test over `convertAppleAuthorizationErrorToFlutterError:` (assert the formatted `message` and the `nativeErrorDomain`/`nativeErrorCode`/underlying keys, with and without an `NSUnderlyingError`) would cover it best. Happy to review once it's up.

Unrelated: the `web (tests)` red looks like infra/flake — this is an iOS-only ObjC change.

demolaf (approved):

Lyokone (approved):

PR #18430, opened 2026-07-10
fix(sql_connect): Refactor to improve caching performance for large result sets

aashishpatil-g:
## Description

Large query caching can block main thread as it tries to normalize the data and write to sqlite. Moving this logic off the main thread to a separate isolate will prevent locking the main thread. Other optimizations 
- Use prepared sql statements 
- Use of a transaction to wrap the full normalization process to consolidate multiple writes.

These changes only impact native platforms. Web caching is still done on main isolate as isolates aren't supported on web. Web performance will be a future topic. 

Also, added tests for
- Unit test for Parsing and storing large query result set
- Unit test to ensure main thread isn't blocked during caching
- Added integration test for caching.

## Related Issues

Fixes - https://github.com/firebase/firebase-tools/issues/10776

## Breaking Change
This is NOT a breaking change.

### DIFF
--- a/packages/firebase_data_connect/firebase_data_connect/example/lib/generated/.guides/usage.md
+++ b/packages/firebase_data_connect/firebase_data_connect/example/lib/generated/.guides/usage.md
@@ -8,9 +8,9 @@ MoviesConnector.instance.addDateAndTimestamp(addDateAndTimestampVariables).execu
 MoviesConnector.instance.seedMovies().execute();
 MoviesConnector.instance.createMovie(createMovieVariables).execute();
 MoviesConnector.instance.deleteMovie(deleteMovieVariables).execute();
+MoviesConnector.instance.deleteAllMovieData().execute();
 MoviesConnector.instance.thing(thingVariables).execute();
 MoviesConnector.instance.seedData().execute();
-MoviesConnector.instance.ListMovies().execute();
 
 ```
 
--- a/packages/firebase_data_connect/firebase_data_connect/example/lib/generated/delete_all_movie_data.dart
+++ b/packages/firebase_data_connect/firebase_data_connect/example/lib/generated/delete_all_movie_data.dart
@@ -20,19 +20,14 @@ class DeleteAllMovieDataVariablesBuilder {
 
 @immutable
 class DeleteAllMovieDataData {
-  final int directedByDeleteMany;
-  final int movieDeleteMany;
-  final int personDeleteMany;
+  final int directedBy_deleteMany;
+  final int movie_deleteMany;
+  final int person_deleteMany;
   DeleteAllMovieDataData.fromJson(dynamic json)
-      : directedByDeleteMany = nativeFromJson<int>(
-          json['directedBy_deleteMany'],
-        ),
-        movieDeleteMany = nativeFromJson<int>(
-          json['movie_deleteMany'],
-        ),
-        personDeleteMany = nativeFromJson<int>(
-          json['person_deleteMany'],
-        );
+      : directedBy_deleteMany =
+            nativeFromJson<int>(json['directedBy_deleteMany']),
+        movie_deleteMany = nativeFromJson<int>(json['movie_deleteMany']),
+        person_deleteMany = nativeFromJson<int>(json['person_deleteMany']);
   @override
   bool operator ==(Object other) {
     if (identical(this, other)) {
@@ -43,29 +38,29 @@ class DeleteAllMovieDataData {
     }
 
     final DeleteAllMovieDataData otherTyped = other as DeleteAllMovieDataData;
-    return directedByDeleteMany == otherTyped.directedByDeleteMany &&
-        movieDeleteMany == otherTyped.movieDeleteMany &&
-        personDeleteMany == otherTyped.personDeleteMany;
+    return directedBy_deleteMany == otherTyped.directedBy_deleteMany &&
+        movie_deleteMany == otherTyped.movie_deleteMany &&
+        person_deleteMany == otherTyped.person_deleteMany;
   }
 
   @override
   int get hashCode => Object.hashAll([
-        directedByDeleteMany.hashCode,
-        movieDeleteMany.hashCode,
-        personDeleteMany.hashCode,
+        directedBy_deleteMany.hashCode,
+        movie_deleteMany.hashCode,
+        person_deleteMany.hashCode
       ]);
 
   Map<String, dynamic> toJson() {
     Map<String, dynamic> json = {};
-    json['directedBy_deleteMany'] = nativeToJson<int>(directedByDeleteMany);
-    json['movie_deleteMany'] = nativeToJson<int>(movieDeleteMany);
-    json['person_deleteMany'] = nativeToJson<int>(personDeleteMany);
+    json['directedBy_deleteMany'] = nativeToJson<int>(directedBy_deleteMany);
+    json['movie_deleteMany'] = nativeToJson<int>(movie_deleteMany);
+    json['person_deleteMany'] = nativeToJson<int>(person_deleteMany);
     return json;
   }
 
-  const DeleteAllMovieDataData({
-    required this.directedByDeleteMany,
-    required this.movieDeleteMany,
-    required this.personDeleteMany,
+  DeleteAllMovieDataData({
+    required this.directedBy_deleteMany,
+    required this.movie_deleteMany,
+    required this.person_deleteMany,
   });
 }
--- a/packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
+++ b/packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
@@ -43,8 +43,8 @@ class Cache {
   // Non-web isolate handling
   Isolate? _isolate;
   SendPort? _toIsolatePort;
-  final Completer<SendPort> _toIsolatePortCompleter = Completer<SendPort>();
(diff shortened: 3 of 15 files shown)


packages/firebase_data_connect/firebase_data_connect/lib/src/cache/sqlite_cache_provider.dart
```diff
@@ -23,21 +24,53 @@ class SQLite3CacheProvider implements CacheProvider {
   late final Database _db;
   final String _identifier;
   final bool memory;
+  final String? customDbPath;
+  bool _inTransaction = false;
```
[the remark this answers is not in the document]
aashishpatil-g: Fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -59,16 +83,91 @@ class Cache {
...
+  }
+
+  void _updateIsolateProvider() async {
+    final identifier = _constructCacheIdentifier();
+    if (_currentIdentifier == identifier) {
+      return;
+    }
+    _currentIdentifier = identifier;
+
+    final completer = Completer<void>();
```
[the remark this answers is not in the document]
aashishpatil-g: Fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -59,16 +83,91 @@ class Cache {
...
+  void _startIsolate() async {
+    try {
+      final receivePort = ReceivePort();
+      _isolate = await Isolate.spawn(_cacheIsolateEntry, receivePort.sendPort);
+
+      final mainReceivePortStream = receivePort.asBroadcastStream();
+      _toIsolatePort = await mainReceivePortStream.first as SendPort;
+      _toIsolatePortCompleter.complete(_toIsolatePort!);
+
+      mainReceivePortStream.listen(_handleIsolateMessage);
```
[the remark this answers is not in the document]
aashishpatil-g: Fixed

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -79,93 +178,355 @@ class Cache {
...
+
+      case 'dispose':
+        if (provider != null) {
+          await provider.dispose();
+        }
+        isolateReceivePort.close();
+        break;
+    }
+  });
+}
```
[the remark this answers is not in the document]
aashishpatil-g: Fixed

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/.guides/usage.md
```diff
@@ -8,9 +8,9 @@ MoviesConnector.instance.addDateAndTimestamp(addDateAndTimestampVariables).execu
 MoviesConnector.instance.seedMovies().execute();
 MoviesConnector.instance.createMovie(createMovieVariables).execute();
 MoviesConnector.instance.deleteMovie(deleteMovieVariables).execute();
+MoviesConnector.instance.deleteAllMovieData().execute();
 MoviesConnector.instance.thing(thingVariables).execute();
 MoviesConnector.instance.seedData().execute();
-MoviesConnector.instance.ListMovies().execute();
```
cynthiajoan, line 13: is this deletion intentional?

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/delete_all_movie_data.dart
```diff
@@ -20,19 +20,14 @@ class DeleteAllMovieDataVariablesBuilder {
 
 @immutable
 class DeleteAllMovieDataData {
-  final int directedByDeleteMany;
-  final int movieDeleteMany;
-  final int personDeleteMany;
+  final int directedBy_deleteMany;
+  final int movie_deleteMany;
+  final int person_deleteMany;
```
cynthiajoan, line 25: what change results in the renaming of the generated file?

cynthiajoan (approved):

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/.guides/usage.md
aashishpatil-g: It was created for the test to clear data

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/delete_all_movie_data.dart
aashishpatil-g: There were some changes done recently to support batch mutations. I suspect this might have caused the code generator to change.

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/.guides/usage.md
aashishpatil-g: My bad. I thought you were referring to the deleteAll. 

Hmm - this seems to be somehting in the code generator logic that generates the dot guides. (markdown files) The example project hasn't been regenerated in a while, so something must have changed. 

This isn't code impacting though.

packages/firebase_data_connect/firebase_data_connect/example/lib/generated/delete_all_movie_data.dart
aashishpatil-g: Took a look at the gql operation definitions. 

The example project got regenerated and this correctly matches the gql definitions. Its possible the original names had been manually tweaked in the generated code.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -79,93 +187,357 @@ class Cache {
     }
 
     dataConnect.auth!.authStateChanges().listen((User? user) {
-      _initializeProvider();
+      if (kIsWeb || _isolateFallbackMode) {
+        _initializeLocalProvider();
+      } else {
+        _updateIsolateProvider().catchError((e) {
```
Lyokone, line 204: We should probably await that and prevent cache access during this call? 
Otherwise we might return the cache of another user when authChange?

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -79,93 +187,357 @@ class Cache {
...
+        final queryId = message['queryId'] as String;
+        final data = message['data'] as Map<String, dynamic>;
+        final ttlMicroseconds = message['ttl'] as int?;
+        final extensions = message['extensions'] as Map<String, dynamic>?;
+
+        final maxAge = ttlMicroseconds != null
+            ? Duration(microseconds: ttlMicroseconds)
+            : Duration.zero;
+
+        final impactedQueryIds = await dehydrateAndUpdateCache(
```
Lyokone: This should run within try/catch I believe? Otherwise if it crashes we will never return to the user and the app will hang.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -79,93 +187,357 @@ class Cache {
...
+            'requestId': requestId,
+            'data': null,
+          });
+          break;
+        }
+
+        final queryId = message['queryId'] as String;
+        final allowStale = message['allowStale'] as bool;
+
+        final hydratedJson = await fetchAndHydrateCache(
```
Lyokone: Same here

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
```diff
@@ -79,93 +187,357 @@ class Cache {
...
+              'CacheProvider failed to initialize in background isolate.'));
+        }
+      }
+      if (!_startupCompleted) {
+        _startupCompleted = true;
+        if (success) {
+          _startupCompleter.complete();
+        } else {
+          _startupCompleter
+              .completeError(StateError('CacheProvider failed to initialize.'));
```
Lyokone: I don't think we have ways of reseting this state if the following cache hits? If subsequent cache access succeed, shouldn't we allow to reset `startupCompleted`?

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
aashishpatil-g: Done. wrapped it in try catch with sending of error.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
aashishpatil-g: done.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
aashishpatil-g: Addressed - there is retry from access methods if last init failed.

packages/firebase_data_connect/firebase_data_connect/lib/src/cache/cache.dart
aashishpatil-g: Addressed - thanks for catching this!

Lyokone (approved):

PR #18460, opened 2026-07-16
feat(auth,ios): add support for `migrateCurrentUser`

### LINKED ISSUE #11870: 🚀 [auth] userAccessGroup setting signs users out, and there's no way to migrate existing users
## What feature would you like to see?

I set the userAccessGroup in my app, however, it causes existing users to log out.

In the firebase swift docs, there's instructions on how to update the user using swift SDK `getStoredUser` and `updateCurrentUser` https://firebase.google.com/docs/auth/ios/single-sign-on

However, this capability is not included in the Flutter SDK, so I cannot perform the migration. If I deploy my app, all my users are going to get logged out, which is an especially big problem for anonymous users.

Lyokone:
## Description

Changing Firebase Auth's Apple keychain access group clears an existing unshared user before Dart can retain it. This adds an opt-in `migrateCurrentUser` setting that captures the native user before switching access groups and restores it with `updateCurrentUser`, preserving existing sessions including anonymous users.

The captured user is `Auth.currentUser`, falling back to reading the keychain directly when Auth has not restored the session yet, so migration also works when it runs at app startup. `setSettings` returns the migrated user so the Dart `currentUser` cache is reconciled before auth-state events arrive.

`migrateCurrentUser` requires a non-null `userAccessGroup` and is iOS/macOS only; other platforms ignore it.

An iOS integration test verifies that an anonymous user's UID is preserved when moving to the auth example app's shared keychain access group.

## Related Issues

- Fixes https://github.com/firebase/flutterfire/issues/11870

## Checklist

- [x] I signed the [CLA].

## Breaking Change


SelaseKay (approved):

packages/firebase_auth/firebase_auth_platform_interface/lib/src/method_channel/method_channel_firebase_auth.dart
```diff
@@ -607,16 +614,33 @@ class MethodChannelFirebaseAuth extends FirebaseAuthPlatform {
...
+      if (migratedUser != null) {
+        MethodChannelMultiFactor? multiFactorInstance =
+            _multiFactorInstances[app.name];
+        if (multiFactorInstance == null) {
+          multiFactorInstance = MethodChannelMultiFactor(this);
+          _multiFactorInstances[app.name] = multiFactorInstance;
+        }
+        final user = MethodChannelUser(this, multiFactorInstance, migratedUser);
+        currentUser = user;
+        sendAuthChangesEvent(app.name, user);
```
russellwheatley: This might double-fire userChanges(). updateCurrentUser on the native iOS side calls possiblyPostAuthStateChangeNotification() internally, which triggers the already-registered ID-token listener and pushes into the same _userChangesListeners[appName] controller that sendAuthChangesEvent writes to here. So a single setSettings(migrateCurrentUser: true) call emits two events: this one, then the real one from the native listener a moment later.

signInAnonymously, signInWithCredential, etc. only do currentUser = ... and skip sendAuthChangesEvent for this reason I think. Maybe drop the manual call here and let the native listener handle it, same as sign-in.

russellwheatley (changes requested):

packages/firebase_auth/firebase_auth_platform_interface/lib/src/method_channel/method_channel_firebase_auth.dart
Lyokone: Good catch, and it is a bit worse than two events.

`updateCurrentUser(_:byForce:savingToDisk:)` calls `possiblyPostAuthStateChangeNotification()` after assigning `_currentUser`, and the migration crosses that path twice: `internalUseUserAccessGroup` first does `updateCurrentUser(nil, byForce: false, savingToDisk: false)` (posts, `lastNotifiedUserToken` -> nil), then our `updateCurrentUser(user)` posts again. Both reach `_handleIdTokenChangesListener`, which writes to `_userChangesListeners` — so with the manual call `userChanges()` saw three emissions for one `setSettings`.

Dropped the `sendAuthChangesEvent` call in 1f8dfb8c9 and kept only `currentUser = ...`, same as the sign-in APIs. The synchronous cache reconciliation was the only thing that needed the returned user.

Note the transient null stays: listeners see null then the user, because the native switch signs out in memory before the restore. That is native behaviour and not something the plugin can collapse from Dart.

russellwheatley (approved):

PR #18494, opened 2026-07-29
fix(sql_connect): non-subscribed queries should be tracked

aashishpatil-g:
QueryRefs should be reused even if queries are non-subscribed. Previously we would only track if a user called subscribe, but that caused multiple queryrefs to be created.

Added new tests to validate fix.


aashishpatil-g:
Doesn't seem to be the right fix. Will reopen more focussed PR.

PR #18484, opened 2026-07-28
feat(sql_connect): add "X-Client-Platform" and "X-Client-Version" headers

dconeybe:
This PR adds `x-client-platform` and `x-client-version` request headers to `firebase_data_connect`. These headers supply the client platform (`flutter`) and SDK version (`packageVersion`) to enable metrics collection in Cloud Monitoring.

**NOTE**: Shortly after merging this PR, it was decided to delete the `x-client-platform` header and merge its value into `x-client-version`. See https://github.com/firebase/flutterfire/pull/18502 for details.

### Highlights

* **New Metadata Headers**: Added `x-client-platform` (set to `"flutter"`) and `x-client-version` (set to `packageVersion`) request headers across REST (`rest_transport.dart`) and WebSocket (`websocket_transport.dart`) transports.
* **Unit Tests**: Added unit tests in `rest_transport_test.dart` to verify `x-client-platform` and `x-client-version` headers are attached on outgoing requests.

<details>

<summary><b>Changelog</b></summary>

* **rest_transport.dart**
  * Added `x-client-platform` and `x-client-version` headers to `invokeOperation`.
* **websocket_transport.dart**
  * Added `x-client-platform` and `x-client-version` headers to `_buildHeaders` and exposed `buildHeaders` for testing.
* **rest_transport_test.dart**
  * Added unit tests asserting `x-client-platform` and `x-client-version` headers are included in REST requests.
</details>

PR #18502, opened 2026-07-30
feat(sql_connect): merge "X-Client-Platform" header into "X-Client-Version"

dconeybe:
This PR updates the gRPC metadata in `firebase_data_connect ` by merging the `X-Client-Platform` header into the `X-Client-Version` header (formatting as `flutter/<version>`) and removing the standalone `X-Client-Platform` header.

This is an update to PR https://github.com/firebase/flutterfire/pull/18484 which was merged yesterday and added both of these headers. This update was made to accommodate the firebase-js-sdk because the "X-Client-Platform" header causes the CORS OPTIONS preflight request to fail with a 403 "Permission Denied" error due to that header being absent from Google's list of whitelisted CORS headers. See https://github.com/firebase/firebase-js-sdk/pull/10217 for a few more details, if interested. Note that this exact problem affects this flutter sdk when running in a web browser, so the fix is critical here in this repository too.

### Highlights

* **Header Consolidation**: Merged the platform identifier into the `X-Client-Version` header (e.g. `flutter/<version>`) and removed the separate `X-Client-Platform` (`x-client-platform`) header.
* **Test Updates**: Updated test suites to verify the new `X-Client-Version` header value format.

### DIFF
--- a/packages/firebase_data_connect/firebase_data_connect/lib/src/network/rest_transport.dart
+++ b/packages/firebase_data_connect/firebase_data_connect/lib/src/network/rest_transport.dart
@@ -86,8 +86,7 @@ class RestTransport implements DataConnectTransport {
       'Accept': 'application/json',
       'x-goog-api-client': getGoogApiVal(sdkType, packageVersion),
       'x-firebase-client': getFirebaseClientVal(packageVersion),
-      'x-client-platform': 'flutter',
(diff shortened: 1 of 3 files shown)


aashishpatil-g (approved):

stephenarosaj (approved):
LGTM

Issue #18561, opened 2026-08-11
[firebase_auth] getIdTokenResult() throws an uncatchable fatal exception on Android when the native token refresh fails on a background thread


luanbatistadenga:
### Steps to Reproduce

We don't have a fully deterministic local repro — this surfaced in production via Crashlytics, and matches the same general class of bug as the closed #11311 ("Crash when calling user.getIdToken(true)"), but with a different (also uncaught) underlying `FirebaseAuthException`.

1. Have a signed-in `User` on Android.
2. Call `user.getIdTokenResult(true)` (force refresh) from a Dart `try`/`catch` block.
3. Have the native token refresh fail — either due to network connectivity to `securetoken.googleapis.com`, or because the refresh token itself is no longer valid (`user-token-expired` / a generic `[firebase_auth/unknown]` "The user's credential is no longer valid. The user must sign in again." message).

### Expected results

The `FirebaseAuthException` should be catchable by the surrounding Dart `try`/`catch`, the same way it is for a normal awaited `Future` rejection.

### Actual results

The exception surfaces as an **uncaught, fatal** crash, even though the call site has an enclosing `try`/`catch` (and, in our case, correct typed classification of the exception's `.code`/`.message` in the `catch` block — none of which ever runs).

Crashlytics-reported stack trace (thread name is notable — the exception originates on a Firebase-managed background thread, not the calling Dart isolate/await chain):

```
Thread: Firebase Background Thread #3 (crashed)
io.flutter.plugins.firebase.crashlytics.FlutterError: [firebase_auth/unknown] The user's credential is no longer valid. The user must sign in again.. Error thrown .
  at ._extractReplyValueOrThrow (package:firebase_auth_platform_interface/src/pigeon/messages.pigeon.dart:26)
  at FirebaseAuthUserHostApi.getIdToken (package:firebase_auth_platform_interface/src/pigeon/messages.pigeon.dart:1988)
  at MethodChannelUser.getIdTokenResult (package:firebase_auth_platform_interface/src/method_channel/method_channel_user.dart:58)
  at FirebaseUserExtension.id (package:denga_mobile/src/core/data/extensions/firebase_user_extension.dart:5)
  at _AppState.authHandle (package:denga_mobile/src/app/app.dart:135)
```

A second variant of the exact same issue (same blame frame, `FirebaseUserExtension.id`) shows the network-failure flavor instead:

```
[...]

PR #11362, opened 2023-07-24
fix(auth, android): `getIdToken()` `IllegalStateException` crash fix

russellwheatley:
## Description

see title.

## Related Issues

fixes https://github.com/firebase/flutterfire/issues/11297

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process. Updating the `pubspec.yaml` and changelogs is not required.

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] The analyzer (`melos run analyze`) does not report any problems on my PR.
- [ ] I read and followed the [Flutter Style Guide].
- [ ] I signed the [CLA].
- [ ] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change.
- [X] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/firebase/flutterfire/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #18605, opened 2026-08-22
fix(auth, android): surface the error code when getIdToken() fails

dyt9120:
## Description

On Android, a token refresh the backend refuses — the account was disabled or deleted, or its refresh token revoked — reaches Dart from `User.getIdToken()` / `getIdTokenResult()` as `[firebase_auth/unknown]` (message `The user's credential is no longer valid. The user must sign in again.` or `The user account has been disabled by an administrator.`) instead of `user-token-expired` / `user-not-found` / `user-disabled`, the codes iOS and web produce for the same situation. An app that branches on the code to sign a dead session out never sees it on Android.

**Cause.** Since #11362, `getIdToken` awaits the native Task with `Tasks.await()`, which reports a failed Task as a `java.util.concurrent.ExecutionException` wrapping the Task's own exception. `FlutterFirebaseAuthPluginException.parserExceptionToFlutter` only reads the error code off the exception it is handed (`nativeException is FirebaseAuthException`), so the wrapped `FirebaseAuthInvalidUserException` is never looked at and `code` stays `UNKNOWN`. The three network-ish arms already check `cause`, which is why `network-request-failed` survived that change and the auth codes did not.

**Fix.** Unwrap an `ExecutionException` to its cause before classifying. `getIdToken` is the only `Tasks.await` call site in the plugin, so this is the only path affected. The message is unchanged — it already came from the cause through `ExecutionException.getMessage()` and the Dart side's `split(': ').last` — only the code changes.

**Verified** on an Android 13 (API 33) emulator against the Auth emulator: disabling the signed-in account behind its back and forcing a refresh yields `user-disabled` with this change and `unknown` without it. The added e2e test (`firebase_auth_user_e2e_test.dart`, `getIdToken()` group) does the same through the existing `emulatorDisableUser` helper; it passes on iOS and web today and fails on Android without the fix.

## Related Issues

- The `[firebase_auth/unknown] "The user's credential is no longer valid…"` noted in step 3 of #18561's repro is this bug.
- Introduced by #11362.

## Checklist

- [ ] All existing and new tests are passing. — The new test's flow was verified by hand on the Android emulator (above); I have not run the full e2e suite locally and am relying on CI for it.
- [x] I updated/added relevant documentation (doc comments with `///`). — none needed; the change is internal to the Android plugin.
- [ ] The analyzer (`melos run analyze`) does not report any problems on my PR. — not run locally (sparse checkout); the only Dart change is one e2e test written in the surrounding file's style.
- [ ] I signed the [CLA]. — will complete when the CLA check asks.

## Breaking Change

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -67,3 +67,4 @@ Liu Zhisong <liuzs0666@gmail.com>
 Ievgenii Kovtun <jeka.ns@gmail.com>
 Dinu-Stefan Rusu <rusudinustefan@gmail.com>
 Marwan Salim Ba Matraf <marwans12@hotmail.com>
+Do Yeong Tak <dyt9120@gmail.com>
--- a/packages/firebase_auth/firebase_auth/android/src/main/kotlin/io/flutter/plugins/firebase/auth/FlutterFirebaseAuthPluginException.kt
+++ b/packages/firebase_auth/firebase_auth/android/src/main/kotlin/io/flutter/plugins/firebase/auth/FlutterFirebaseAuthPluginException.kt
@@ -12,10 +12,75 @@ import com.google.firebase.auth.FirebaseAuthException
 import com.google.firebase.auth.FirebaseAuthMultiFactorException
 import com.google.firebase.auth.FirebaseAuthUserCollisionException
 import com.google.firebase.auth.FirebaseAuthWeakPasswordException
+import java.io.IOException
 import java.util.UUID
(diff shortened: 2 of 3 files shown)


SelaseKay:
Hi @dyt9120, thanks for the contribution. This LGTM! Kindly sign CLA and resolve the formatting issues in ci so I can approve. Thanks

PR #18628, opened 2026-08-29
fix(firestore,windows): report unimplemented for sum/average aggregations

m1roxx:
## Description

On Windows, `sum()` and `average()` aggregations are silently unsupported.

The Firebase C++ SDK exposes only `Query::Count()` — there is no sum or average aggregation in `firebase/firestore/aggregate_query.h`. The Windows plugin handled that by printing to stdout and moving on:

```cpp
case AggregateType::kSum:
  std::cout << "Sum is not supported on C++" << std::endl;
  break;
```

Two user-visible consequences:

- `collection.aggregate(count(), sum('foo'))` **resolves successfully** with the count filled in and `getSum('foo')` returning `null`. `getSum` is typed `double?`, so this is indistinguishable from a query that legitimately has no sum — the app silently reads wrong data.
- `collection.aggregate(sum('foo'))` leaves `aggregate_query` default constructed, so `AggregateQuery::Get()` returns `FailedFuture` ([aggregate_query.cc](https://github.com/firebase/firebase-cpp-sdk/blob/main/firestore/src/common/aggregate_query.cc#L97-L101)) and the caller gets an opaque error with no indication that the operation is unsupported on this platform.

This PR rejects any aggregation request containing `sum()` or `average()` with an explicit `unimplemented` error, following the precedent set for Windows Storage listing in #18449 (*"throw unimplemented instead of returning misleading empty results"*). `count()` is unaffected.

While adding the integration test I found that `MethodChannelAggregateQuery.get` is the only Firestore method-channel call not wrapped in `convertPlatformException`, so the error reached callers as a raw `PlatformException` instead of a `FirebaseException`. That is fixed here too, the same way `MethodChannelQuery.get` does it — it also affects Android/iOS, where any aggregate-query failure (for example `permission-denied`) currently surfaces in the wrong exception type.

## Related Issues

No existing issue — found while auditing the Windows plugin. Happy to file one if you'd prefer the report tracked separately.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

An app that today reads `null` from `getSum()` on Windows will now see a `FirebaseException`. That is the point of the change: the previous value was wrong, not absent.

### DIFF
--- a/packages/cloud_firestore/cloud_firestore/example/integration_test/query_e2e.dart
+++ b/packages/cloud_firestore/cloud_firestore/example/integration_test/query_e2e.dart
@@ -4139,6 +4139,40 @@ void runQueryTests() {
         },
         skip: defaultTargetPlatform == TargetPlatform.windows,
       );
+
+      test(
+        'sum() & average() report that they are unsupported on Windows',
+        () async {
+          final collection = await initializeTest('unsupported-aggregations');
+
+          await collection.add({'foo': 1});
+
+          final unsupportedError = isA<FirebaseException>()
+              .having((error) => error.code, 'code', 'unimplemented')
+              .having(
+                (error) => error.message,
+                'message',
+                'Sum and average aggregations are not supported by the '
+                    'Firebase C++ SDK on Windows.',
+              );
+
+          await expectLater(
+            collection.aggregate(sum('foo')).get(),
+            throwsA(unsupportedError),
+          );
+          await expectLater(
+            collection.aggregate(average('foo')).get(),
+            throwsA(unsupportedError),
+          );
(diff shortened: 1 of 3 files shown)


m1roxx:
Friendly ping — this has been open since Aug 29 with no triage. CI is green. Small change: the Windows Firestore implementation silently returns wrong results for `sum()` / `average()` aggregations, this makes it report `unimplemented` instead.

SelaseKay (changes requested):
Hi @m1roxx, thanks for the contribution. Kindly merge latest `main` into your branch to resolve CI. Besides that, LGTM.

SelaseKay (approved):

Lyokone (approved):

PR #18641, opened 2026-09-03
fix(auth, iOS): guard URL handlers when no FirebaseApp is configured

jobweegink:
## Description

`FLTFirebaseAuthPlugin` (iOS) handles incoming URLs in `application(_:open:options:)` and `scene(_:openURLContexts:)` by calling `Auth.auth()` unconditionally. When the default `FirebaseApp` has not been configured yet, `Auth.auth()` hits FirebaseAuth's fatal error

> The default FirebaseApp instance must be configured before the default Auth instance can be initialized

and the process dies on any URL delivered through one of the app's `CFBundleURLSchemes`.

This happens in apps that initialise Firebase from Dart with explicit `FirebaseOptions` (no `GoogleService-Info.plist`) some time after start-up — e.g. after reading which environment to use from local storage — and in debug/test binaries that never call `Firebase.initializeApp` at all. In both cases the plugin is still registered as an app/scene lifecycle delegate and receives the URL.

`ensureAPNSTokenSetting()` in the same file already guards with `FirebaseApp.app() != nil`. This PR applies the same guard to the two URL handlers: when no app is configured they return `false` (not handled) instead of trapping.

**Reproduction** (firebase_auth 6.6.1, iOS 26 simulator): a `CFBundleURLTypes` entry with a custom scheme, no `GoogleService-Info.plist`, `Firebase.initializeApp` not yet called, then `xcrun simctl openurl booted "<scheme>:///x"` → crash in `scene(_:openURLContexts:)` (`Auth.swift:151`, `FLTFirebaseAuthPlugin.swift:194`).

## Related Issues

None found for this exact trap; happy to link one if maintainers know of it.

## Checklist

- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors — the change is a two-line guard in native iOS lifecycle-delegate code; there is no existing harness for delivering a URL before `FirebaseApp.configure`. I can add one if you point me at the right place.
- [ ] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.swift
+++ b/packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.swift
@@ -186,11 +186,15 @@ public class FLTFirebaseAuthPlugin: NSObject, FlutterPlugin, FLTFirebasePluginPr
       open url: URL,
       options: [UIApplication.OpenURLOptionsKey: Any] = [:]
     ) -> Bool {
-      Auth.auth().canHandle(url)
+      // Auth.auth() traps when no FirebaseApp has been configured yet (e.g. an
+      // app that initialises Firebase from Dart-side options after start-up).
+      guard FirebaseApp.app() != nil else { return false }
+      return Auth.auth().canHandle(url)
     }
 
     public func scene(_ scene: UIScene, openURLContexts urlContexts: Set<UIOpenURLContext>) -> Bool
     {
+      guard FirebaseApp.app() != nil else { return false }
       for urlContext in urlContexts where Auth.auth().canHandle(urlContext.url) {
         return true
       }


packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.swift
```diff
@@ -186,11 +186,15 @@ public class FLTFirebaseAuthPlugin: NSObject, FlutterPlugin, FLTFirebasePluginPr
       open url: URL,
       options: [UIApplication.OpenURLOptionsKey: Any] = [:]
     ) -> Bool {
-      Auth.auth().canHandle(url)
+      // Auth.auth() traps when no FirebaseApp has been configured yet (e.g. an
+      // app that initialises Firebase from Dart-side options after start-up).
+      guard FirebaseApp.app() != nil else { return false }
+      return Auth.auth().canHandle(url)
```
russellwheatley, line 192: Might be worth a debug-only log here so it's visible during development that a URL got dropped because Firebase wasn't configured yet.

```suggestion
      // Auth.auth() traps when no FirebaseApp has been configured yet (e.g. an
      // app that initialises Firebase from Dart-side options after start-up).
      guard FirebaseApp.app() != nil else {
        #if DEBUG
          print(
            "[firebase_auth] Ignoring URL because no FirebaseApp is configured yet. " +
              "Call Firebase.initializeApp() before this URL is delivered if Auth should handle it."
          )
        #endif
        return false
      }
      return Auth.auth().canHandle(url)
```

packages/firebase_auth/firebase_auth/ios/firebase_auth/Sources/firebase_auth/FLTFirebaseAuthPlugin.swift
```diff
@@ -186,11 +186,15 @@ public class FLTFirebaseAuthPlugin: NSObject, FlutterPlugin, FLTFirebasePluginPr
...
     }
 
     public func scene(_ scene: UIScene, openURLContexts urlContexts: Set<UIOpenURLContext>) -> Bool
     {
+      guard FirebaseApp.app() != nil else { return false }
```
russellwheatley, line 197: Same here for consistency with the other handler.

```suggestion
      guard FirebaseApp.app() != nil else {
        #if DEBUG
          print(
            "[firebase_auth] Ignoring URL because no FirebaseApp is configured yet. " +
              "Call Firebase.initializeApp() before this URL is delivered if Auth should handle it."
          )
        #endif
        return false
      }
```

russellwheatley (changes requested):
Thanks for tracking this down, the repro steps are clear and the guard matches the pattern `ensureAPNSTokenSetting()` already uses in this file.

Left two suggestions: logging a debug-only breadcrumb when a URL gets silently dropped, so it's easier to notice in dev builds if this path fires and a real deep link ends up going nowhere.

Also, the CLA check is currently failing, could you sign it at https://cla.developers.google.com/ so this can get merged? 🙏

PR #18653, opened 2026-09-07
fix(auth,windows): complete reauthenticateWithCredential on success

### LINKED ISSUE #18632: 🐛 [firebase_auth] Windows: reauthenticateWithCredential never completes on success
### Bug report

On Windows, `User.reauthenticateWithCredential()` never completes when the credential is valid. The returned `Future` stays pending forever. An invalid credential rejects normally.

`FirebaseAuthPlugin::ReauthenticateWithCredential` in the Windows plugin only invokes the Pigeon `result` callback on the error branch:

```cpp
// windows/firebase_auth_plugin.cpp
firebase::Future<void> future =
    user.Reauthenticate(getCredentialFromArguments(input, app));

future.OnCompletion([result](const firebase::Future<void>& completed_future) {
  if (completed_future.error() == 0) {
    // TODO: wrong return type
  } else {
    result(FirebaseAuthPlugin::ParseError(completed_future));
  }
});
```

`User::Reauthenticate()` in the C++ SDK returns `Future<void>`, but the Pigeon signature expects an `InternalUserCredential`, so the success branch was left empty. Nothing calls `result` and the Dart side waits forever. It is the only method in the file that returns nothing on success: `ReauthenticateWithProvider` just below calls `result(credential)`, and `Reload` deals with the same `Future<void>` shape by reading `firebaseAuth->current_user()` once the future completes and building the reply from that.

### Steps to reproduce

1. Run a Flutter Windows app with an email/password Firebase user signed in.
2. Call:

```dart
final cred = EmailAuthProvider.credential(email: user.email!, password: correctPassword);
await user.reauthenticateWithCredential(cred);
print('never printed');
```

3. Pass the correct password.

### Expected

The future completes with a `UserCredential`, as it does on Android, iOS, macOS and web.

### Actual

The future never completes, with no error and no exception. Anything awaiting it is stuck until the process exits.

A wrong password completes normally with a `FirebaseAuthException`, since the error branch calls `result`.

### Impact

Any Windows flow that requires credential reauthentication first cannot complete. Ours is the account email-change flow, where `reauthenticateWithCredential` runs ahead of `verifyBeforeUpdateEmail`. The verification email is never requested and the user sees nothing happen. `verifyBeforeUpdateEmail` is implemented correctly, it just isn't reached.

### Possible fix

`Reload` already shows the pattern for a `Future<void>` C++ call that has to return a richer Pigeon type: read the current user after completion and construct the reply from it.

### Version

firebase_auth 6.5.7. The same code is in 6.6.1 (current latest) at the same place.

SelaseKay:
## Description

On Windows, `User.reauthenticateWithCredential()` never completed when the credential was valid. The C++ plugin called `User::Reauthenticate()`, which returns `Future<void>`, then left the Pigeon success branch empty (`// TODO: wrong return type`). Dart's `Future` stayed pending forever. Invalid credentials still failed normally because only the error branch invoked `result`.

This switches to `User::ReauthenticateAndRetrieveData()`, which returns `AuthResult`, and completes the Pigeon callback the same way as `LinkWithCredential` / `ReauthenticateWithProvider`.

The `reauthenticateWithCredential()` e2e group now runs on Windows. The last case in that group is skipped on Windows and macOS because it exercises `signInWithEmailAndPassword`, not reauthentication (a test-level skip overrides the group skip). The `user-mismatch` assertion accepts the Windows C++ SDK message, which omits the trailing period. macOS still skips the rest of the group.

## Related Issues

- Fixes https://github.com/firebase/flutterfire/issues/18632

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_auth/firebase_auth/example/integration_test/firebase_auth_user_e2e_test.dart
+++ b/packages/firebase_auth/firebase_auth/example/integration_test/firebase_auth_user_e2e_test.dart
@@ -298,40 +298,33 @@ void main() {
                 defaultTargetPlatform == TargetPlatform.macOS),
       );
 
-      // Kept outside the skipped group so Windows covers the success path
-      // that previously hung. Remaining cases stay skipped on desktop
-      // because they assert platform-specific error messages.
-      test(
-        'reauthenticateWithCredential() should reauthenticate correctly',
-        () async {
-          // Setup
-          await FirebaseAuth.instance.createUserWithEmailAndPassword(
-            email: email,
-            password: testPassword,
-          );
-          final initialUser = FirebaseAuth.instance.currentUser;
-
-          // Test
-          AuthCredential credential = EmailAuthProvider.credential(
-            email: email,
-            password: testPassword,
-          );
-          await FirebaseAuth.instance.currentUser!
-              .reauthenticateWithCredential(credential);
-
-          // Assertions
-          final currentUser = FirebaseAuth.instance.currentUser;
-          expect(currentUser, isNot(equals(null)));
-          expect(initialUser, isNot(equals(null)));
-          expect(currentUser?.email, equals(email));
-          expect(currentUser?.uid, equals(initialUser?.uid));
-        },
-        skip: !kIsWeb && defaultTargetPlatform == TargetPlatform.macOS,
-      );
-
       group(
         'reauthenticateWithCredential()',
         () {
+          test('should reauthenticate correctly', () async {
+            // Setup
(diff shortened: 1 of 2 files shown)


JulianPscheid:
We're hitting this in production. Two Windows users couldn't change their email in the last two weeks because `reauthenticateWithCredential` never completes, and our only workaround is sending them to another platform. A review when someone has time would be appreciated.

Lyokone (approved):

demolaf (approved):

